PR #3, opened 2022-07-07
add scripts/TestTradesV2.py

wakamex:
- add file scripts/TestTradesV2.py to test v2 trades against simulated values
- add class YieldsSpacev2_Pricing_model in scripts/PrincingModels.py
- removes recursive solving for liquidity at initiation in v1
- adds ,c,u parameters to v1 calls to be compatible with updated Market class


jrhea (approved):
This looks good. I'm going to test this manana

PR #19, opened 2022-10-02
Hyperdrive

jrhea:
example code wasn't using the updated reserves AND closed form equation didn't account for total supply


wakamex (approved):
I thoroughly approved every line and it looks good

PR #25, opened 2022-10-20
user sim

wakamex:
implementing beginning of user behavior, starting with stochastic trade direction to hit a target range of pool apy. this allows us to simulate rate hike and drop scenarios.


wakamex:
lgtm

dpaiton (approved):

PR #26, opened 2022-10-22
reorganizes user class to be more abstracted

dpaiton:


src/elfpy/simulators.py
```diff
@@ -304,38 +296,12 @@ def run_simulation(self, override_dict=None):
...
-                self.trade_amount_usd = self.get_trade_amount_usd(
-                    self.token_in,
-                    self.trade_direction,
-                    self.target_daily_volume,
-                    self.base_asset_price,
+                (self.token_in, self.token_out, self.trade_amount_usd) = self.user.get_trade(
+                    self.market, self.tokens, self.trade_direction, self.target_daily_volume, self.base_asset_price
                 )
+                if self.user_type.lower() == "weightedrandom":  # sync user with market
+                    self.user.set_market_apy(self.market.apy(self.get_days_remaining()))
```
wakamex, line 304: why are we calling this from the simulators class? the user could update this themselves by accessing the market, no? is it more efficient this way, since we only need to update once after we know the apy changed, as opposed to calling the function every time the apy is referenced?

dpaiton: You set it up so that the weighting in WeightedRandom is a function of the Market APY. This was to avoid having to pass the entire market object to the User class, which was previously done only for that one variable.

wakamex (approved):
lgtm

src/elfpy/simulators.py
wakamex: right, i thought we were still doing that for a moment. now we pass just the apy, which is much more lightweight.

PR #28, opened 2022-10-26
Incorporate the Hyperdrive Pricing Model

jalextowle:
This PR adds a full unit test suite for the Hyperdrive pricing model


tests/test_pricing_model.py
```diff
@@ -0,0 +1,1279 @@
...
+                # k = c/u*(u*z)**(1-τ) + (2*y + c*z)**(1-τ)
+                # k = 100000**0.9774641559684029528500691670222 + (2*100000 + 100000*1)**0.9774641559684029528500691670222
+                # k = 302929.51067963685
+                (
+                    TradeResult(
+                        # p = ((2y+cz)/uz)**τ
+                        #   = 1.0250671833648672
+                        # without_fee_or_slippage = 1/p * out = 97.55458141947516
+                        without_fee_or_slippage=97.55458141947516
+                        # fee is 10% of discount before slippage = (100-97.55458141947516)*0.2 = 2.4454185805248443*0.2 = 0.4887960189720616
```
dpaiton, line 181: should comment say 20%?

tests/test_pricing_model.py
```diff
@@ -0,0 +1,1279 @@
...
+                test_case.token_in,
+                test_case.fee_percent,
+                time_remaining,
+                test_case.init_share_price,
+                test_case.share_price,
+            )
+            actual = TradeResult(without_fee_or_slippage, with_fee, without_fee, fee)
+            compare_trade_results(actual, expected)
+
+    def test_calc_out_given_in_success(self):
```
dpaiton, line 403: should probably rename this to match the other function, i.e. `test_calc_out_given_in` right?

tests/test_pricing_model.py
```diff
@@ -0,0 +1,1279 @@
...
+                    #   d_y = 300_000 - (k - 110_000 ** (1 - T)) ** (1 / (1 - T)) = 10235.514826394327
+                    #
+                    # Note that this is smaller than the without slippage value
+                    10235.514826394327,
+                    # Since we are buying bonds, in_ is an amount of base and we calculate the fee using the spot price as:
+                    #   fee = 0.01 * (p - 1) * 10_000 = 2.506718336486724
+                    2.506718336486724,
+                ),
+            ),
+            # TODO: The slippage should arguably be much higher. This is something
```
dpaiton, line 519: might consider moving this comment into the pricing model code and referencing the test, otherwise we might lose track of it.

tests/test_pricing_model.py
dpaiton, line 403: Can you add a comment block outlining all of the test cases below?

tests/test_pricing_model.py
```diff
@@ -0,0 +1,1279 @@
...
+    np.testing.assert_almost_equal(
+        actual.without_fee_or_slippage, expected.without_fee_or_slippage, err_msg="unexpected without_fee_or_slippage"
+    )
+    np.testing.assert_almost_equal(actual.with_fee, expected.with_fee, err_msg="unexpected without_fee")
+    np.testing.assert_almost_equal(actual.without_fee, expected.without_fee, err_msg="unexpected fee")
+    np.testing.assert_almost_equal(actual.fee, expected.fee, err_msg="unexpected with_fee")
+
+
+class TestHyperdrivePricingModel(unittest.TestCase):
+    def test_calc_in_given_out(self):
```
dpaiton, line 114: Can you add a comment block outlining all of the test cases below?

src/elfpy/pricing_models.py
```diff
@@ -4,6 +4,7 @@
 TODO: rewrite all functions to have typed inputs
 """
 
+import types
```
dpaiton, line 7: This is unused -- can you delete it?

src/elfpy/pricing_models.py
```diff
@@ -994,7 +1014,23 @@ def calc_out_given_in(
...
+        # we skip straight from non-negative reals to the complex plane without
+        # hitting negative reals.
+        #
+        # Ensure that the outputs are all non-negative floats. We only need to
+        # check with_fee since without_fee_or_slippage will always be a positive
+        # float due to the constraints on the inputs, without_fee = with_fee + fee
+        # so it is a positive float if with_fee and fee are positive floats, and
+        # fee is a positive float due to the constraints on the inputs.
+        assert (
+            type(with_fee) == float
```
dpaiton, line 1028: I think it is better to use [isinstance](https://docs.python.org/3/library/functions.html#isinstance) instead. Also, might as well raise a TypeError instead of failed assertion.

dpaiton: https://switowski.com/blog/type-vs-isinstance/

tests/test_pricing_model.py
```diff
@@ -0,0 +1,1281 @@
...
+                    #
+                    # The output is d_x = c * d_z. Since c = 1, d_x = d_z.
+                    # Note that this is slightly smaller than the without slippage value
+                    without_fee=9740.77011591768,
+                    # Combining the without_fee and the fee, we calculate with_fee as:
+                    #   with_fee = 9740.77011591768 - 2.4454185805248496 = 9738.324697337155
+                    with_fee=9738.324697337155,
+                ),
+            ),
+            # TODO: The slippage should arguably be much higher. This is something
```
dpaiton, line 834: This is a repeat of the above comment, right? I recommend just including one TODO comment.

tests/test_pricing_model.py
```diff
@@ -0,0 +1,1281 @@
...
+                    without_fee=212.47551672440022,
+                    # Combining the without_fee and the fee, we calculate with_fee as:
+                    #   with_fee = 212.47551672440022 - 0.1247814132813505 = 212.35073531111888
+                    with_fee=212.35073531111888,
+                ),
+            ),
+        ]
+
+        # Test cases where token_out = "base".
+        base_out_test_cases = [
```
dpaiton, line 719: Are all of these tests a repeat of the above but with `token_out = "base"`? If so, is there a simpler way to specify it? I think the below isn't exactly correct, but something like this:
```
base_out_test_cases = []
for test_case in pt_out_test_cases:
    base_test_case = test_case.copy()
    base_test_case.token_out = "base"
    base_out_test_cases.append(base_test_case)
```

tests/test_pricing_model.py
dpaiton: Oh, I see, there is a `success` and `failure` test. Then do we need those for in_given_out?

tests/test_pricing_model.py
```diff
@@ -0,0 +1,1281 @@
...
+                (
+                    TradeResult(
+                        # p = ((2y+cz)/uz)**τ
+                        #   = 1.0250671833648672
+                        # without_fee_or_slippage = 1/p * out = 97.55458141947516
+                        without_fee_or_slippage=97.55458141947516
+                        # fee is 10% of discount before slippage = (100-97.55601990513969)*0.1 = 2.4454185805248443*0.1 = 0.24454185805248443
+                        ,
+                        fee=0.24454185805248443
+                        # deltaZ = 1/u * (u/c*(k - (2*y + c*z - deltaY)**(1-τ)))**(1/(1-τ)) - z
```
dpaiton, line 144: I think this is the equation for delta z prime, not delta z in the hackmd. Is that ok?

jalextowle: Yes. @wakamex can you make this update when you update the comments for your test? I updated mine already.

tests/test_pricing_model.py
jalextowle: Yes we do. @wakamex should be on it, but please let me know if you need me to handle it @wakamex.

jrhea (approved):
This looks good to merge, but i have one strong suggestion for how to target test cases:

I've found that it is more useful to think about test cases in terms of: apr( 1%, 5%, 10%, 20%, ... ,80%, 90%, 100%) and trade size as a fraction of the reserves (.01, .1, .2,...,max feasible trade). This removes the reserve size question and simplifies it to only needing to test a trade as a fraction of the reserves. Additionally, using apr specifically and letting the reserves fallout out from that calculation allows us to target specific cases rather than just randomly picking a reserve amount and letting the apr fallout from that.

I think it is fine to do in a subsequent PR though

PR #29, opened 2022-10-28
Multi users

dpaiton:


pr_strategy.txt
```diff
@@ -0,0 +1,98 @@
+--------------------------------------------------------------------
```
jalextowle, line 1: Let's remove this file

src/elfpy/markets.py
```diff
@@ -48,71 +55,32 @@ def __init__(
...
-        if self.time_remaining < 0:
-            assert False, (
-                f"ERROR: the time variable market.time_remaining={self.time_remaining} should never be negative."
-                + f"\npricing_model={self.pricing_model}"
-            )
+    def base_reserves(self):
+        """get base reserves"""
+        return self.share_reserves * self.share_price
 
     def get_target_reserves(self, token_in, trade_direction):
```
jalextowle, line 65: I think we may ultimately get rid of this function. Is the intention behind keeping it with "fyt" to avoid breaking more things? Regardless, I'm not concerned about it for the merge to dev.

src/elfpy/pricing_models.py
```diff
@@ -69,7 +69,7 @@ def norm_days(days, normalizing_constant=365):
         return days / normalizing_constant
 
     @staticmethod
-    def _stretch_time(time, time_stretch=1):
+    def stretch_time(time, time_stretch=1):
```
jalextowle, line 72: Again, this shouldn't block a merge, but don't we want to remove this from the class and put it in a standalone module?

src/elfpy/pricing_models.py
```diff
@@ -462,132 +472,578 @@ def calc_liquidity(
...
+            "d_token_asset_volume": 0,
+        }
+        wallet_deltas = {
+            "base_in_wallet": output_with_fee,
+            "base_in_protocol": [trade_details["mint_time"], 0],
+            "token_in_wallet": [trade_details["mint_time"], -1 * trade_details["trade_amount"]],
+            "token_in_protocol": [trade_details["mint_time"], 0],
+            "fee": [trade_details["mint_time"], fee],
+        }
+        return market_deltas, wallet_deltas
```
jalextowle, line 665: I think it would be nicer to have these functions in the market class. They aren't actually pricing model specific since they just use the generic `calc_in_given_out` and `calc_out_given_in` functions, so we could technically support all of these operations with any pricing model that implements those basic functions. Furthermore, these functions are fairly accounting heavy (especially the short), which isn't really the point of the pricing model from my perspective. We could use the Hyperdrive pricing model to create other actions that aren't enumerated here as well as to accomplish the same goals with different accounting. 

As with the rest of my comments so far, I don't see this as blocking. Having said that, I feel strongly that we should discuss the merits of including this logic in this layer of abstraction.

src/elfpy/simulators.py
```diff
@@ -235,34 +226,43 @@ def setup_pricing_and_market(self, override_dict=None):
...
+            time_stretch_constant,
             self.init_share_price,  # u from YieldSpace w/ Yield Baring Vaults
             self.init_share_price,  # c from YieldSpace w/ Yield Baring Vaults
         )
         init_base_asset_reserves, init_token_asset_reserves = init_reserves[:2]
-        init_time_remaining = self.pricing_model.days_to_time_remaining(self.pool_duration, self.init_time_stretch)
         self.market = Market(
-            base_asset=init_base_asset_reserves,  # x
-            token_asset=init_token_asset_reserves,  # y
+            share_reserves=init_base_asset_reserves,  # x
```
jalextowle, line 244: Nit: "x" => "z". It would also be good to leave a TODO here so that we remember to update from `init_base_asset_reserves` to another value.

src/elfpy/utils/time.py
```diff
@@ -0,0 +1,34 @@
+"""
+Helper functions for converting time units
+"""
+
+
+import datetime
+import pytz
+
+import numpy as np
+
```
jalextowle, line 10: It would be good to add time stretching functions here.

src/elfpy/user.py
```diff
@@ -4,178 +4,114 @@
...
+            self.wallet["base_in_wallet"],
+            self.market.share_reserves,
+            self.market.bond_reserves,
+            "base",
+            self.market.fee_percent,
+            stretched_time_remaining,
+            self.market.init_share_price,
+            self.market.share_price,
+        )[1]
+        max_short = self.wallet["base_in_wallet"] + output_with_fee - eps
```
jalextowle, line 62: I'm pretty skeptical of this. I'll circle back on this at the end of the review to double check on this.

jalextowle (commented):
Everything aside from `get_max_short` is looking good to me. That may be correct, but it isn't obvious to me at a glance. I'll take another look now.

jalextowle (approved):

dpaiton:
I think get_max_short is ok as an initial pass. We need to think of a fast approximation to iteratively solving it, and probably include an optional precise iterative solution.

examples/notebooks/multi_user.ipynb
```diff
@@ -0,0 +1,161 @@
...
+        "    \"verbose\": False,\n",
+        "}\n",
+        "\n",
+        "config[\"user_policy\"] = {\n",
+        "    \"trade\": {\n",
+        "        \"conditional\": {\n",
+        "            \"if\": {\n",
+        "                (\"gt\", \"market.apy\", 0.5)\n",
+        "            },\n",
+        "            \"then\": \"buy\"\n",
```
MazyGio, line 80: missing comma. Update to:
        "            \"then\": \"buy\",\n",

src/elfpy/markets.py
```diff
@@ -151,185 +118,64 @@ def check_fees(
...
+        trade_details["share_reserves"] = self.share_reserves
+        trade_details["bond_reserves"] = self.bond_reserves
+        # for each position, specify how to forumulate trade and then execute
+        if user_action["action_type"] == "open_long":  # buy to open long
+            trade_details["direction"] = "out"  # calcOutGivenIn
+            trade_details["token_out"] = "pt"  # buy unknown PT with known base
+            market_deltas, wallet_deltas = self.pricing_model.open_long(trade_details)
+        elif user_action["action_type"] == "close_long":  # sell to close long
+            trade_details["direction"] = "out"  # calcOutGivenIn
+            trade_details["token_out"] = "base"  # buy unknown PT with known base
```
MazyGio, line 156: I might be misunderstanding, but I think the comment should read
`# sell known PT for unknown base`

src/elfpy/markets.py
dpaiton: I believe Mihai added this for something that no longer exists. I'll remove it.

src/elfpy/pricing_models.py
dpaiton: Moved these functions to the time utils file

src/elfpy/simulators.py
dpaiton: Added the TODO

src/elfpy/utils/time.py
dpaiton: Done.

examples/notebooks/multi_user.ipynb
dpaiton: removed this notebook -- it's dumb.

src/elfpy/user.py
dpaiton: Added a comment to resolve later

PR #46, opened 2022-11-15
Fix test trade (WIP)

wakamex:
basic logging shows endless attempts to close long, which should only be getting called once:
image


wakamex:
this should be merging into dev not main

wakamex:
this should be merging into dev not main

PR #48, opened 2022-11-16
added failure tests: if the failure tests pass, are they a success?

wakamex:


src/elfpy/pricing_models.py
```diff
@@ -824,6 +842,23 @@ def calc_in_given_out(
...
             f"\n\twithout_fee={without_fee}\n\tfee={fee}"
         )
+
+        # TODO(jalextowle): With some analysis, it seems possible to show that
+        # we skip straight from non-negative reals to the complex plane without
+        # hitting negative reals.
+        #
+        # Ensure that the outputs are all non-negative floats. We only need to
+        # check with_fee since without_fee_or_slippage will always be a positive
+        # float due to the constraints on the inputs, without_fee = with_fee + fee
```
jalextowle, line 852: In this case, `without_fee = with_fee - fee`, so we should instead check that `without_fee` is a non-negative float. We can change the comment to indicate that we don't need to check `with_fee` since `with_fee = without_fee + fee`.

jalextowle (approved):
LGTM once that comment is updated.

src/elfpy/pricing_models.py
wakamex: sounds like we can remove it, don't see an upper-bound that's worth checking

wakamex: oh i see, we'll just check the lower of the two, before fees for non-negativity 👌

Issue #34, opened 2022-11-10
backwards compatibility for Element V1


dpaiton:
Ensure element v1 works with new refactored code & is tested.

dpaiton:
should also include verifying that previous helper functions still work (e.g. getting APY, spot price, share price, etc)

dpaiton:
@jalextowle can we close this issue now? or are you waiting to finish the tests?

PR #50, opened 2022-11-16
Pricing Model and Market Refactor

jalextowle:
Fixes: #34.


src/elfpy/markets.py
```diff
@@ -115,80 +203,207 @@ def check_fees(
...
+        """Update the spot price"""
+        self.spot_price = self.pricing_model.calc_spot_price_from_reserves(
+            share_reserves=self.share_reserves,
+            bond_reserves=self.bond_reserves,
+            init_share_price=self.init_share_price,
+            share_price=self.share_price,
+            time_remaining=time_utils.stretch_time(self.token_duration, self.time_stretch_constant),
+        )
+
+    def __open_short(self, trade_details):
```
dpaiton, line 220: Why are you using the mangled lead instead of a single underscore?
https://docs.python.org/3/tutorial/classes.html#private-variables

Seems like this would only be relevant if open_short also existed as other class methods?

src/elfpy/markets.py
```diff
@@ -115,80 +203,207 @@ def check_fees(
...
-                time_remaining=time_utils.stretch_time(
-                    self.token_duration, self.time_stretch_constant
-                )
-            )
-        else:
-            self.spot_price = np.nan
+    def __get_time_parameter_for_model(self, trade_details):
+        pricing_model_name = self.pricing_model.model_name()
+        if pricing_model_name == "Element":
+            return trade_details["time_remaining"]
```
dpaiton, line 404: element has time stretch too, right? What's the difference here?

src/elfpy/pricing_models.py
```diff
@@ -65,282 +66,71 @@ def model_name(self):
...
-            )
-        return (base_asset_reserves, token_asset_reserves, total_liquidity)
-
 
 class ElementPricingModel(PricingModel):
     """
     Element v1 pricing model
 
-    Does not use the Yield Bearing Vault `init_share_price` (u) and `share_price` (c) variables.
+    Does not use the Yield Bearing Vault `init_share_price` (mu) and `share_price` (c) variables.
```
dpaiton, line 133: nit: you probably want to make mu to the actual mu character like below?

src/elfpy/simulators.py
```diff
@@ -329,53 +331,57 @@ def run_simulation(self, override_dict=None):
...
+                    and (self.market.time - last_user_action_time) % 0.5 < 1 / 365 / self.num_blocks_per_day / 2
+                ):
                     print(
-                        f"t={bcolors.HEADER}{self.market.time}{bcolors.ENDC}"+
-                        f" reserves=[x={bcolors.OKBLUE}{self.market.share_reserves}{bcolors.ENDC}"+
-                        f",y={bcolors.OKBLUE}{self.market.bond_reserves}{bcolors.ENDC}]\n"+
-                        f" no user action 😴"+
-                        f" user report = {self.user_list[0].status_report()}"
+                        f"t={bcolors.HEADER}{self.market.time}{bcolors.ENDC}"
+                        + f" reserves=[x={bcolors.OKBLUE}{self.market.share_reserves}{bcolors.ENDC}"
```
dpaiton, line 365: I don't think you need the plus at all

src/elfpy/pricing_models.py
```diff
@@ -65,282 +66,71 @@ def model_name(self):
...
+            share_reserves,
+            bond_reserves,
             time_remaining,
             init_share_price,
             share_price,
         )
-        days_remaining = self.time_to_days_remaining(time_remaining, time_stretch)
-        apy = self.calc_apy_from_spot_price(spot_price, time_utils.norm_days(days_remaining))
+        days_remaining = time_utils.time_to_days_remaining(time_remaining, time_stretch)
+        apy = price_utils.calc_apr_from_spot_price(spot_price, time_utils.norm_days(days_remaining))
```
dpaiton, line 125: change to apr?

src/elfpy/pricing_models.py
```diff
@@ -864,7 +750,6 @@ def calc_in_given_out(
     def calc_out_given_in(
         self,
         in_,
-        # TODO: This should be share_reserves when we update the market class
         share_reserves,
         bond_reserves,
         token_out,
```
dpaiton, line 755: Can you move these TODOs out of the function handle?

src/elfpy/utils/time.py
```diff
@@ -39,6 +39,12 @@ def norm_days(days, normalizing_constant=365):
     return days / normalizing_constant
 
 
+def calc_time_stretch(apy):
```
dpaiton, line 42: I think this should probably live with the pricing model. Right now it's the same, but we've talked a lot about potentially changing this up (e.g. making it dynamic) or removing this for HyperDrive.

tests/test_pricing_model.py
```diff
@@ -80,37 +95,32 @@ class TestCaseCalcOutGivenInFailure:
...
-        )
-        np.testing.assert_almost_equal(self.with_fee, other.with_fee, err_msg="unexpected without_fee")
-        np.testing.assert_almost_equal(self.without_fee, other.without_fee, err_msg="unexpected fee")
-        np.testing.assert_almost_equal(self.fee, other.fee, err_msg="unexpected with_fee")
-
 
-class TestHyperdrivePricingModel(unittest.TestCase):
-    """Unit tests for the Hyperdrive pricing model"""
+# NOTE: To maximally re-use the unit test cases, we pass in stretched time to
+# the Element pricing model, which is not what happens in practice. This said,
```
dpaiton, line 113: What do you mean by this? We use stretched time in Element v1.

tests/test_pricing_model.py
```diff
@@ -11,8 +11,8 @@
 import unittest
```
dpaiton, line 11: Might want to consider making this several files so that it is easier to find stuff

src/elfpy/simulators.py
jalextowle: This was automatically done by my auto-formatter. I'll see if I can add something about this to the black configuration.

src/elfpy/markets.py
jalextowle: I was just trying to make them private, but it seems like I read some bad information. I'll change these.

tests/test_pricing_model.py
jalextowle: I got this from looking at the v0.0.1 release: https://github.com/element-fi/elf-simulations/blob/v0.0.1/src/elfpy/markets.py#L229. I'll take your word for the fact that V1 uses stretched time (now that you mention it, it would have to since the time stretch is in the construction paper).

dpaiton (approved):
Looks good!

Issue #35, opened 2022-11-10
Add LPer user type & refactor market initialization


dpaiton:
The market should be empty on init, and then an LPer has to add funds before trades can happen.

Add LP user type with ability to receive fees and add/remove liquidity.

wakamex:
to do
- [ ] make market initialize with 0 assets
- [ ] make initial LP provide liquidity to hit target liquidity and APY
- [ ] withdrawal logic

wakamex:
1 of 3 check boxes done 😭

wakamex:
Fixes: #35

wakamex:
5 of 5 check boxes done

PR #52, opened 2022-11-17
add LP user

wakamex:
Fixes: #35


src/elfpy/markets.py
```diff
@@ -115,42 +124,56 @@ def check_fees(
...
-        self.base_asset_orders += market_deltas["d_base_asset_orders"]
-        self.token_asset_orders += market_deltas["d_token_asset_orders"]
-        self.base_asset_volume += market_deltas["d_base_asset_volume"]
-        self.token_asset_volume += market_deltas["d_token_asset_volume"]
-
-    def swap(self, user_action):
+        self.share_buffer += market_deltas["d_share_buffer"] if "d_share_buffer" in market_deltas else 0
+        self.bond_buffer += market_deltas["d_bond_buffer"] if "d_bond_buffer" in market_deltas else 0
+        if trade_type in ["add_liquidity", "remove_liquidity"]:
+            self.liquidity_pool += market_deltas["d_liquidity_pool"] if "d_liquidity_pool" in market_deltas else 0
```
dpaiton, line 142: If you're already checking that `"d_liquidity_pool"` is in `market_deltas` then why do you also need to check the value of `trade_type`? Couldn't we keep the function signature the same and check for `"d_liquidity_pool"` and then set += to either `market_deltas["d_liquidity_pool"]` or `0`?

src/elfpy/markets.py
```diff
@@ -115,42 +124,56 @@ def check_fees(
...
-        self.cum_token_asset_fees += market_deltas["d_token_asset_fee"]
-        self.base_asset_orders += market_deltas["d_base_asset_orders"]
-        self.token_asset_orders += market_deltas["d_token_asset_orders"]
-        self.base_asset_volume += market_deltas["d_base_asset_volume"]
-        self.token_asset_volume += market_deltas["d_token_asset_volume"]
-
-    def swap(self, user_action):
+        self.share_buffer += market_deltas["d_share_buffer"] if "d_share_buffer" in market_deltas else 0
+        self.bond_buffer += market_deltas["d_bond_buffer"] if "d_bond_buffer" in market_deltas else 0
+        if trade_type in ["add_liquidity", "remove_liquidity"]:
```
dpaiton, line 141: If you're already checking that `"d_liquidity_pool"` is in `market_deltas` then why do you also need to check the value of `trade_type`? Couldn't we keep the function signature the same and check for `"d_liquidity_pool"` and then set += to either `market_deltas["d_liquidity_pool"]` or `0`?

Also could we set the other deltas to 0 in the case of adding liquidity? so it always looks like a "swap" to simplify this logic?

src/elfpy/markets.py
```diff
@@ -29,19 +29,26 @@ def __init__(
...
         share_reserves,
         bond_reserves,
+        liquidity_pool,
         fee_percent,
         token_duration,
         pricing_model,
         time_stretch_constant=1,
         init_share_price=1,
         share_price=1,
+        end_of_time=None,
```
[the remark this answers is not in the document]
wakamex: yup it was sketchy. it's useful to have the user close out their position at the last moment, so i can value their return more easily.
instead of that, could 1) have the market class tell the user to close up their positions or 2) get the valuation of unclosed positions
we probably need 2) anyway... just feels like actually closing the position is the easiest way to do that. maybe a wrapper like user.closePositionButNotReallyJustGiveMeItsValuationHehe()

CPSTL:
Task linked: [CU-3w3988r Add LPers & market refactor](https://app.clickup.com/t/3w3988r)

config/hyperdrive_config.toml
```diff
@@ -0,0 +1,38 @@
+title = "example simulation config"
```
dpaiton, line 1: Why did you add this file? Can you just modify `simulation_config` instead? What is the difference between the two? We should be weary about this config folder blowing up for random use-cases.

config/simulation_config.toml
```diff
@@ -27,6 +27,7 @@ floor_fee = 0  # minimum fee percentage (bps)
 pool_duration = 180  # in days
```
dpaiton, line 27: pool_duration is no longer used afaik

src/elfpy/utils/basic_dataclass.py
```diff
@@ -0,0 +1,44 @@
...
+class BasicDataclass:
+    """A basic dataclass with a few useful methods"""
+
+    def __getitem__(self, key):
+        return getattr(self, key)
+
+    def __setitem__(self, key, value):
+        setattr(self, key, value)
+
+    def display(self):
```
dpaiton, line 15: how is this different from `__repr__`?

tests/test_lpers.py
```diff
@@ -0,0 +1,42 @@
...
+
+class LPTests(BaseLPTest):
+    """Tests for the Simple LP policy"""
+
+    def test_base_LPs(self):
+        """Tests base LP setups"""
+        self.run_base_lp_test(user_policies=["simple_LP"],config_file="config/hyperdrive_config.toml")
+
+    # def test_complicated_LPs(self):
+        """Tests complicated LP setups"""
```
dpaiton, line 42: add newline

src/elfpy/simulators.py
```diff
@@ -4,20 +4,13 @@
...
 """
-
-
 from importlib import import_module
-
-
-import numpy as np
-
-
+from elfpy.utils.fmt import *  # float→str formatter, also imports numpy as np
```
dpaiton, line 8: imports should be explicit per file -- so import it here even if it is imported in fmt.

src/elfpy/markets.py
```diff
@@ -27,21 +52,24 @@ class Market:
 
     def __init__(
         self,
-        share_reserves,
-        bond_reserves,
         fee_percent,
         token_duration,
         pricing_model,
+        share_reserves = 0,
```
dpaiton, line 58: remove space around =

src/elfpy/markets.py
```diff
@@ -4,8 +4,12 @@
 TODO: rewrite all functions to have typed inputs
 """
 
-import numpy as np
```
dpaiton, line 7: add this back -- if we use it then we should include it here.

src/elfpy/markets.py
```diff
@@ -377,23 +470,148 @@ def _close_long(self, trade_details):
...
+        output_string += f"\n liquidity_pool_history={self.liquidity_pool_history}"
+        output_string += f"\n share_fee_history={self.share_fee_history}"
+        output_string += f"\n token_fee_history={self.token_fee_history}"
+        if self.verbose:
+            output_string += f" fees=["\
+                + f"x:{bcolors.OKBLUE}{self.share_fees}{bcolors.ENDC}"\
+                + f",y:{bcolors.OKBLUE}{self.token_fees}{bcolors.ENDC}"
+        output_string += f" p={bcolors.FAIL}{self.spot_price}{bcolors.ENDC}"
+        output_string += f" rate={bcolors.FAIL}{self.rate}{bcolors.ENDC}"
+        return output_string
```
dpaiton, line 617: add newline

src/elfpy/markets.py
```diff
@@ -377,23 +470,148 @@ def _close_long(self, trade_details):
...
+                + f",y_b:{bcolors.OKBLUE}{self.bond_buffer}{bcolors.ENDC}"\
+            + f"]"
+        output_string += f"\n liquidity_pool_history={self.liquidity_pool_history}"
+        output_string += f"\n share_fee_history={self.share_fee_history}"
+        output_string += f"\n token_fee_history={self.token_fee_history}"
+        if self.verbose:
+            output_string += f" fees=["\
+                + f"x:{bcolors.OKBLUE}{self.share_fees}{bcolors.ENDC}"\
+                + f",y:{bcolors.OKBLUE}{self.token_fees}{bcolors.ENDC}"
+        output_string += f" p={bcolors.FAIL}{self.spot_price}{bcolors.ENDC}"
```
dpaiton, line 615: why do you call this `FAIL`? is the spot price incorrect here?

src/elfpy/markets.py
```diff
@@ -55,7 +83,7 @@ def __init__(
         if pricing_model_name == "Element":
             self.allowed_actions = ["open_long", "close_long"]
```
dpaiton, line 84: Shouldn't element also have `add_liquidity` and `remove_liquidity`?

src/elfpy/markets.py
```diff
@@ -55,7 +83,7 @@ def __init__(
...
         elif pricing_model_name == "Hyperdrive":
-            self.allowed_actions = ["open_long", "close_long", "open_short", "close_short"]
+            self.allowed_actions = ["open_long", "close_long", "open_short", "close_short", "add_liquidity", "remove_liquidity"]
```
dpaiton, line 86: Shouldn't element also have `add_liquidity` and `remove_liquidity`?

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
-        self.spot_price = 0
+        self.share_fees = 0
+        self.share_fee_history = {}
+        self.token_fees = 0
+        self.token_fee_history = {}
+        self.spot_price = None
         self.total_supply = self.share_reserves + self.bond_reserves
         self.verbose = verbose
+        self.rate = 0
+        self.update_spot_price_and_rate() if self.share_reserves > 0 else None
```
dpaiton, line 106: The else in this conditional is nonsensical. Can you just make it two lines?

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
         """
-        # assign general trade details, irrespective of trade type
-        user_action.fee_percent = self.fee_percent
-        user_action.init_share_price = self.init_share_price
-        user_action.share_price = self.share_price
-        user_action.share_reserves = self.share_reserves
-        user_action.bond_reserves = self.bond_reserves
         # ensure that the user action is an allowed action for this market
-        if not user_action["action_type"] in self.allowed_actions:
+        if not agent_action.action_type in self.allowed_actions:
```
dpaiton, line 114: Why is this a class member variable if it is only used here?

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
-        self.bond_reserves += market_deltas["d_token_asset"]
-        self.cum_base_asset_slippage += market_deltas["d_base_asset_slippage"]
-        self.cum_token_asset_slippage += market_deltas["d_token_asset_slippage"]
-        self.cum_base_asset_fees += market_deltas["d_base_asset_fee"]
-        self.cum_token_asset_fees += market_deltas["d_token_asset_fee"]
-        self.base_asset_orders += market_deltas["d_base_asset_orders"]
-        self.token_asset_orders += market_deltas["d_token_asset_orders"]
-        self.base_asset_volume += market_deltas["d_base_asset_volume"]
-        self.token_asset_volume += market_deltas["d_token_asset_volume"]
+        for field in market_deltas.__dataclass_fields__:
```
dpaiton, line 171: isn't this resolved more easily with the `BasicDataclass` additions? like `market_deltas.items()` or whatever?

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
-        user_action.init_share_price = self.init_share_price
-        user_action.share_price = self.share_price
-        user_action.share_reserves = self.share_reserves
-        user_action.bond_reserves = self.bond_reserves
         # ensure that the user action is an allowed action for this market
-        if not user_action["action_type"] in self.allowed_actions:
+        if not agent_action.action_type in self.allowed_actions:
             raise AssertionError(
-                f'markets.swap: ERROR: user_action["action_type"] should be an allowed action for the model={self.pricing_model.model_name()}, not {user_action["action_type"]}!'
+                f'markets.swap: ERROR: agent_action.action_type should be an allowed action for the model={self.pricing_model.model_name()}, not {agent_action.action_type}!'
```
dpaiton, line 116: wrong function name? should be `markets.trade_and_update`

src/elfpy/markets.py
```diff
@@ -216,50 +294,71 @@ def update_spot_price(self):
             share_price=self.share_price,
             time_remaining=time_utils.stretch_time(self.token_duration, self.time_stretch_constant),
         )
+        self.rate = price_utils.calc_apr_from_spot_price(self.spot_price, self.token_duration)
+
+    def swap(self, trade_details):
+        if trade_details.direction == "in":
+            trade_results = self.pricing_model.calc_out_given_in(
+                in_                     = trade_details.trade_amount,
```
dpaiton, line 302: remove spaces around equals sign. Should be `var = value,` always

src/elfpy/markets.py
```diff
@@ -377,23 +470,148 @@ def _close_long(self, trade_details):
...
+            # update values used for next iteration
+            cum_liq += new_liq
+            cum_liq_by_agent[acting_agent] += new_liq
+            prev_time = current_time
+            prev_share_fees, prev_token_fees = share_fees_till_now, token_fees_till_now
+        if self.verbose:
+            print(f"cum_liq_by_agent = {cum_liq_by_agent}")
+            print(f"      share_owed = {share_owed} calculated sum = {sum(share_owed.values())}"
+                + f"\n                                           vs. direct = {self.share_fees} it's a "
+                + f"match 😁" if sum(share_owed.values()) == self.share_fees else f"mismatch 😢")
```
dpaiton, line 588: I do not want to include emojis in the terminal outputs.
1) it is a branding decision that we haven't discussed
2) it requires checking to ensure that it is supported across all terminal types for our end-users, which doesn't exist and we don't have time to make.

If you'd like to have the discussion, you should make a separate PR that adds emojis.

src/elfpy/markets.py
```diff
@@ -4,8 +4,12 @@
 TODO: rewrite all functions to have typed inputs
 """
 
-import numpy as np
 import elfpy.utils.time as time_utils
+from elfpy.user import AgentWallet
+from elfpy.utils.bcolors import bcolors
```
dpaiton, line 9: Have you tested that these colors work on different terminal types?
Is this addition helpful enough to justify maintenance and increased overhead with setting up logging & printing?

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+        elif agent_action.action_type == "close_short":  # buy PT to close short
+            agent_action.direction = "in"  # calcOutGivenIn
+            agent_action.token_in = "pt"  # buy known PT for unknown base
+            market_deltas, agent_deltas = self._close_short(agent_action)
+        elif agent_action.action_type == "add_liquidity":
+            # pricing model computes new market deltas
+            # market updates its "liquidity pool" wallet, which stores each trade's mint time and user address
+            # LP tokens are also storeds in user wallet as fungible amounts, for ease of user
+            market_deltas, agent_deltas = self._add_liquidity(agent_action)
+            pass
```
dpaiton, line 145: Why `pass`? you're executing an action, so you don't need this.

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+        elif agent_action.action_type == "remove_liquidity":
+            # market figures out how much the user has contributed (calcualtes their fee weighting)
+            # market resolves fees, adds this to the agent_action
+            # pricing model computes new market deltas
+            # market updates its "liquidity pool" wallet, which stores each trade's mint time and user address
+            # LP tokens are also storeds in user wallet as fungible amounts, for ease of user
+            # TODO: implement fee attribution and withdrawal
+            market_deltas, agent_deltas = self._remove_liquidity(agent_action)
+            pass
```
dpaiton, line 154: see above comment

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
-            raise ValueError(f'ERROR: Unknown trade type "{user_action["action_type"]}".')
+            raise ValueError(f'ERROR: Unknown trade type "{agent_action["action_type"]}".')
         # update market state
         self.update_market(market_deltas)
-        # TODO: self.update_LP_pool(wallet_deltas["fees"])
-        return wallet_deltas
-
+        self.update_spot_price_and_rate()
+        print(f"{self.get_market_step_string()}")
+        # TODO: self.update_LP_pool(agent_deltas["fees"])
```
dpaiton, line 161: This is resolved now, right? if so, remove the TODO

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+        elif agent_action.action_type == "open_short":  # sell PT to open short
+            agent_action.direction = "in"  # calcOutGivenIn
+            agent_action.token_out = "pt"  # sell known PT for unknown base
+            market_deltas, agent_deltas = self._open_short(agent_action)
+        elif agent_action.action_type == "close_short":  # buy PT to close short
+            agent_action.direction = "in"  # calcOutGivenIn
+            agent_action.token_in = "pt"  # buy known PT for unknown base
+            market_deltas, agent_deltas = self._close_short(agent_action)
+        elif agent_action.action_type == "add_liquidity":
+            # pricing model computes new market deltas
```
dpaiton, line 141: I think it would. make more sense for these comments to live inside the function docstring for the `_add_liquidity` class.

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+            # market updates its "liquidity pool" wallet, which stores each trade's mint time and user address
+            # LP tokens are also storeds in user wallet as fungible amounts, for ease of user
+            market_deltas, agent_deltas = self._add_liquidity(agent_action)
+            pass
+        elif agent_action.action_type == "remove_liquidity":
+            # market figures out how much the user has contributed (calcualtes their fee weighting)
```
dpaiton, line 147: I think it would. make more sense for these comments to live inside the function docstring for the `_remove_liquidity` class.

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
         self.update_market(market_deltas)
-        # TODO: self.update_LP_pool(wallet_deltas["fees"])
-        return wallet_deltas
-
+        self.update_spot_price_and_rate()
+        print(f"{self.get_market_step_string()}")
+        # TODO: self.update_LP_pool(agent_deltas["fees"])
+        agent_action.agent.update_wallet(agent_deltas)  # update user state since market doesn't know about users
+        if self.verbose:
+            print(f" agent  Δs ={agent_deltas.display()}\n market Δs ={market_deltas.display()}")
```
dpaiton, line 164: remove extra space before delta sign

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
-        self.cum_token_asset_fees += market_deltas["d_token_asset_fee"]
-        self.base_asset_orders += market_deltas["d_base_asset_orders"]
-        self.token_asset_orders += market_deltas["d_token_asset_orders"]
-        self.base_asset_volume += market_deltas["d_base_asset_volume"]
-        self.token_asset_volume += market_deltas["d_token_asset_volume"]
+        for field in market_deltas.__dataclass_fields__:
+            value = getattr(market_deltas, field)
+            if value: # check that it's instantiated
+                if field == "d_liquidity_pool_history":
+                    assert isinstance(value, list), f"markets.update_market: Error:"\
```
dpaiton, line 175: black prefers parenthesis -- run it to fix these.

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+                    + f" {field} has value={value} should be a dict"
+                elif field in ["d_share_fee_history","d_token_fee_history"]:
+                    assert isinstance(value, dict), f"markets.update_market: Error:"\
+                    + f" {field} has value={value} should be a dict"
+                else:
+                    assert np.isfinite(value), f"markets.update_market: ERROR: market delta key {field} is not finite."
+        self.share_reserves             += market_deltas.d_base_asset/self.share_price
```
dpaiton, line 182: remove excess spacing. single space around = or +=

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+        self.bond_reserves              += market_deltas.d_token_asset
+        self.share_buffer               += market_deltas.d_share_buffer
+        self.bond_buffer                += market_deltas.d_bond_buffer
+        self.liquidity_pool             += market_deltas.d_liquidity_pool
+        if market_deltas.d_liquidity_pool_history != []:
```
dpaiton, line 187: This conditional is strange. Are you checking if it is not a list, or checking if it is not empty? Rewrite with a more explicit conditional. For example, if you want to check if it is a list use `isinstance` and to check if it is empty use `len`.

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+            self.liquidity_pool_history.append(market_deltas.d_liquidity_pool_history)
+        self.share_fees                 += market_deltas.d_share_fee
+        if market_deltas.d_share_fee_history != {}:
```
dpaiton, line 190: see above comment on being more explicit with this conditional

src/elfpy/markets.py
```diff
@@ -67,67 +95,117 @@ def __init__(
...
+                else:
+                    self.share_fee_history[key] += value
+        self.token_fees                 += market_deltas.d_token_fee
+        if market_deltas.d_token_fee_history != {}:
+            for key, value in market_deltas.d_token_fee_history.items():
+                if key not in self.token_fee_history:
+                    self.token_fee_history.update(market_deltas.d_token_fee_history)
+                else:
+                    self.token_fee_history[key] += value
+        self.cum_base_asset_slippage    += market_deltas.d_base_asset_slippage
```
dpaiton, line 203: remove extra spacing around = and +=

src/elfpy/markets.py
```diff
@@ -207,7 +285,7 @@ def tick(self, delta_time):
         """Increments the time member variable"""
         self.time += delta_time
 
-    def update_spot_price(self):
+    def update_spot_price_and_rate(self):
```
dpaiton, line 288: can this live inside `update_market` instead of being its own function? In other words, are there times when we want to update the spot price & rate but do not want to update the other market variables?

src/elfpy/markets.py
```diff
@@ -216,50 +294,71 @@ def update_spot_price(self):
...
         # TODO: _in_protocol values should be managed by pricing_model and referenced by user
         max_loss = trade_details.trade_amount - output_with_fee
-        wallet_deltas = {
-            "base_in_wallet": -1 * max_loss,
-            "base_in_protocol": [trade_details.mint_time, max_loss],
-            "token_in_wallet": None,
-            "token_in_protocol": [trade_details.mint_time, trade_details.trade_amount],
-            "fee": [trade_details.mint_time, fee],
-        }
+        wallet_deltas = AgentWallet( # write out explicit signs, so it's clear what's happening
```
dpaiton, line 356: I don't think this comment is really necessary, it's clear what you're doing below.

src/elfpy/markets.py
```diff
@@ -216,50 +294,71 @@ def update_spot_price(self):
...
-            trade_details.trade_amount,
-            trade_details.share_reserves,
-            trade_details.bond_reserves,
-            trade_details.token_out,
-            trade_details.fee_percent,
-            trade_details.stretched_time_remaining,
-            trade_details.init_share_price,
-            trade_details.share_price,
+        without_fee_or_slippage, output_with_fee, output_without_fee, fee = self.swap(trade_details)
+        print(f"opening short: {without_fee_or_slippage, output_with_fee, output_without_fee, fee}")
```
dpaiton, line 343: This should be wrapped in a verbose statement

src/elfpy/markets.py
```diff
@@ -216,50 +294,71 @@ def update_spot_price(self):
             share_price=self.share_price,
             time_remaining=time_utils.stretch_time(self.token_duration, self.time_stretch_constant),
         )
+        self.rate = price_utils.calc_apr_from_spot_price(self.spot_price, self.token_duration)
+
+    def swap(self, trade_details):
```
dpaiton, line 299: Why does this function exist? It is only called in `_open_short` and the conditional is already fixed in `trade_and_update`, so it's unnecessary.

src/elfpy/markets.py
```diff
@@ -377,23 +470,148 @@ def _close_long(self, trade_details):
...
+        if self.verbose:
+            print(f"cum_liq_by_agent = {cum_liq_by_agent}")
+            print(f"      share_owed = {share_owed} calculated sum = {sum(share_owed.values())}"
+                + f"\n                                           vs. direct = {self.share_fees} it's a "
+                + f"match 😁" if sum(share_owed.values()) == self.share_fees else f"mismatch 😢")
+            print(f"      token_owed = {token_owed} calculated sum = {sum(token_owed.values())}"
+                + f"\n                                           vs. direct = {self.token_fees} it's a "
+                + f"match 😁" if sum(token_owed.values()) == self.token_fees else f"mismatch 😢")
+        dur = time.time() - start_time
+        print(f"calc_fees took ", end=f"{fmt(dur,precision=2)} seconds"\
```
dpaiton, line 593: This print should be in `if verbose`

src/elfpy/simulators.py
```diff
@@ -276,93 +292,48 @@ def run_simulation(self, override_dict=None):
...
+                self.market.trade_and_update(user_action)
+                self.update_analysis_dict()
+                self.run_trade_number += 1
+                self.last_user_action_time = self.market.time
+                # TODO: convert to proper logging
+        log_at_least_every_n_years = 0.1
+        if (self.market.time - self.last_user_action_time > log_at_least_every_n_years / 2) and (
+            self.market.time - self.last_user_action_time
+        ) % log_at_least_every_n_years <= 1 / 365 / self.config.simulator.num_blocks_per_day:
+            print(f"{self.market.get_market_step_string()} 😴" + f" {self.user_list[0].status_report()}")
```
dpaiton, line 336: only print if verbose

src/elfpy/pricing_models.py
```diff
@@ -27,14 +27,14 @@ class PricingModel:
     # TODO: set up member object that owns attributes instead of so many individual instance attributes
     # pylint: disable=too-many-instance-attributes
 
-    def __init__(self, verbose=None):
+    def __init__(self, verbose=False):
```
dpaiton, line 30: Please change this back to `None`. https://docs.python-guide.org/writing/gotchas/

src/elfpy/pricing_models.py
```diff
@@ -366,7 +369,6 @@ def calc_out_given_in(
             (out' - \Delta x) \cdot \phi, &\text{ if } token\_out = \text{"pt"}
             \end{cases} \\
             out = out' + f
-
```
dpaiton, line 369: we want a new line before the arguments are listed

src/elfpy/pricing_models.py
```diff
@@ -528,6 +530,7 @@ def calc_out_given_in(
         return (without_fee_or_slippage, with_fee, without_fee, fee)
 
 
+
```
dpaiton, line 533: remove extra newline

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        .. math::
+
+        y = \frac{(z - \Delta z)(\mu \cdot (\frac{1}{1 + r \cdot t(d)})^{\frac{1}{\tau(d_b)}} - c)}{2}
+
+        """
+        if self.verbose:
+            print(f"  inputs: base_asset_in={base_asset_in}, share_reserves={share_reserves}, bond_reserves={bond_reserves}, share_buffer={share_buffer}, init_share_price={init_share_price}, share_price={share_price}, liquidity_pool={liquidity_pool}, rate={rate}, time_remaining={time_remaining}, stretched_time_remaining={stretched_time_remaining}")
+        d_share_reserves = base_asset_in / share_price
+        if self.verbose:
```
dpaiton, line 597: why is this here?

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+            print(f"  d_share_reserves={d_share_reserves} (base_asset_in / share_price = {base_asset_in} / {share_price})")
+        if share_reserves > 0:  # normal case where we have some share reserves
+            lp_out = d_share_reserves * liquidity_pool / (share_reserves - share_buffer)
+        else:  # initial case where we have 0 share reserves
+            lp_out = d_share_reserves
+        d_token_reserves = (share_reserves + d_share_reserves) / 2 * (
+            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
+        ) - bond_reserves
+        if self.verbose:
```
dpaiton, line 606: can you combine all of the verbose prints per function? and remove debugging prints

src/elfpy/pricing_models.py
```diff
@@ -722,6 +899,10 @@ def calc_in_given_out(
             # This can also be expressed as:
             #
             # fee = ((1 / p) - 1) * φ * c * d_z
+            print(f"fee = ((1 / spot_price) - 1) * fee_percent * share_price * d_shares = "
```
dpaiton, line 902: can you put this in a verbose statement?

src/elfpy/simulators.py
```diff
@@ -177,56 +163,69 @@ def reset_rng(self, rng):
...
+                if key == "vault_apy":  # support for float or list[float] types
+                    if isinstance(value, float):
+                        self.config.simulator.vault_apy = [value] * self.config.simulator.num_trading_days
+                    else:
+                        assert len(value) == self.config.simulator.num_trading_days, (
+                            "vault_apy must have len equal to num_trading_days = "
+                            + f"{self.config.simulator.num_trading_days},"
+                            + f" not {len(value)}"
+                        )
+                print(f" overridding {key} = {getattr(self, key) if hasattr(self,key) else ''} with {value}")
```
dpaiton, line 194: wrap this in verbose conditional

src/elfpy/simulators.py
```diff
@@ -276,93 +292,48 @@ def run_simulation(self, override_dict=None):
...
-                    self.vault_apy[self.day]  # current day's apy
+                    self.config.simulator.vault_apy[self.day]  # current day's apy
                     / 365  # convert annual yield to daily
                     * self.market.init_share_price  # APR, apply return to starting price (no compounding)
                     # * self.market.share_price # APY, apply return to latest price (full compounding)
                 )
             for daily_block_number in range(self.config.simulator.num_blocks_per_day):
+                if self.day == self.config.simulator.num_trading_days - 1:
+                    if daily_block_number == self.config.simulator.num_blocks_per_day - 1:
+                        last_block_in_sim = True
```
dpaiton, line 311: this might be more clear if you say `last_block_in_sim = (daily_block_number == self.config.simulator.num_blocks_per_day - 1)`

src/elfpy/strategies/LP_and_withdraw.py
```diff
@@ -0,0 +1,36 @@
...
+                )
+            )
+        elif self.has_LPd:
+            enough_time_has_passed = self.market.time > self.time_to_withdraw
+            if enough_time_has_passed:
+                self.create_user_action(
+                        action_type="remove_liquidity",
+                        trade_amount=self.wallet.lp_in_wallet
+                )
+        return action_list
```
dpaiton, line 36: add newline

src/elfpy/strategies/LP_and_withdraw.py
```diff
@@ -0,0 +1,36 @@
+from elfpy.strategies.basic import BasicPolicy
+
+
+class Policy(BasicPolicy):
+    """
+    simple LP
+    only has one LP open at a time
+    """
+
+    def __init__(self, market, rng, wallet_address, budget=1000, amount_to_LP=100, verbose=False):
```
dpaiton, line 10: Why is `amount_to_lp` included in the constructor? I think elsewhere we were setting user variables in the init itself, but not in the constructor.

src/elfpy/strategies/LP_and_withdraw.py
```diff
@@ -0,0 +1,36 @@
...
+        """call basic policy init then add custom stuff"""
+        self.amount_to_LP = amount_to_LP
+        self.time_to_withdraw = rng.uniform(0.5, 1.5)
```
dpaiton, line 13: I would move this to be after the `super().__init__` and use `self.rng`

src/elfpy/strategies/basic.py
```diff
@@ -13,6 +13,9 @@ class BasicPolicy(User):
     """
     most basic policy setup
     """
+    def __init__(self, market, rng, wallet_address, budget=1000, verbose=False):
```
dpaiton, line 16: Don't need this at all if the inits are the same.

src/elfpy/strategies/simple_LP.py
```diff
@@ -0,0 +1,29 @@
+from elfpy.strategies.basic import BasicPolicy
+
+
+class Policy(BasicPolicy):
```
dpaiton, line 4: Is this agent type necessary? What purpose does it serve that `init_LP` and `LP_and_withdraw` does not?

src/elfpy/user.py
```diff
@@ -3,17 +3,24 @@
 
 TODO: rewrite all functions to have typed inputs
 """
+from elfpy.utils.basic_dataclass import *
```
dpaiton, line 6: don't import `*` anywhere.

src/elfpy/user.py
```diff
@@ -23,54 +30,92 @@ class User:
...
+                    output_string += f" {key}: "
+                    if value<2:
+                        output_string += f"{value:.5f}"
+                    elif value<100:
+                        output_string += f"{value:.2f}"
+                    else:
+                        output_string += f"{value:,.0f}"
+            print(output_string)
+
+        def update_market_variables(self, market, wallet_address=None):
```
dpaiton, line 87: Is there a better way to do this? The user has access to self.market. So as long as you're careful about passing `user.market` as an arg, you should automatically have access to the latest market variables. I don't like that we now have a bunch of user member variables that are manually set to track market member variables.

src/elfpy/user.py
```diff
@@ -115,13 +164,17 @@ def get_trade(self):
...
         """
+        self.status_update()
         action_list = self.action()  # get the action list from the policy
         for action in action_list:  # edit each action in place
             if action.mint_time is None:
                 action.mint_time = self.market.time
             if action.action_type == "close_short":
                 action.token_in_protocol = self.wallet.token_in_protocol[action.mint_time]
                 action.base_in_protocol = self.wallet.base_in_protocol[action.mint_time]
+            if action.action_type == "close_short":
```
dpaiton, line 175: why is this conditional listed twice now?

src/elfpy/user.py
```diff
@@ -133,44 +186,129 @@ def get_trade(self):
         return action_list
 
     def update_spend(self):
```
dpaiton, line 188: add function comments, remove commented prints

src/elfpy/user.py
```diff
@@ -133,44 +186,129 @@ def get_trade(self):
...
+                for mint_time, position in self.wallet.token_in_protocol.items():
+                    if self.verbose:
+                        print(f"  liquidate() evaluating closing short: mint_time={mint_time} position={position}")
+                    if position < 0:
+                        action_list.append(self.create_user_action(
+                                action_type="close_short",
+                                trade_amount= -position,
+                                mint_time=mint_time,
+                        ))
+        # if self.verbose:
```
dpaiton, line 252: remove commented code if it is not important

src/elfpy/pricing_models.py
```diff
@@ -101,6 +101,9 @@ def calc_spot_price_from_reserves(
         """
         total_reserves = share_price * share_reserves + bond_reserves
         bond_reserves_ = bond_reserves + total_reserves
+        # if share_reserves == 0:
+        #     return np.inf
+        # else:
```
jalextowle, line 106: Would be better to just assert that the reserves aren't zero.

src/elfpy/pricing_models.py
```diff
@@ -240,6 +242,7 @@ def calc_in_given_out(
         # share price:
         #
         # k = x**(1 - τ) + (2y + x)**(1 - τ)
```
jalextowle, line 244: Can you just remove this comment?

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
     def model_name(self):
         return "Hyperdrive"
 
+    def calc_lp_out_given_tokens_in(
+        self,
+        base_asset_in,
```
jalextowle, line 547: Stylistically, `d_base` would be more consistent.

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        d_share_reserves = base_asset_in / share_price
+        if self.verbose:
+            print(f"  d_share_reserves={d_share_reserves} (base_asset_in / share_price = {base_asset_in} / {share_price})")
+        if share_reserves > 0:  # normal case where we have some share reserves
+            lp_out = d_share_reserves * liquidity_pool / (share_reserves - share_buffer)
+        else:  # initial case where we have 0 share reserves
+            lp_out = d_share_reserves
+        d_token_reserves = (share_reserves + d_share_reserves) / 2 * (
+            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
+        ) - bond_reserves
```
jalextowle, line 605: The `- bond_reserves` should be inside the parenthesis.

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        Computes the amount of LP tokens to be minted for a given amount of base asset
+        
+        .. math::
+
+        y = \frac{(z - \Delta z)(\mu \cdot (\frac{1}{1 + r \cdot t(d)})^{\frac{1}{\tau(d_b)}} - c)}{2}
+
+        """
+        d_share_reserves = base_asset_out / share_price
+        lp_in = d_share_reserves * liquidity_pool / (share_reserves - share_buffer)
+        d_token_reserves = (share_reserves + d_share_reserves) / 2 * (
```
jalextowle, line 661: Nit: Should be `d_bonds`

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        )
+        r"""
+        Computes the amount of LP tokens to be minted for a given amount of base asset
+        
+        .. math::
+
+        y = \frac{(z - \Delta z)(\mu \cdot (\frac{1}{1 + r \cdot t(d)})^{\frac{1}{\tau(d_b)}} - c)}{2}
+
+        """
+        d_share_reserves = base_asset_out / share_price
```
jalextowle, line 659: Nit: should be `d_shares`

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
+        ) - bond_reserves
+        if self.verbose:
+            print(f"  lp_out={lp_out} (d_share_reserves * liquidity_pool / (share_reserves - share_buffer) = {d_share_reserves} * {liquidity_pool} / ({share_reserves} - {share_buffer}))")
+            print(f"  d_token_reserves={d_token_reserves} ((share_reserves + d_share_reserves) / 2 * (init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price) - bond_reserves = ({share_reserves} + {d_share_reserves}) / 2 * ({init_share_price} * (1 + {rate} * {time_remaining}) ** (1 / {stretched_time_remaining}) - {share_price}) - {bond_reserves})")
+        return lp_out, base_asset_in, d_token_reserves
+
+    def calc_lp_in_given_tokens_out(
+        self,
+        base_asset_out,
```
jalextowle, line 613: Nit: Should be `d_base`

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        .. math::
+
+        y = \frac{(z - \Delta z)(\mu \cdot (\frac{1}{1 + r \cdot t(d)})^{\frac{1}{\tau(d_b)}} - c)}{2}
+
+        """
+        d_share_reserves = base_asset_out / share_price
+        lp_in = d_share_reserves * liquidity_pool / (share_reserves - share_buffer)
+        d_token_reserves = (share_reserves + d_share_reserves) / 2 * (
+            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
+        ) - bond_reserves
```
jalextowle, line 663: Put the `- bond_reserves` in the parenthesis.

src/elfpy/pricing_models.py
jalextowle, line 661: Should be `share_reserves - d_shares`

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        assert (
+            stretched_time_remaining >= 0
+        ), f"pricing_models.calc_lp_out_given_tokens_in: ERROR: expected stretched_time_remaining >= 0, not {stretched_time_remaining}!"
+        assert share_price >= init_share_price >= 1, (
+            "pricing_models.calc_lp_out_given_tokens_in: ERROR: expected share_price >= init_share_price >= 1, not"
+        )
+        if self.verbose:
+            print(f"  inputs: lp_in={lp_in}, share_reserves={share_reserves}, bond_reserves={bond_reserves}, share_buffer={share_buffer}, init_share_price={init_share_price}, share_price={share_price}, liquidity_pool={liquidity_pool}, rate={rate}, time_remaining={time_remaining}, stretched_time_remaining={stretched_time_remaining}")
+        d_base_reserves = share_price * (share_reserves - share_buffer) * lp_in / liquidity_pool
+        d_share_reserves = d_base_reserves / share_price
```
jalextowle, line 709: Please use `d_base` and `d_shares`.

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+
+    def calc_lp_in_given_tokens_out(
+        self,
+        base_asset_out,
+        share_reserves,
+        bond_reserves,
+        share_buffer,
+        init_share_price,
+        share_price,
+        liquidity_pool,
```
jalextowle, line 619: Nit: Maybe `lp_reserves`?

src/elfpy/pricing_models.py
```diff
@@ -539,6 +542,180 @@ class HyperdrivePricingModel(PricingModel):
...
+        )
+        if self.verbose:
+            print(f"  inputs: lp_in={lp_in}, share_reserves={share_reserves}, bond_reserves={bond_reserves}, share_buffer={share_buffer}, init_share_price={init_share_price}, share_price={share_price}, liquidity_pool={liquidity_pool}, rate={rate}, time_remaining={time_remaining}, stretched_time_remaining={stretched_time_remaining}")
+        d_base_reserves = share_price * (share_reserves - share_buffer) * lp_in / liquidity_pool
+        d_share_reserves = d_base_reserves / share_price
+        if self.verbose:
+            print(f"  d_share_reserves={d_share_reserves} (d_base_reserves / share_price = {d_base_reserves} / {share_price})")
+        d_token_reserves = (share_reserves + d_share_reserves) / 2 * (
+            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
+        ) - bond_reserves
```
jalextowle, line 714: Couple things:

1. `d_token_reserves` => `d_bonds`.
2. `(share_reserves + d_shares) / 2` should be `(share_reserves - d_shares) / 2` since the shares are leaving the system.
3. `- bond_reserves` must be within the parenthesis. 

Finally (and this applies everywhere in this file), I'd be more comfortable with putting parenthesis around things like `(share_reserves + d_shares) / 2` so that we don't accidentally run into an order of operations problem.

src/elfpy/pricing_models.py
```diff
@@ -849,37 +1040,41 @@ def calc_out_given_in(
...
         # We precompute the YieldSpace constant k using the current reserves and
         # share price:
         #
-        # k = (c / μ) * (μ * z)**(1 - τ) + (2y + cz)**(1 - τ)
         k = price_utils.calc_k_const(share_reserves, bond_reserves, share_price, init_share_price, time_elapsed)
         if token_out == "base":
-            d_bonds = in_
-            in_reserves = bond_reserves + total_reserves
+            d_bonds = in_  # PTs passed in
+            in_reserves = bond_reserves + total_reserves  # add virtual liquidity
```
jalextowle, line 1077: If we're going to add these comments, let's add them in all if-clauses in the pricing model with this characteristic.

src/elfpy/utils/basic_dataclass.py
```diff
@@ -0,0 +1,44 @@
+from dataclasses import dataclass
+from dataclasses import field
+from elfpy.utils.fmt import *   # float→str formatter, also imports numpy as np
+
+@dataclass
+class BasicDataclass:
```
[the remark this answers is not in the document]
dpaiton: Looking at this now.

src/elfpy/strategies/LP_and_withdraw.py
MazyGio: We were indeed doing that (e.g. look at single_short.py and single_long.py), but I feel like being able to set the trade amounts in the constructor makes individual tests a bit more flexible. Wdyt?

src/elfpy/markets.py
wakamex: no it's just the color red

wakamex: they're not very descriptive. one is .WARNING (orange, i think)

src/elfpy/user.py
MazyGio: Removed the repeated conditional block

src/elfpy/markets.py
wakamex: haha, imagine that PR 🙃
they're nice short-hand (i like making everything short)
but agree it was a pain in the ass to make them show up properly

src/elfpy/strategies/LP_and_withdraw.py
dpaiton: Yeah I think that makes sense. I am worried that you shouldn't add variables to subclasses on __init__, it kind of defeats the point of the parent class. We use those to establish an API for downstream implementations, but if we're always overwriting the API then it is an anti-pattern. I think you'd be better to limit the number of named args to an essential subset, and then use `**kwargs` for the rest.

src/elfpy/strategies/simple_LP.py
dpaiton: necessary for tests

src/elfpy/strategies/LP_and_withdraw.py
wakamex: leaving as is for now, will consider improving in future PR

src/elfpy/user.py
wakamex: 9 were passthrough assignments, moved those to where they're called in the _long/_short functions. 3 pieces of logic i moved out to where the function was.

src/elfpy/user.py
wakamex: keeping commented prints, so we can add them back in w/ attached logging TODO:

src/elfpy/markets.py
```diff
@@ -51,83 +106,154 @@ def __init__(
...
+            agent_action.direction = "in"  # calcOutGivenIn
+            agent_action.token_in = "pt"  # buy known PT for unknown base
+            market_deltas, agent_deltas = self._close_short(agent_action)
+        elif agent_action.action_type == "add_liquidity":
+            market_deltas, agent_deltas = self._add_liquidity(agent_action)
+        elif agent_action.action_type == "remove_liquidity":
+            market_deltas, agent_deltas = self._remove_liquidity(agent_action)
         else:
-            raise ValueError(f'ERROR: Unknown trade type "{user_action["action_type"]}".')
+            raise ValueError(f'ERROR: Unknown trade type "{agent_action["action_type"]}".')
```
dpaiton, line 195: I don't think this syntax will work because the `UserAction` dataclass does not have the required dunder methods.

src/elfpy/markets.py
```diff
@@ -320,40 +432,37 @@ def _open_long(self, trade_details):
         #        use current mint time because this is a fresh
         trade_results = self.pricing_model.calc_out_given_in(
             trade_details.trade_amount,
-            trade_details.share_reserves,
-            trade_details.bond_reserves,
+            trade_details.market.share_reserves,
```
dpaiton, line 435: Is this right? Should this not be `self.share_reserves`, etc? I'm not sure, just checking that you checked.

src/elfpy/markets.py
```diff
@@ -363,37 +472,185 @@ def _close_long(self, trade_details):
...
+        )
+        return market_deltas, agent_deltas
+
+    def _add_liquidity(self, trade_details):
+        """
+        Computes new deltas for bond & share reserves after liquidity is added
+        """
+        lp_out, d_base_reserves, d_token_reserves = self.pricing_model.calc_lp_out_given_tokens_in(
+            base_asset_in=trade_details.trade_amount,
+            share_reserves=trade_details.share_reserves,
```
dpaiton, line 512: Should these be `trade_details.market.share_reserves`, etc? or `self.share_reserves`, as mentioned in the other comment?

src/elfpy/pricing_models.py
```diff
@@ -99,8 +99,12 @@ def calc_spot_price_from_reserves(
         float
             The spot price of principal tokens.
         """
+        assert(share_reserves > 0, f"pricing_models.calc_spot_price_from_reserves: ERROR: expected share_reserves > 0, not {share_reserves}!")
         total_reserves = share_price * share_reserves + bond_reserves
         bond_reserves_ = bond_reserves + total_reserves
+        # if share_reserves == 0:
```
dpaiton, line 105: Can you delete these, I don't think they're needed?

src/elfpy/strategies/simple_short.py
```diff
@@ -0,0 +1,42 @@
+# pylint: disable=duplicate-code
```
dpaiton, line 1: Can you rename this file so that it is meaningfully different from `single_short.py`?

src/elfpy/user.py
```diff
@@ -4,73 +4,134 @@
...
-        token_in_wallet: dict = field(default_factory=dict)
-        base_in_protocol: dict = field(default_factory=dict)
-        token_in_protocol: dict = field(default_factory=dict)
-
-        def __getitem__(self, key):
-            return getattr(self, key)
-
-        def __setitem__(self, key, value):
+        self.position_list = []
+        for key, value in kwargs.items():
```
dpaiton, line 80: I added this and the `**kwargs` in the signature, but it is a really bad coding pattern (creates easy opportunity for attacks, also v opaque in terms of what args are accepted). I don't think we need them, so can you delete?

src/elfpy/utils/float_to_string.py
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        digits = int(np.floor(np.log10(abs(value)))) + 1  #  calculate number of digits in value
+    except:
+        if debug:
+            print(
+                f"Error in float_to_string: value={value}({type(value)}), precision={precision},"
+                f" min_digits={min_digits}"
+            )
+        return str(value)
+    decimals = min(max(precision - digits, min_digits), precision)  #  calculate desired decimals
```
dpaiton, line 23: I think you can resolve this with [np.around](https://numpy.org/doc/stable/reference/generated/numpy.around.html#numpy.around)?

src/elfpy/utils/parse_config.py
```diff
@@ -75,4 +77,5 @@ def parse_simulation_config(config_file):
         market=MarketConfig(**toml_config["market"]),
         amm=AMMConfig(**toml_config["amm"]),
         simulator=SimulatorConfig(**toml_config["simulator"]),
+        pricing_model=PricingModelConfig(**toml_config["pricing_model"]),
```
dpaiton, line 80: You removed `pricing_model`, right?

tests/test_sequential.py
```diff
@@ -0,0 +1,35 @@
+# pylint: disable=duplicate-code
+""" Manually test things in a sequence """
```
dpaiton, line 2: Can you make this more descriptive? I'm not sure what is going on.

dpaiton:
Nearly there! Here's what I see that is left:

- [ ] Remove `hyperdrive_config.toml` & rename the other one to `example_config.toml`; update tests & notebooks accordingly
- [ ]  Get the tests to pass as best as possible. Some are not expected to, but those should be commented out. The remaining, including some that are new to this PR, should pass.
- [ ] Resolve remaining comments
- [ ] `git pull --rebase origin dev`; then walk through the rebase; then `git push -f origin add_lpers`

src/elfpy/simulators.py
```diff
@@ -236,22 +239,43 @@ def setup_simulated_entities(self, override_dict=None):
...
+                f"\n vault_apy = {self.config.simulator.vault_apy[0]:.2%}"
+                f"\n fee_percent = {self.market.fee_percent}"
+                f"\n share_price = {self.market.share_price}"
+                f"\n init_share_price = {self.market.init_share_price}"
+                f"\n init_time_stretch = {self.market.time_stretch_constant}"
+            )
+            print(f"{self.market.get_market_step_string()}")
+        self.block_number = 0
+        print(f" should I initialize the market with a user? init_LP = {self.config.simulator.init_LP}")
+        if self.config.simulator.init_LP:
```
dpaiton, line 256: Is the init_LP flag meant to allow support for backwards compatibility with the old method of initializing the market? I don't think we really need to support this. Both pricing models should be compatible with the new setup.

dpaiton (approved):

PR #61, opened 2022-11-30
sim tests

Padraic-O-Mhuiris:


src/elfpy/pricing_models.py
```diff
@@ -539,6 +539,58 @@ def calc_out_given_in(
...
+        bond_reserves,
+        share_buffer,
+        init_share_price,
+        share_price,
+        lp_reserves,
+        rate,
+        time_remaining,
+        stretched_time_remaining,
+    ):
+        return HyperdrivePricingModel().calc_lp_out_given_tokens_in(
```
dpaiton, line 555: I'm not sure that these would be the same. If they are, then this function should live in the root class (`PricingModel`).

Padraic-O-Mhuiris:
image

Current status is that there are two "sims" which will run 15x

src/elfpy/pricing_models.py
Padraic-O-Mhuiris: This has been removed

tests/test_sim.py
```diff
@@ -6,386 +6,65 @@
...
+        simulator = YieldSimulator(
+            apply_config_logging(
+                Config(
+                    market=MarketConfig(),
+                    amm=AMMConfig(),
+                    simulator=SimulatorConfig(pricing_model_name="Hyperdrive"),
+                )
+            )
+        )
+        for rng_index in range(1, 15):
```
dpaiton, line 49: `rng_seed` is probably a more accurate name

tests/test_sim.py
```diff
@@ -6,386 +6,65 @@
...
-# )
-# sim_days_remaining = pricing_model.time_to_days_remaining(market.time_remaining, time_stretch)
-# market_apy = market.apy(sim_days_remaining)
-# np.testing.assert_allclose(calculated_apy, market_apy)
+import logging
+import unittest
+import numpy as np
+import os, sys
+
+sys.path.insert(1, os.path.join(os.getcwd(), "src"))
```
dpaiton: remove this -- not needed if you use `pip install -e`

PR #63, opened 2022-12-01
improved testing

wakamex:
**major changes**:
- fix get_max_pt_short
- print market_step only if executed trades
- pass kwargs to agents to allow arbitrary parameters passed in testing
- add testing to verify that custom parameters are working as intended
- make deleting log file optional with a parameter

**fixes**:
- fix bug that initialized mint_time to 0 instead of None which resulted in wrong time_remaining calculation
- fix close_short accounting not receiving sale proceeds
- revert previous incorrect attempts to fix max short
- add handler close statements to avoid file permission errors in windows

**minor changes**:
- remove float_to_string since it's not useful for logging (only for printing to terminal)
- remove fees_paid since it isn't needed
- remove effective_price since it's informative only, can be calculated post-sim
- rename time_remaining to stretched_time_remaining to clarify what should be passed in
- additional comments to clarify what units are being passed in
- rename amount_to_lp to base_to_lp
- few additional logging statements and clarifications

### DIFF
--- a/src/elfpy/agent.py
+++ b/src/elfpy/agent.py
@@ -3,13 +3,12 @@ Implements abstract classes that control agent behavior
 """
 
 import logging
-
 import numpy as np
 
+import elfpy.utils.time as time_utils
 from elfpy.markets import Market
 from elfpy.pricing_models.base import PricingModel
 from elfpy.types import MarketAction, MarketActionType
-from elfpy.utils.outputs import float_to_string
 from elfpy.wallet import Wallet
 
 
@@ -19,7 +18,7 @@ class Agent:
     date value is an inte with how many tokens they have for that date
     """
 
-    def __init__(self, wallet_address: int, budget: float):
+    def __init__(self, market: Market, rng: Generator, wallet_address: int, budget: float = 1000, **kwargs):
         """
         Set up initial conditions
         """
@@ -29,9 +28,13 @@ class Agent:
         self.last_update_spend: float = 0  # timestamp
         self.product_of_time_and_base: float = 0
         self.wallet: Wallet = Wallet(address=wallet_address, base=budget)
+        for key, value in kwargs.items():
+            if value and hasattr(self, key):
+                logging.info("setting agent's %s to %s", key, str(value))
+                setattr(self, key, value)
 
     def create_agent_action(
-        self, action_type: MarketActionType, trade_amount: float, mint_time: float = 0
+        self, action_type: MarketActionType, trade_amount: float, mint_time: float = None
     ) -> MarketAction:
         """Instantiate a agent action"""
         agent_action = MarketAction(
@@ -59,10 +62,20 @@ class Agent:
         """
         if market.market_state.share_reserves == 0:
             return 0
-        max_pt_short = (
-            market.market_state.share_reserves * market.market_state.share_price / market.get_spot_price(pricing_model)
+        time_remaining = time_utils.get_yearfrac_remaining(market.time, market.time, market.token_duration)
+        stretched_time_remaining = time_utils.stretch_time(time_remaining, market.time_stretch_constant)
+        trade_results = pricing_model.calc_in_given_out(
+            market.share_reserves,
+            market.share_reserves,
+            market.bond_reserves,
+            "pt",
+            market.fee_percent,
+            stretched_time_remaining,
+            market.init_share_price,
+            market.share_price,
         )
-        return max_pt_short
+        output_with_fee = trade_results[1]
+        return output_with_fee
 
     def get_trade_list(self, market: Market, pricing_model: PricingModel) -> list:
         """
@@ -77,11 +90,18 @@ class Agent:
         we spend what we have to spend, and get what we get.
         """
         action_list = self.action(market, pricing_model)  # get the action list from the policy
+        # TODO: is this the right place to modify an action's mint_time? why not in __post_init__?
         for action in action_list:  # edit each action in place
             if action.mint_time is None:
                 action.mint_time = market.time
         # TODO: Add safety checks
         # e.g. if trade amount > 0, whether there is enough money in the account
+        # if len(trade_action) > 0: # there is a trade
+        #    token_in, token_out, trade_amount_usd = trade_action
+        #    assert trade_amount_usd >= 0, (
+        #        f"agent.py: ERROR: Trade amount should not be negative, but is {trade_amount_usd}"
+        #        f" token_in={token_in} token_out={token_out}"
+        #    )
         return action_list
 
     def update_wallet(self, wallet_deltas: Wallet, market: Market) -> None:
@@ -94,10 +114,10 @@ class Agent:
             if value_or_dict is None:
                 pass
             # handle updating a value
-            if key in ["base", "lp_tokens", "fees_paid"]:
+            if key in ["base", "lp_tokens"]:
                 if value_or_dict != 0 or self.wallet[key] != 0:
-                    logging.debug(
-                        "agent #%g %s pre-trade = %.0g\npost-trade = %1g\ndelta = %1g",
+                    logging.info(
+                        "agent #%03.0f %s pre-trade = %.0g\npost-trade = %1g\ndelta = %1g",
                         self.wallet_address,
                         key,
                         self.wallet[key],
@@ -108,8 +128,8 @@ class Agent:
             # handle updating a dict, which have mint_time attached
             elif key in ["margin", "longs", "shorts"]:
                 for mint_time, amount in value_or_dict.items():
-                    logging.debug(
-                        "agent #%g trade %s, mint_time = %g\npre-trade amount = %s\ntrade delta = %s",
+                    logging.info(
+                        "agent #%03.0f trade %s, mint_time = %g\npre-trade amount = %s\ntrade delta = %s",
                         self.wallet_address,
                         key,
                         mint_time,
@@ -120,8 +140,6 @@ class Agent:
                         self.wallet[key][mint_time] += amount
                     else:
                         self.wallet[key].update({mint_time: amount})
-            elif key in ["fees_paid", "effective_price"]:
-                pass
             elif key in ["address"]:
                 pass
             else:
@@ -152,11 +170,10 @@ class Agent:
 
     def log_status_report(self) -> None:
         """Logs the current user state"""
-        logging.debug(
-            "agent %g base = %1g and fees_paid = %1g",
+        logging.info(
+            "agent #%03.0f base = %s",
             self.wallet_address,
             self.wallet.base,
-            self.wallet.fees_paid if self.wallet.fees_paid else 0,
         )
 
     def log_final_report(self, market: Market, pricing_model: PricingModel) -> None:
@@ -182,19 +199,18 @@ class Agent:
         lost_or_made = "lost" if profit_and_loss < 0 else "made"
         logging.info(
             (
-                "agent #%g %s %s on $%s spent, APR = %g"
-                " (%.2g in %s years), net worth = $%s"
-                " from %s base and %s tokens at p = %g\n"
+                "agent #%03.0f %s %s on $%s spent, APR = %g (%.2g in %s years), "
+                "net worth = $%s from %s base and %s tokens at p = %g\n"
             ),
             self.wallet_address,
             lost_or_made,
-            float_to_string(profit_and_loss),
-            float_to_string(spend),
+            profit_and_loss,
+            spend,
             annual_percentage_rate,
             holding_period_rate,
-            float_to_string(market.time, precision=2),
-            float_to_string(worth),
-            float_to_string(base),
-            float_to_string(tokens),
+            market.time,
+            worth,
+            base,
+            tokens,
             price,
         )
--- a/src/elfpy/pricing_models.py
+++ b/src/elfpy/pricing_models.py
@@ -1,0 +1,1401 @@
+"""
+Pricing models implement automated market makers (AMMs)
+"""
+
+from abc import ABC, abstractmethod
+from typing import NamedTuple
+import logging
+
+from elfpy.types import TokenType
+import elfpy.utils.price as price_utils
+import elfpy.utils.time as time_utils
+
+# TODO: Currently many functions use >5 arguments.
+# These should be packaged up into shared variables, e.g.
+#     reserves = (in_reserves, out_reserves)
+#     share_prices = (init_share_price, share_price)
+# pylint: disable=too-many-arguments
+
+# TODO: This module is too big, we should break it up into pricing_models/{base.py, element.py, hyperdrive.py}
+# pylint: disable=too-many-lines
+
+# TODO: some functions have too many local variables (15 is recommended max),
+#     we should consider how to break them up or delete this TODO if it's not possible
+# pylint: disable=too-many-locals
+
+
+class TradeResult(NamedTuple):
+    """
+    Result from a calc_out_given_in or calc_in_given_out.  The values are the amount of asset required
+    for the trade, either in or out.  Fee is the amount of fee collected, if any.
+    """
+
+    without_fee_or_slippage: float
+    with_fee: float
+    without_fee: float
+    fee: float
+
+
+class PricingModel(ABC):
+    """
+    Contains functions for calculating AMM variables
+
+    Base class should not be instantiated on its own; it is assumed that a user will instantiate a child class
+    """
+
+    # TODO: Change argument defaults to be None & set inside of def to avoid accidental overwrite
+    # TODO: set up member object that owns attributes instead of so many individual instance attributes
+    # pylint: disable=too-many-instance-attributes
+
+    @abstractmethod
+    def calc_in_given_out(
+        self,
+        out: float,
+        share_reserves: float,
+        bond_reserves: float,
+        token_in: TokenType,
+        fee_percent: float,
+        stretched_time_remaining: float,
+        init_share_price: float,
+        share_price: float,
+    ) -> TradeResult:
+        """Calculate fees and asset quantity adjustments"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def calc_out_given_in(
+        self,
+        in_: float,
+        share_reserves: float,
+        bond_reserves: float,
+        token_out: TokenType,
+        fee_percent: float,
+        time_remaining: float,
+        init_share_price: float,
+        share_price: float,
+    ) -> TradeResult:
+        """Calculate fees and asset quantity adjustments"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def calc_lp_out_given_tokens_in(
+        self,
+        d_base: float,
+        share_reserves: float,
+        bond_reserves: float,
+        share_buffer: float,
+        init_share_price: float,
+        share_price: float,
+        lp_reserves: float,
+        rate: float,
+        time_remaining: float,
+        stretched_time_remaining: float,
+    ) -> tuple[float, float, float]:
+        """
+        Computes the amount of LP tokens to be minted for a given amount of base asset"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def calc_lp_in_given_tokens_out(
+        self,
+        d_base: float,
+        share_reserves: float,
+        bond_reserves: float,
+        share_buffer: float,
+        init_share_price: float,
+        share_price: float,
+        lp_reserves: float,
+        rate: float,
+        time_remaining: float,
+        stretched_time_remaining: float,
+    ) -> tuple[float, float, float]:
+        """
+        Computes the amount of LP tokens to be minted for a given amount of base asset"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def calc_tokens_out_given_lp_in(
+        self,
+        lp_in: float,
+        share_reserves: float,
+        bond_reserves: float,
+        share_buffer: float,
+        init_share_price: float,
+        share_price: float,
+        lp_reserves: float,
+        rate: float,
+        time_remaining: float,
+        stretched_time_remaining: float,
+    ) -> tuple[float, float, float]:
+        """Calculate how many tokens should be returned for a given lp addition"""
+        raise NotImplementedError
+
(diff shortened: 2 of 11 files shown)


src/elfpy/agent.py
```diff
@@ -3,10 +3,10 @@
 """
 
 import logging
-
```
dpaiton, line 6: newline between stdlib & external imports is a good idea (black should enforce this?)

src/elfpy/agent.py
```diff
@@ -27,9 +32,14 @@ def __init__(self, market: Market, rng: Generator, wallet_address: int, budget:
         # TODO: remove this, wallet_address is a property of wallet, not the agent
         self.wallet_address: int = wallet_address
         self.budget: float = budget
+        self.verbose: bool = False if verbose is None else verbose
```
dpaiton, line 35: We've gotten rid of verbose in the logging PR.

dpaiton:
> pass kwargs to agents to allow arbitrary parameters passed in testing

Why is this a better pattern than assigning parameters in the config?

For example, we set simulation configs in the simulator by modifying the config object directly when it is called. For example,  in [test_sim](https://github.com/element-fi/elf-simulations/blob/main/tests/test_sim.py#L44) :
```
        simulator = YieldSimulator(
            Config(
                market=MarketConfig(),
                amm=AMMConfig(pricing_model_name="Hyperdrive"),
                simulator=SimulatorConfig(
                    num_trading_days=10,
                    num_blocks_per_day=10,
                    logging_level=logging.INFO,
                ),
            )
        )
```

For the user, we can do the same. Set user_list = [] on init, then add our own with args passed. e.g.:
```
simulator = YieldSimulator(config)
simulator.reset_rng(
    np.random.default_rng(simulator.config.simulator.random_seed)
)
simulator.set_random_variables()
simulator.setup_simulated_entities({"user_policies": []}) # no users at first
simulator.agents.update(
    {1: CustomUser(
            market=simulator.market,
            rng=simulator.rng,
            wallet_address=1
        )
    }
)
simulator.run_simulation()
```

CustomUser would be specified elsewhere, e.g. from a dictionary in the notebook or a toml file. 

This functionality already exists in main. So I'm thinking it's not really worth adding an alternative user pattern in your PR. I'm also weary because combining `setattr` with `kwargs` as you do is hard to parse. I didn't find a great explanation from a quick Google, but these two links get at the idea: [stack overflow answer](https://stackoverflow.com/a/13982449) and [some other weird forum answer](https://teamtreehouse.com/community/is-using-setattr-with-kwargs-a-good-idea). It's not like a terrible thing to do, but I'd rather avoid it if we can.

src/elfpy/agent.py
```diff
@@ -18,7 +18,12 @@ class Agent:
     date value is an inte with how many tokens they have for that date
     """
 
-    def __init__(self, market: Market, rng: Generator, wallet_address: int, budget: float):
+    # pylint: disable=too-many-instance-attributes
```
dpaiton, line 21: If you're going to add pylint disables, please also add descriptions of the problem and a path to fix them. We should avoid disabling these if possible.

src/elfpy/agent.py
```diff
@@ -64,8 +74,22 @@ def get_max_pt_short(self) -> float:
         """
         if self.market.share_reserves == 0:
             return 0
-        max_pt_short = self.market.share_reserves * self.market.share_price / self.market.get_spot_price()
-        return max_pt_short
+        time_remaining = time_utils.get_yearfrac_remaining(
```
dpaiton, line 77: Remove the TODO comment above as well please!

src/elfpy/agent.py
```diff
@@ -64,8 +74,22 @@ def get_max_pt_short(self) -> float:
...
+            self.market.time, self.market.time, self.market.token_duration
```
dpaiton, line 78: can you name these args, so that they're more clear. Since the values are the same.

src/elfpy/agent.py
```diff
@@ -109,7 +133,7 @@ def update_wallet(self, wallet_deltas: Wallet) -> None:
             if key in ["base_in_wallet", "lp_in_wallet", "fees_paid"]:
                 if value_or_dict != 0 or self.wallet[key] != 0:
                     logging.debug(
-                        "agent #%g %s pre-trade = %.0g\npost-trade = %1g\ndelta = %1g",
+                        "agent %03.0f %s pre-trade = %.0g\npost-trade = %1g\ndelta = %1g",
```
dpaiton, line 136: Why did you also remove the `#`?

src/elfpy/simulators.py
```diff
@@ -347,18 +362,25 @@ def collect_and_execute_trades(self, last_block_in_sim=False):
...
                     [key for key in self.agents if key > 0]  # exclude init_lp before shuffling
                 )
                 wallet_ids = np.append(wallet_ids, 0)  # add init_lp so that they're always last
-            else:
+            else:  # include init_lp only on the last block, to let it unwind
                 wallet_ids = self.rng.permutation(list(self.agents))
+        else:  # we are in a deterministic mode
+            # reverse the list excluding 0 (init_lp)
+            wallet_ids = [key for key in self.agents if key > 0][::-1]
+            if last_block_in_sim:  # prepend init_lp to the list
```
dpaiton, line 382: Why is the init_lp excluded for other blocks?

src/elfpy/simulators.py
```diff
@@ -347,18 +362,25 @@ def collect_and_execute_trades(self, last_block_in_sim=False):
         """Get trades from the agent list, execute them, and update states"""
         if not isinstance(self.market, Market):
             raise ValueError("market not defined")
+        number_of_executed_trades = 0
         # TODO: This is a HACK to prevent the initial LPer from rugging other agents.
         # The initial LPer should be able to remove their liquidity and any open shorts can still be closed.
         # But right now, if the LPer removes liquidity while shorts are open,
         # then closing the shorts results in an error (share_reserves == 0).
+        wallet_ids = [key for key in self.agents if key > 0]  # exclude init_lp before shuffling
```
dpaiton, line 370: Is this needed? I don't think it's used.

src/elfpy/simulators.py
```diff
@@ -376,10 +398,14 @@ def collect_and_execute_trades(self, last_block_in_sim=False):
                 agent.log_status_report()
                 self.update_analysis_dict()
                 self.run_trade_number += 1
+                number_of_executed_trades += 1
+        if number_of_executed_trades > 0:
+            logging.debug("executed %f trades at %s", number_of_executed_trades, self.market.get_market_state_string())
```
dpaiton, line 403: This adds a debug print for every trade, which will generate a massive amount of outputs. Maybe we should introduce a level higher than debug to warn users that they're getting hella outputs? It seems like there would be good argument for a debug output that is still viable with more than a few days, blocks per day, and trades per block.

src/elfpy/strategies/single_lp.py
```diff
@@ -10,29 +10,14 @@
...
-        """call basic policy init then add custom stuff"""
-        self.amount_to_lp = 100
-        super().__init__(
-            market=market,
-            rng=rng,
-            wallet_address=wallet_address,
-            budget=budget,
-        )
+    "single LP: only has one LP open at a time"
+    amount_to_lp = 100
```
dpaiton, line 14: This is now a static element that belongs to the class. Before, the variable was an object member variable (belongs to `self`). The difference is whether you want the variable to exist even if the class is not instantiated. What you've done here is not necessarily _wrong_, but it is conveying the wrong message. We want `amount_to_lp` to be belonged by some instantiated user. e.g.
```
from elfpy.strategies.single_lp import Policy
alice = Policy()
alice.amount_to_lp = 100
bob = Policy()
bob.amount_to_lp = 1000
```

so in this example `amount_to_lp` belongs to alice or bob, not to Policy. 

Long story short, you should put this back to how it was.

src/elfpy/strategies/single_lp.py
```diff
@@ -10,29 +10,14 @@
...
-        implement user strategy
-        LP if you can, but only do it once
-        """
-        action_list = []
         has_lp = self.wallet.lp_in_wallet > 0
         can_lp = self.wallet.base_in_wallet >= self.amount_to_lp
         if can_lp and not has_lp:
-            action_list.append(self.create_agent_action(action_type="add_liquidity", trade_amount=self.amount_to_lp))
-        return action_list
+            self.action_list.update(
```
dpaiton, line 20: It appears as though you deleted the docstring. Please put it back. Also, I think the action_list should not be a member variable of agent (mentioned in an earlier comment).

tests/test_logging.py
```diff
@@ -60,6 +60,7 @@ def test_logging(self):
             simulator.setup_simulated_entities(override_dict)
             simulator.run_simulation()
             self.assertLogs(level=level)
+            handler.close()
```
dpaiton, line 63: Why did you need to add this? I think I was getting errors from not having it, but I'm curious what lead you to figure it out?

tests/test_trade.py
```diff
@@ -122,3 +123,8 @@ def test_single_short(self):
     def test_base_lps(self):
         """Tests base LP setups"""
         self.run_base_lp_test(user_policies=["single_lp"], config_file="config/example_config.toml")
+
+
+if __name__ == "__main__":
```
dpaiton, line 128: I don't think we should commit tests with options to run as a script

tests/utils/test_time_utils.py
```diff
@@ -23,10 +23,13 @@ def test_current_datetime(self):
 
         now = datetime.datetime.now(pytz.timezone("Etc/GMT-0"))
         test_time = time_utils.current_datetime()
+        time_delta = datetime.timedelta(milliseconds=100)
```
dpaiton, line 26: This is testing something different now, right? Why did you make the change?

src/elfpy/agent.py
wakamex: looks like it doesn't, but i can add it

wakamex:
> Why is this a better pattern than assigning parameters in the config?
the only reason would be if it's easier for the user (which it may not be)

> CustomUser would be specified elsewhere, e.g. from a dictionary in the notebook or a toml file.
this sounds like additional boilerplate required for the user the understand and use. whereas adding `:parameter=value` inside a string they're already passing in, seems easier for the user?

> This functionality already exists in main.
it doesn't exist for agents, which is why i added this

> So I'm thinking it's not really worth adding an alternative user pattern in your PR
agree we shouldn't have two methods for the same thing

> I'm also weary because combining `setattr` with `kwargs` as you do is hard to parse.
it's definitely a bit hackish and hard to read. this is necessary to allow custom parameters to be created at init-time. related to this usage, i found the need to explicitly declare all member variables before this loop, to avoid making typos and setting a non-existent member variable that isn't going to be used (i made this mistake). that's messy, requiring each member variable to be declared here, 1 line for each variable. a cleaner version of that is using __slots__ to define all member variables ahead of time, in 1 line!!! while also enforcing stricter use of member variables (i assume you're a fan 😏). check this out:
```
    __slots__ = ["market", "rng", "wallet_address", "budget", "last_update_spend", "product_of_time_and_base",
                 "wallet", "action_list", "base_to_lp", "pt_to_short"]
```
making type on custom parameter definition in test_sequential.py: 
`ValueError: ERROR: Attempted to manually override non-existent Agent attribute based_to_lp`
making type trying to set a new member variable without running it by Dylan (he owns __slots__):
`haven't gotten this working yet, but my reading tells me it's possible` based on this: [https://wiki.python.org/moin/UsingSlots](https://wiki.python.org/moin/UsingSlots)
```
x1.slot1 = 1

    ---------------------------------------------------------------------------

    AttributeError                            Traceback (most recent call last)

    Input In [3], in <module>
    ----> 1 x1.slot1 = 1


    AttributeError: 'Example' object has no attribute 'slot1'
```

> Also I think the regular expression code added to the simulator would need to be pretty thoroughly tested to avoid weird inputs causing errors.
it's only used to chain together two splits. so it's not running an open-ended regular expression. just custom splitting. can probably be done without the regular expression library.

src/elfpy/pricing_models.py
```diff
@@ -931,7 +931,7 @@ def calc_in_given_out(
         bond_reserves: float,
         token_in: TokenType,
         fee_percent: float,
-        time_remaining: float,
+        stretched_time_remaining: float,
```
dpaiton, line 934: Good call -- I think this var name makes more sense.

src/elfpy/pricing_models.py
```diff
@@ -1006,32 +1006,32 @@ def calc_in_given_out(
...
         # We precompute the YieldSpace constant k using the current reserves and
         # share price:
         #
         # k = (c / μ) * (μ * z)**(1 - t) + (2y + cz)**(1 - t)
         k = price_utils.calc_k_const(share_reserves, bond_reserves, share_price, init_share_price, time_elapsed)
         if token_in == "base":
-            in_reserves = share_reserves
-            out_reserves = bond_reserves + total_reserves
-            d_bonds = out
+            in_reserves = share_reserves  # in is in units of shares
```
dpaiton, line 1032: Thanks for adding units!

src/elfpy/pricing_models.py
```diff
@@ -1322,7 +1323,16 @@ def calc_out_given_in(
             #
             # fee = (1 - p) * φ * d_y
             fee = (1 - spot_price) * fee_percent * d_bonds
+
+            # To get the amount paid with fees, subtract the fee from the
```
dpaiton, line 1327: Super clear -- thanks for adding this

dpaiton, line 1327: consistent commenting 🤌

src/elfpy/simulators.py
```diff
@@ -280,11 +280,26 @@ def setup_simulated_entities(self, override_dict=None):
             self.agents = {}
         self.market.log_market_step_string()
         # continue adding other users
-        for policy_number, policy_name in enumerate(self.config.simulator.user_policies):
+        for policy_number, policy_instruction in enumerate(self.config.simulator.user_policies):
+            if ":" in policy_instruction:  # we have custom parameters
```
dpaiton, line 284: This looks a lot cleaner. I think it would be a good idea to also add a check function that receives the custom parameter and verifies the format with asserts. This way we provide more useful error messaging in situations where the custom parameters are formatted incorrectly or trying to do something that the user can't.

src/elfpy/simulators.py
```diff
@@ -347,18 +362,25 @@ def collect_and_execute_trades(self, last_block_in_sim=False):
...
                     [key for key in self.agents if key > 0]  # exclude init_lp before shuffling
                 )
                 wallet_ids = np.append(wallet_ids, 0)  # add init_lp so that they're always last
-            else:
+            else:  # include init_lp only on the last block, to let it unwind
                 wallet_ids = self.rng.permutation(list(self.agents))
+        else:  # we are in a deterministic mode
+            # reverse the list excluding 0 (init_lp)
+            wallet_ids = [key for key in self.agents if key > 0][::-1]
+            if self.config.simulator.init_lp and last_block_in_sim:  # prepend init_lp to the list
```
dpaiton, line 382: My understanding is that forcing init_lp to be last was a hack to ensure that the end-of-simulation accounting worked out and the printing didn't break. afaik, there's no hard reason why the init_lp agent has to stay in the simulation until the end, or has to be last in the closing trades.

If this is correct, do you think it's possible to update this PR to remedy the situation and remove the hack? If this is not correct, and the ordering is necessary, can you add a comment explaining why?

src/elfpy/strategies/single_lp.py
```diff
@@ -6,33 +6,29 @@
...
-    def __init__(self, market, rng, wallet_address, budget=1000):
-        """call basic policy init then add custom stuff"""
-        self.amount_to_lp = 100
-        super().__init__(
-            market=market,
-            rng=rng,
-            wallet_address=wallet_address,
-            budget=budget,
-        )
+    "single LP: only has one LP open at a time"
```
dpaiton, line 15: This is now a static element that belongs to the class. Before, the variable was an object member variable (belongs to `self`). The difference is whether you want the variable to exist even if the class is not instantiated. We want `amount_to_lp` to be owned by some instantiated user. e.g.
```
from elfpy.strategies.single_lp import Policy
alice = Policy()
alice.amount_to_lp = 100
bob = Policy()
bob.amount_to_lp = 500
```

so in this example `amount_to_lp` belongs to alice or bob, not to Policy.  As you have it above, it belongs to the policy class.

tl;dr -- put it back in the init

src/elfpy/strategies/single_short.py
```diff
@@ -4,29 +4,33 @@
...
-    simple short
-    only has one long open at a time
-    """
+    """single short: only has one long open at a time"""
 
-    def __init__(self, market, rng, wallet_address, budget=100):
-        """call basic policy init then add custom stuff"""
-        self.pt_to_short = 100
-        super().__init__(market, rng, wallet_address, budget)
+    pt_to_short = 100
```
dpaiton, line 15: See above; put `__init__` back

src/elfpy/strategies/single_lp.py
```diff
@@ -6,33 +6,29 @@
...
     def action(self):
-        """
-        implement user strategy
-        LP if you can, but only do it once
-        """
         action_list = []
         has_lp = self.wallet.lp_in_wallet > 0
-        can_lp = self.wallet.base_in_wallet >= self.amount_to_lp
+        can_lp = self.wallet.base_in_wallet >= self.base_to_lp
+        logging.info(
```
dpaiton, line 22: This should probably be logging.debug since it prints every trade

src/elfpy/strategies/single_short.py
```diff
@@ -4,29 +4,33 @@
...
 
     def action(self):
-        """
```
dpaiton, line 22: Every function, class, and python file needs a docstring

src/elfpy/wallet.py
```diff
@@ -5,7 +5,7 @@
 from dataclasses import dataclass
 from dataclasses import field
 
-from elfpy.utils.outputs import float_to_string
+from elfpy.utils.price import calc_apr_from_spot_price
```
dpaiton, line 8: Is this getting used? I can't find where

tests/test_trade.py
```diff
@@ -39,7 +39,7 @@ def setup_logging():
             handler,
         ]
 
-    def run_base_trade_test(self, user_policies, config_file, additional_overrides=None):
+    def run_base_trade_test(self, user_policies, config_file, delete_log_file=True, additional_overrides=None):
```
dpaiton, line 42: Good idea, this is better than commenting.

tests/test_trade.py
```diff
@@ -90,18 +93,21 @@ def run_base_lp_test(self, user_policies, config_file, additional_overrides=None
...
         market_apr = simulator.market.get_rate()
-        # check that apr is within a 0.1% of the target
-        assert np.allclose(
-            market_apr, target_pool_apr, atol=0.001
-        ), f"test_trade.run_base_lp_test: ERROR: {target_pool_apr=} does not equal {market_apr=}"
-        # check that the liquidity is within 5% of the target
-        assert np.allclose(
-            total_liquidity, target_liquidity, atol=target_liquidity * 0.05
-        ), f"test_trade.run_base_lp_test: ERROR: {target_liquidity=} does not equal {total_liquidity=}"
+        if config.simulator.init_lp:
```
dpaiton, line 96: this was my bad -- good catch

tests/utils/test_time_utils.py
```diff
@@ -23,10 +23,13 @@ def test_current_datetime(self):
...
+        before = now - time_delta
+        after = now + time_delta
 
         assert (
-            now < test_time < (now + datetime.timedelta(milliseconds=100))
-        ), f"Unexpected time value {test_time} should be close to {now}"
+            before < test_time < after
```
dpaiton, line 31: This is testing something different now, right? Why did you make the change?

tests/test_logging.py
wakamex: it gives an error on windows

wakamex (commented):
this reverts another previous incorrect attempt at a fix

tests/test_logging.py
wakamex: the error is
```
>       os.remove(file_loc)
E       PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\code\\elfpy\\.logging\\test_sim.log'
```

src/elfpy/simulators.py
wakamex: added in validate_custom_parameters() but needs testing, left in as a TODO

src/elfpy/simulators.py
wakamex: I guess depends whether we want to treat init_lp as an external force or as just another agent. if closer to the latter, i think we want him to act last, since he "opens the door" on initialising liquidity, *then some things happen*, it makes sense to me we'd also ask him to "shut the door" on the way out. then may also want to check if he lost money, or otherwise treat him like any other agent. otherwise, if it's just an external force we don't care about, he doesn't even need to act at the end. but i think i prefer treating him equally.

src/elfpy/strategies/single_lp.py
wakamex: ah that makes the difference clear, thanks. i was mostly trying to remove __init__ for brevity, which again results in screwing things up 🤦‍♂️

tests/utils/test_time_utils.py
wakamex: now < time_time seemed like a spurious test to me anyway, since both are initialized the exact same way

src/elfpy/simulators.py
wakamex: added testing, seems to work

src/elfpy/agent.py
wakamex: removed, not needed any more

src/elfpy/strategies/single_lp.py
wakamex: added docstrings everywhere

wakamex:
not needed, incorporating ideas into other PRs

PR #65, opened 2022-12-01
Add a bunch of type hints to Agent, Market, Pricing Models

sentilesdal:
The commits tell a story, but a lot of it is me chasing down type hints and fixing weird bugs because of it.  Might be easier to just look at the overall diffs. 

I mostly concentrated on agent.py and market.py which leaked out to pricing_models.py and a few other files.  I'm cutting myself off at this point to make sure I haven't broken anything.  While I was mostly adding type hints, I was doing drive-by cleanup as well.  For reviewing this PR, please make sure logic changes in places you've authored didn't break.

### DIFF
--- a/src/elfpy/agent.py
+++ b/src/elfpy/agent.py
@@ -17,6 +17,7 @@ from elfpy.wallet import Wallet
 AgentActionType = Literal["close_short", "close_long", "open_short", "open_long", "add_liquidity", "remove_liquidity"]
 TradeDirection = Literal["out", "in"]
 
+
 # TODO: The agent class has too many instance attributes (8/7)
 #     we should move some, like budget and wallet_address, into the agent wallet and out of User
 class Agent:
@@ -71,7 +72,9 @@ class Agent:
                     output_string += f" {key}: {float_to_string(value)}"
             print(output_string)
 
-    def create_agent_action(self, action_type: AgentActionType, trade_amount: float, mint_time: float=0) -> AgentAction:
+    def create_agent_action(
+        self, action_type: AgentActionType, trade_amount: float, mint_time: float = 0
+    ) -> AgentAction:
         """Instantiate a agent action"""
         agent_action = self.AgentAction(
             # these two variables are required to be set by the strategy
--- a/src/elfpy/agent.py
+++ b/src/elfpy/agent.py
@@ -18,7 +18,7 @@ from elfpy.wallet import Wallet
 #     we should move some, like budget and wallet_address, into the agent wallet and out of User
 class Agent:
     """
-    Implements abstract classes that control user behavior
+    Implements abstract classes that control agent behavior
     user has a budget that is a dict, keyed with a date
     value is an inte with how many tokens they have for that date
     """
@@ -33,12 +33,13 @@ class Agent:
         """
         self.market = market
         self.rng = rng
+        # TODO: remove this, wallet_address is a property of wallet, not the agent
         self.wallet_address = wallet_address
         self.budget = budget
         self.verbose = verbose
         self.last_update_spend = 0
         self.product_of_time_and_base = 0
-        self.wallet = Wallet(base_in_wallet=budget)
+        self.wallet = Wallet(address=wallet_address, base_in_wallet=budget)
 
     @dataclass
     class AgentAction:
--- a/src/elfpy/agent.py
+++ b/src/elfpy/agent.py
@@ -64,7 +64,7 @@ class AgentWallet:
 # TODO: The user class has too many instance attributes (8/7)
 #     we should move some, like budget and wallet_address, into the agent wallet and out of User
 # pylint: disable=too-many-instance-attributes
-class User:
+class Agent:
     """
     Implements abstract classes that control user behavior
     user has a budget that is a dict, keyed with a date
--- a/src/elfpy/agent.py
+++ b/src/elfpy/agent.py
@@ -2,22 +2,14 @@
 Implements abstract classes that control agent behavior
 """
 
-from dataclasses import dataclass
-from typing import Literal, TypeAlias
 
 import numpy as np
 from numpy.random._generator import Generator
-from elfpy.markets import Market
+from elfpy.markets import Market, MarketAction, MarketActionType
 
 from elfpy.utils.outputs import float_to_string
 from elfpy.utils.bcolors import Bcolors as bcolors
 from elfpy.wallet import Wallet
-
-
-AgentActionType: TypeAlias = Literal[
-    "close_short", "close_long", "open_short", "open_long", "add_liquidity", "remove_liquidity"
-]
-TradeDirection = Literal["out", "in"]
 
 
 # TODO: The agent class has too many instance attributes (8/7)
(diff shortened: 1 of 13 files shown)


src/elfpy/agent.py
```diff
@@ -1,142 +1,70 @@
...
-                        output_string += "[" + ", ".join([float_to_string(x) for x in value]) + "]"
-                    elif isinstance(value, dict):
-                        output_string += "{" + ", ".join([f"{k}: {float_to_string(v)}" for k, v in value.items()]) + "}"
-                    else:
-                        output_string += f"{value}"
-        return output_string
-
-
-# TODO: The user class has too many instance attributes (8/7)
+# TODO: The agent class has too many instance attributes (8/7)
```
dpaiton, line 15: Looks like you can remove this TODO, the pylint disable is gone and there is no linting error.

src/elfpy/agent.py
```diff
@@ -1,142 +1,70 @@
...
     """
-    Implements abstract classes that control user behavior
-    user has a budget that is a dict, keyed with a date
+    Implements abstract classes that control agent behavior
+    agent has a budget that is a dict, keyed with a date
     value is an inte with how many tokens they have for that date
     """
 
-    # TODO: variables assigned by child classes are referenced by User -- need to fix up user inheritance
+    # TODO: variables assigned by child classes are referenced by User -- need to fix up agent inheritance
```
dpaiton, line 24: User -> Agent
Although I think this PR might be addressing this TODO?

src/elfpy/markets.py
```diff
@@ -1,25 +1,56 @@
...
+
+    # these two variables are required to be set by the strategy
+    action_type: MarketActionType
+    trade_amount: float
+    # wallet_address is always set automatically by the basic agent class
+    wallet_address: int
+    # mint time is set only for trades that act on existing positions (close long or close short)
+    mint_time: float = 0
+
+    def print_description_string(self) -> None:
```
dpaiton, line 42: I know this func existed before the PR, but since I'm here -- is this not just an __str__ method with an extra print at the end? I think it makes more sense to overwrite __str__ with the fancy formatting, and then whoever calls `print_description_string` can instead use `print(MarketAction(...))`

src/elfpy/markets.py
```diff
@@ -176,38 +215,29 @@ def trade_and_update(self, agent_action):
...
         elif agent_action.action_type == "close_long":  # sell to close long
-            agent_action.direction = "out"  # calcOutGivenIn
-            agent_action.token_out = "base"  # sell known PT for unknown base
-            market_deltas, agent_deltas = self._close_long(agent_action)
+            market_deltas, agent_deltas = self._close_long(agent_action, "base", stretched_time_remaining)
         elif agent_action.action_type == "open_short":  # sell PT to open short
-            agent_action.direction = "in"  # calcOutGivenIn
-            agent_action.token_out = "pt"  # sell known PT for unknown base
-            market_deltas, agent_deltas = self._open_short(agent_action)
+            market_deltas, agent_deltas = self._open_short(agent_action, "pt", time_remaining)
```
dpaiton, line 234: I think open_short wants `stretched_time_remaining` as an argument to `calc_out_given_in`

src/elfpy/markets.py
```diff
@@ -346,23 +376,25 @@ def check_fees(
...
         trade_results = self.pricing_model.calc_out_given_in(
-            in_=trade_details.trade_amount,
+            in_=agent_action.trade_amount,
             share_reserves=self.share_reserves,
             bond_reserves=self.bond_reserves,
-            token_out=trade_details.token_out,
+            token_out=token_out,
             fee_percent=self.fee_percent,
-            time_remaining=trade_details.stretched_time_remaining,
+            time_remaining=time_remaining,
```
dpaiton, line 397: the arg name is time_remaining, but I do think it wants stretched_time_remaining. If I understand correctly, then `calc_out_given_in` is technically agnostic to whether the `time_remaining` is stretched or not. But the current iteration of both ElementV1 & Hyperdrive assume stretched time.

src/elfpy/simulators.py
```diff
@@ -41,11 +43,11 @@ def __init__(self, config_file):
         seconds_in_a_day = 86400
         self.time_between_blocks = seconds_in_a_day / self.config.simulator.num_blocks_per_day
         self.run_trade_number = 0
-        self.start_time = None
-        self.init_share_price = None
+        self.start_time: datetime.datetime = datetime.datetime.now()
```
dpaiton, line 46: I don't think this should be in init. `start_time` was meant to indicate the start time of the simulation (i.e. `run_simulation()` was invoked), not the time when the simulator class was instantiated. Also, the version that used the time_utils function included a time zone of GMT.

src/elfpy/simulators.py
```diff
@@ -149,6 +153,8 @@ def set_pricing_model(self, model_name):
         else:
             raise ValueError(f'pricing_model_name must be "HyperDrive" or "Element", not {model_name}')
 
+        return self.pricing_model
```
dpaiton, line 156: If it is being returned, then why is it also being assigned to `self` in the function? I think the better pattern is to make this a static method and assign the returned statement to `self.pricing_model` later.

src/elfpy/simulators.py
```diff
@@ -300,6 +315,16 @@ def run_simulation(self, override_dict=None):
...
         """Get trades from the agent list, execute them, and update states"""
+        if not isinstance(self.market, Market):
+            raise ValueError("market not defined")
+
+        # TODO: BIG HACK to make this PR pass.  There is a bug with removing liquidity that
+        # doesn't account for open shorts.  Someone can remove all liquidity while there are open
+        # shorts, then when another user goes to close a short, there are no reserves, so the calc
+        # functions error due to division by zero.  Need to fix the accounting of reserves to
+        # account for open short positions.
+        if last_block_in_sim:  # get all of a agent's trades
```
dpaiton, line 326: maybe change this comment to "reverse trade order so that init_lp is last" to remind us what `[::-1]` does.

src/elfpy/markets.py
sentilesdal: good catch, done.

src/elfpy/markets.py
sentilesdal: yup you're right

src/elfpy/markets.py
sentilesdal: really good catch, didn't know this.

src/elfpy/agent.py
sentilesdal: yes it did!

src/elfpy/simulators.py
sentilesdal: yeah you are right again.  I was trying to avoid having a type of `datetime | None` as that makes downstream things more annoying, but I'll just fix the downstream issues.

src/elfpy/simulators.py
sentilesdal: 100% agree here, there is no reason to keep `pricing_model` in the simulator class

dpaiton (approved):

PR #66, opened 2022-12-01
Mazygio/tests/utils

MazyGio:
Added tests for utils/price.py and utils/parse_config.py
Cleaned up comments in utils/price.py
Added tests for utils/price.py
Cleaned up comments in utils/time.py
Added tests for utils/time.py

TODO: add tests for outputs.py and data.py


tests/utils/test_price_utils.py
```diff
@@ -0,0 +1,577 @@
...
+                test_case["spot_price"]
+            )
+            assert liquidity == test_case["expected_result"], \
+                f"Calculated liquidity doesn't match the expected amount for these inputs: {test_case}"
+
+
+    def test_calc_base_asset_reserves(self):
+        """Unit tests for the calc_base_asset_reserves function"""
+
+        test_cases = [
```
sentilesdal, line 89: we should always be testing unhappy paths.  can we can a test where APR is 0, 100 and > 100%?

tests/utils/test_price_utils.py
```diff
@@ -0,0 +1,577 @@
...
+                base_asset_reserves,
+                test_case["expected_result"],
+                err_msg="unexpected base_asset_reserves",
+            )
+
+
+    def test_calc_liquidity(self):
+        """Unit tests for the calc_liquidity function"""
+
+        test_cases = [
```
sentilesdal, line 245: same here, let's try to break the function with crazy APR, market_price, share_price

tests/utils/test_price_utils.py
```diff
@@ -0,0 +1,577 @@
...
+                total_liquidity,
+                test_case["expected_total_liquidity"],
+                err_msg="unexpected total_liquidity"
+            )
+
+
+    # ### Spot Price and APR ###
+
+
+    def test_calc_apr_from_spot_price(self):
```
sentilesdal, line 369: price > 1.0?  If this breaks it, then have the expected result handle the error case.

tests/utils/test_price_utils.py
```diff
@@ -0,0 +1,577 @@
...
+            )
+
+
+    # ### YieldSpace ###
+
+
+    def test_calc_k_const(self):
+        """Unit tests for calc_k_const function"""
+
+        test_cases = [
```
sentilesdal, line 509: i like before how you had test cases for bond and share reserves == 0, add those here too!

tests/utils/test_time_utils.py
```diff
@@ -0,0 +1,364 @@
...
+            )
+
+            assert datetime.datetime.strftime(yearfrac_time, '%d/%m/%Y %H:%M:%S') ==  \
+                test_case["expected_result"], f'unexpected time value {yearfrac_time}'
+
+
+    def test_get_yearfrac_remaining(self):
+        """Unit tests for the get_yearfrac_remaining function"""
+
+        test_cases = [
```
sentilesdal, line 98: if mint_time > market_time is there an error?

sentilesdal (changes requested):
This code is SO clean.  Nice work Mazy.  I left some comments about testing unhappy paths, which are arguably more important to test than the happy paths.  Its crucial to know when and how a method will break.  This isn't as important for the simpler utils like in time.py, but a good rule of thumb is that you want to test any asserts/exceptions you see in the function.  also try to test things that don't make sense so we can see how it fails and verify to ourselves that the failure is graceful or breaks in a predictable/desired manner.

tests/utils/test_time_utils.py
MazyGio: Great question! There is not. Should there be? As it currently is, the function won't error out so testing for that situation will give a successful (albeit nonsensical) result

dpaiton: We should probably have an error -- this should be an impossible situation. mint_time is assigned to market_time when minting happens, and market_time is a monotonically increasing variable. So mint_time should always be <= market_time.

src/elfpy/utils/parse_config.py
```diff
@@ -91,16 +91,27 @@ def parse_simulation_config(config_file):
...
-        case "info":
-            level = logging.INFO
-        case "warning":
-            level = logging.WARNING
-        case "error":
-            level = logging.ERROR
-        case "critical":
-            level = logging.CRITICAL
+    
+    if simulation_config.simulator.logging_level.lower() == "debug":
```
dpaiton, line 95: We updated the repo requirements to python 3.10 for the `match` statement below to work, among other reasons. I don't think the if statements are necessarily bad, but the match certainly look cleaner. Is the reason you changed because your python version is <3.10? I think it's better to keep the old version, but I'm open to discussing it.

tests/utils/test_parse_config_success_data.toml
```diff
@@ -0,0 +1,44 @@
...
+verbose = false # toggle verbose output level
+shuffle_users = true
+init_lp = true
+target_liquidity = 10000000
+target_daily_volume = 1000
+init_pool_apy = 0.05
+fee_percent = 0.10
+init_vault_age = 0
+vault_apy = [0.05]
+logging_level = "warning"
```
dpaiton, line 44: add a newline at the end of the file

tests/utils/test_price_utils.py
dpaiton: I think this is resolved? @sentilesdal or @MazyGio can you resolve it if you're happy?

tests/utils/test_price_utils.py
dpaiton: I think this is resolved? @sentilesdal or @MazyGio can you resolve it if you're happy?

sentilesdal (approved):
these look great, thanks for adding error tests!

tests/utils/test_price_utils.py
MazyGio: Added error cases, resolving this now. Stumbled upon negative results though under some edge-case-ish situations where fixed APR is very low and share_price is higher than init_share_price (i.e. users bought up all the PTs). The code runs but we might want to check the math for those cases

dpaiton (approved):

PR #71, opened 2022-12-06
updates the agent_list to be a dictionary called agents

dpaiton:
also reworks init_lp trade ordering hack


dpaiton:
change log:

**Major updates:**

* refactors lp initialization to use correct amounts for target_liquidity and apr
* adds tests to ensure that market apr & liquidity match targets
* adjusts units in price.calc_liquidity from USD to base
* changes simulator.agents type to a dictionary instead of a list
* fixes wallet delta logic in `_open_short` and `_close_short`

**Minor updates:**

* fixes bug with close_short to use `token_out=base` instead of `pt`
* fixes up some logging statements
* adds `vault_apy` member variable to the market class (temporary)
* Market.get_rate checks if `share_reserves is <= 0` instead of `== 0`
* fixes incorrect variable name in `Markets.open_short`
* updates agent deltas to include wallet address
* updates `init_share_price` to be hard-coded to `1` for element pricing model
* reorganizes `analysis_dict` output to reflect the same order as the declared keys in `YieldSimulator.__init__`
* renames `init_lp` arguments and member variables to reflect actual actions & units
* creates `setup_logging` func in `test_trade` and `test_sim`

src/elfpy/markets.py
```diff
@@ -224,7 +224,7 @@ def update_market(self, market_deltas: MarketDeltas) -> None:
     def get_rate(self):
         """Returns the current market apr"""
         # calc_apr_from_spot_price will throw an error if share_reserves is zero
-        if self.share_reserves == 0:  # market is empty
+        if self.share_reserves <= 0:  # market is empty
```
sentilesdal, line 235: update the comment.  if less than zero, this is probably a rounding error?

src/elfpy/simulators.py
```diff
@@ -246,11 +246,11 @@ def setup_simulated_entities(self, override_dict=None):
                 pt_to_short=init_token_asset_reserves / 10,
                 short_until_apr=self.config.simulator.init_pool_apy,
             )
-            self.agent_list = [initial_lp]
+            self.agents = {initial_lp.wallet_address: initial_lp}
```
sentilesdal, line 271: yes, much better

src/elfpy/simulators.py
```diff
@@ -170,6 +173,63 @@ def override_variables(self, override_dict):
...
+            days_remaining=self.config.simulator.token_duration,
+            time_stretch=self.market.time_stretch_constant,
+            init_share_price=self.market.init_share_price,
+            share_price=self.market.init_share_price,
+        )[:2]
+        normalized_days_until_maturity = self.config.simulator.token_duration  # `t`; full duration
+        stretch_time_remaining = time_utils.stretch_time(
+            normalized_days_until_maturity, self.market.time_stretch_constant
+        )
+        output_with_fee = self.market.pricing_model.calc_out_given_in(
```
sentilesdal, line 202: now we get the trade amount to set the ratios correctly

src/elfpy/simulators.py
```diff
@@ -170,6 +173,63 @@ def override_variables(self, override_dict):
...
+            in_=init_bond_reserves,
+            share_reserves=init_share_reserves,
+            bond_reserves=0,
+            token_out="pt",
+            fee_percent=self.config.simulator.fee_percent,
+            time_remaining=stretch_time_remaining,
+            init_share_price=self.market.init_share_price,
+            share_price=self.market.init_share_price,
+        )[1]
+        amount_to_lp = init_share_reserves + output_with_fee
```
sentilesdal: the total amount to deposit initially?  we need extra because we are going to be doing a trade?

sentilesdal (approved):

src/elfpy/simulators.py
```diff
@@ -299,95 +321,92 @@ def run_simulation(self):
...
-        if last_block_in_sim:  # get all of a agent's trades
-            # we are reversing order here to make sure that agent 0's remove_liquidity transaction
-            # is last
-            self.agent_list = self.agent_list[::-1]
-
-        for agent in self.agent_list:  # trade is different on the last block
+        # TODO: This is a HACK to prevent the initial LPer from rugging other agents.
+        # The initial LPer should be able to remove their liquidity and any open shorts can still be closed.
+        # But right now, if the LPer removes liquidity while shorts are open,
+        # then closing the shorts results in an error (share_reserves == 0).
```
wakamex, line 353: I think this is fine, but we can create a test to evaluate this behavior. once we do, we can remove this TODO. tracking issue #74

tests/test_trade.py
```diff
@@ -51,25 +61,47 @@ def run_base_trade_test(self, user_policies, config_file, additional_overrides=N
             override_dict.update(additional_overrides)
         simulator.setup_simulated_entities(override_dict)
         simulator.run_simulation()
+        # comment this to view the generated log files
+        file_loc = logging.getLogger().handlers[0].baseFilename
+        os.remove(file_loc)
```
wakamex, line 66: instead of forcing us to comment out a line, shouldn't we move this file wipe step at the beginning of logging, and leave the file on the hard drive in between runs by default?

PR #75, opened 2022-12-08
test on windows

wakamex:


wakamex:
the solution is to not use windows

PR #78, opened 2022-12-12
Simulator refactor

dpaiton:
Creates `example_main.py` and demonstrates a new class inheritance architecture.


dpaiton:
**Major changes**
* moves src/elfpy/strategies to src/elfpy/policies
* renames YieldSimulator to Simulator
* refactors to removes class codependencies
* adds example_main.py to illustrate an example workflow
* moves several Simulator class functions to a utility folder
* modifies override_dict functionality: it now uses a src/elfpy/utils/sim_utils and there is one for config and one for random variables

**Minor changes**
* removes target_daily_volume from example config and tests
* updates tests to use a boolean variable to determine if test logs should be deleted
* updates tests to reflect new workflow

sentilesdal (approved):

PR #81, opened 2022-12-13
Integrate the flat+curve pricing model

jalextowle:
This PR makes several large changes to the simulations repo (ordered by importance):

1. Adds the flat+curve pricing model.
2. Refactored the `TradeResult` type output by trading functions. This update allows pricing models to provide more granular information to the `Market` class, which is used by the flat+curve pricing model.
3. Added a type that holds the state for the `Market` class. This cuts down significantly on the amount of parameters that need to be passed to trading functions.
4. Unified all of the time formats into a single `StretchedTime` format. Follow-up PRs will refine the usage of this type.
5. Added a type that holds an amount with a unit called `Quantity`. Bundling trading amounts with trading units will cut down on mistakes made when needing to pass in units to trading functions.


src/elfpy/markets.py
```diff
@@ -181,50 +115,57 @@ def trade_and_update(self, agent_action: MarketAction, pricing_model: PricingMod
         self.check_action_type(agent_action.action_type, pricing_model.model_name())
         # TODO: check the desired amount is feasible, otherwise return descriptive error
         # update market variables which may have changed since the user action was created
-        time_remaining = time_utils.get_yearfrac_remaining(self.time, agent_action.mint_time, self.token_duration)
-        stretched_time_remaining = time_utils.stretch_time(time_remaining, self.time_stretch_constant)
+        time_remaining = StretchedTime(
+            # TODO: We should improve the StretchedTime API so that it accepts yearfracs in
```
dpaiton, line 119: @sentilesdal has been advocating for moving away from yearfracs in general, and just reporting seconds. With the StretchedTime class we can probably include nice unit converters so people can use whatever units they like? Worth thinking about if we're already addressing the API.

src/elfpy/markets.py
```diff
@@ -181,50 +115,57 @@ def trade_and_update(self, agent_action: MarketAction, pricing_model: PricingMod
...
-                stretched_time_remaining=stretched_time_remaining,
+                time_remaining=time_remaining,
             )
         elif agent_action.action_type == "add_liquidity":
             market_deltas, agent_deltas = self._add_liquidity(
                 pricing_model=pricing_model,
                 agent_action=agent_action,
-                time_remaining=time_remaining,
-                stretched_time_remaining=stretched_time_remaining,
+                time_remaining=time_remaining.normalized_days,
```
dpaiton, line 160: I think this should receive the `time_remaining: StretchedTime` variable and then it can pull out `normalized days` & `stretched_time` within the function? The trade-off is that it is less explicit about how the argument is being used, but reducing the signature and keeping it consistent with the other actions is worth it imo.

src/elfpy/markets.py
```diff
@@ -241,35 +182,30 @@ def update_market(self, market_deltas: MarketDeltas) -> None:
         for key, value in market_deltas.__dict__.items():
             if value:  # check that it's instantiated and non-empty
                 assert np.isfinite(value), f"markets.update_market: ERROR: market delta key {key} is not finite."
-        self.share_reserves += market_deltas.d_base_asset / self.share_price
-        self.bond_reserves += market_deltas.d_token_asset
-        self.share_buffer += market_deltas.d_share_buffer
-        self.bond_buffer += market_deltas.d_bond_buffer
-        self.lp_reserves += market_deltas.d_lp_reserves
+        self.market_state.apply_delta(market_deltas)
```
dpaiton, line 185: Love this!

src/elfpy/markets.py
```diff
@@ -471,24 +428,35 @@ def _open_long(
...
-        stretched_time_remaining: float,
+        time_remaining: StretchedTime,
     ) -> tuple[MarketDeltas, Wallet]:
         """
         take trade spec & turn it into trade details
         compute wallet update spec with specific details
             will be conditional on the pricing model
         """
-        if agent_action.trade_amount <= self.bond_reserves:
+        # TODO: Why are we clamping elsewhere but we don't apply the trade at
```
dpaiton, line 438: I agree, I think we shouldn't do either. I think it is too opaque to the end user. The agent should check this before making the trade. Also I think a realistic response would be that the market rejects the trade and informs the agent (& user). We could set a market variable  that allows "illegal" trades, for testing or whatever.

src/elfpy/markets.py
jalextowle: Yeah I agree, I was trying to avoid making these changes in this PR, but I'm probably going to go ahead and throw them in so that we can be done with them.

src/elfpy/markets.py
jalextowle: I also like the approach of all of our components being strict about their inputs. This will serve us well if we ever try to run the simulation framework on our smart contracts, which will certainly be strict about the inputs that they accept.

src/elfpy/markets.py
jalextowle: We can address these changes in follow-up PRs (as discussed offline).

src/elfpy/pricing_models/base.py
```diff
@@ -0,0 +1,223 @@
...
+from elfpy.types import Quantity, MarketState, StretchedTime, TradeResult
+import elfpy.utils.price as price_utils
+
+# TODO: Currently many functions use >5 arguments.
+# These should be packaged up into shared variables, e.g.
+#     reserves = (in_reserves, out_reserves)
+#     share_prices = (init_share_price, share_price)
+# pylint: disable=too-many-arguments
+
+# TODO: This module is too big, we should break it up into pricing_models/{base.py, element.py, hyperdrive.py}
```
dpaiton, line 14: I think this TODO & pylint error are resolved now.

src/elfpy/pricing_models/base.py
```diff
@@ -0,0 +1,223 @@
...
+# pylint: disable=too-many-lines
+
+# TODO: some functions have too many local variables (15 is recommended max),
```
dpaiton, line 17: This is also probably resolved now.

src/elfpy/pricing_models/base.py
```diff
@@ -0,0 +1,223 @@
+"""The base pricing model."""
+
+from abc import ABC, abstractmethod
+
+from elfpy.types import Quantity, MarketState, StretchedTime, TradeResult
+import elfpy.utils.price as price_utils
+
+# TODO: Currently many functions use >5 arguments.
```
dpaiton, line 8: Can you check these pylint disables & TODOS? They all might be resolved. If not, leave them.

src/elfpy/pricing_models/hyperdrive.py
```diff
@@ -0,0 +1,245 @@
...
+        time_remaining : StretchedTime
+            The time remaining for the asset (incorporates time stretch).
+
+        Returns
+        -------
+        TradeResult
+            The result of performing the trade.
+        """
+
+        market_state = copy.copy(market_state)
```
dpaiton, line 192: Why is this done here and not in `calc_in_given_out`?

dpaiton (commented):
I added a few small comments & questions that should be quick to address. Please also make sure tests/lints are passing and then we can push it through.

dpaiton (approved):

PR #88, opened 2022-12-16
Use enums instead of literals

jalextowle:


src/elfpy/policies/init_lp.py
```diff
@@ -45,20 +46,30 @@ def action(self, market: Market, pricing_model: PricingModel):
...
+                        action_type=MarketActionType.ADD_LIQUIDITY, trade_amount=self.second_base_to_lp
+                    ),
                 ]
             elif pricing_model.model_name() == HyperdrivePricingModel().model_name():
                 # TODO: This PM fails the tests
                 action_list = [
-                    self.create_agent_action(action_type="add_liquidity", trade_amount=self.first_base_to_lp),
-                    self.create_agent_action(action_type="open_short", trade_amount=self.pt_to_short),
-                    self.create_agent_action(action_type="add_liquidity", trade_amount=self.second_base_to_lp),
+                    self.create_agent_action(
```
dpaiton, line 56: Glad to see this is more constrained, good call

src/elfpy/types.py
```diff
@@ -1,18 +1,30 @@
 """A set of common types used throughtout the simulation codebase."""
 
 from dataclasses import dataclass
-from typing import TypeAlias
-from typing import Literal
-from elfpy.utils.outputs import float_to_string
+from enum import Enum
```
dpaiton, line 4: stdlib should go above elfpy imports

dpaiton (approved):
Looks good after the import line is moved, thanks for making the change.

PR #90, opened 2022-12-16
LP refactor

dpaiton:


config/example_config.toml
```diff
@@ -12,7 +12,7 @@ max_vault_apy = 0.9  # as a decimal
 base_asset_price = 1  # aka market price
 
 [amm]
-pricing_model_name = "Hyperdrive" # specify a pricing model
+pricing_model_name = "Yieldspace" # specify a pricing model
```
jalextowle, line 15: I think "Hyperdrive" is a better default. Happy to keep it like this until @sentilesdal finished us flat+curve testing though.

src/elfpy/agent.py
```diff
@@ -2,15 +2,19 @@
 Implements abstract classes that control agent behavior
 """
 
+from __future__ import annotations  # types will be strings by default in 3.11
```
jalextowle, line 5: Perhaps it makes sense to make a todo for this? Not a big deal either way, but good to keep track of things like this.

src/elfpy/markets.py
```diff
@@ -141,15 +142,13 @@ def trade_and_update(self, agent_action: MarketAction, pricing_model: PricingMod
             market_deltas, agent_deltas = self._add_liquidity(
                 pricing_model=pricing_model,
                 agent_action=agent_action,
-                time_remaining=time_remaining.normalized_time,
-                stretched_time_remaining=time_remaining.stretched_time,
+                time_remaining=time_remaining,
```
jalextowle, line 145: 🙏

src/elfpy/pricing_models/base.py
```diff
@@ -7,16 +7,11 @@
 
 
 class PricingModel(ABC):
-    """
-    Contains functions for calculating AMM variables
+    """Contains functions for calculating AMM variables
```
jalextowle, line 10: Nit: Was this intentional? I prefer the """ on a separate line, but you know the correct style better than I do.

src/elfpy/pricing_models/base.py
```diff
@@ -39,71 +34,45 @@ def calc_out_given_in(
...
-        # TODO: This pylint will be resolved when we switch to MarketState
-        # pylint: disable=too-many-arguments
         raise NotImplementedError
 
     @abstractmethod
     def model_name(self) -> str:
         """Unique name given to the model, can be based on member variable states"""
         raise NotImplementedError
 
+    # TODO: FIXME: This does not work with yieldspace/hyperdrive
```
jalextowle, line 75: How so? It looks fine to me, so I'd be curious to hear what isn't working.

config/example_config.toml
dpaiton: oh actually I agree -- I changed this when hunting a bug and never changed it back.

src/elfpy/utils/price.py
```diff
@@ -174,7 +235,7 @@ def calc_apr_from_spot_price(price: float, time_remaining: StretchedTime):
         "utils.price.calc_apr_from_spot_price: ERROR: "
         f"time_remaining.normalized_time should be greater than zero, not {time_remaining.normalized_time}"
     )
-    return (1 - price) / (price * time_remaining.normalized_time)  # price = 1 / (1 + r * t)
+    return (1 - price) / (price * time_remaining.normalized_time)  # r = ((1/s)-1)/t = (1-s)/(st)
```
jalextowle, line 238: Nit: Would you mind using `p` instead of `s`? We use `p` pretty universally in our comments and docs when referring to spot price.

src/elfpy/agent.py
dpaiton: I think as a general rule of thumb any time we increment the minimum python version we should go through all __future__ imports and update them.

src/elfpy/pricing_models/base.py
dpaiton: It's not required, but the examples do it that way: https://peps.python.org/pep-0008/#documentation-strings

I like it because it separates the heading from the rest of the doc strings. It's also how docs are written in a few of the python science computation repos I follow.

The most important thing is that we are consistent so that the auto-generated docs will work.

src/elfpy/pricing_models/base.py
dpaiton: I removed the TODO, because I'm not certain now.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -56,42 +49,49 @@ def calc_lp_out_given_tokens_in(
...
+            f"Expected bond_reserves >= 0, not {market_state.bond_reserves}!"
+        )
+        assert market_state.base_buffer >= 0, (
+            "pricing_models.calc_lp_out_given_tokens_in: ERROR: "
+            f"Expected base_buffer >= 0, not {market_state.base_buffer}!"
+        )
+        assert market_state.lp_reserves >= 0, (
+            "pricing_models.calc_lp_out_given_tokens_in: ERROR: "
+            f"Expected lp_reserves >= 0, not {market_state.lp_reserves}!"
+        )
```
jalextowle, line 67: I think we'll ultimately want to de-couple assertions from these LP functions in the same way that we did with the trading functions assertions. This strengthens the argument that LP and trading logic should be separately IMO since it would make the naming for these assertion functions cleaner.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -56,42 +49,49 @@ def calc_lp_out_given_tokens_in(
...
         assert rate >= 0, f"pricing_models.calc_lp_out_given_tokens_in: ERROR: expected rate >= 0, not {rate}!"
-        assert (
-            1 > time_remaining >= 0
-        ), f"pricing_models.calc_lp_out_given_tokens_in: ERROR: expected 1 > time_remaining >= 0, not {time_remaining}!"
-        assert stretched_time_remaining >= 0, (
+        assert 1 > time_remaining.normalized_time >= 0, (
+            "pricing_models.calc_lp_out_given_tokens_in: ERROR: "
+            f"expected 1 > time_remaining >= 0, not {time_remaining.normalied_time}!"
+        )
+        assert time_remaining.stretched_time >= 0, (
```
jalextowle, line 73: We should probably encapsulate sanity checks within our time structures so that we don't need to run the asserts every time a time value is used (I realize that this may make things harder during the refactor of time in the simulator, so I'd encourage us to punt on this for now).

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -56,42 +49,49 @@ def calc_lp_out_given_tokens_in(
...
             lp_out = d_shares
         # TODO: Move this calculation to a helper function.
-        d_bonds = (share_reserves + d_shares) / 2 * (
-            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
-        ) - bond_reserves
+        d_bonds = (market_state.share_reserves + d_shares) / 2 * (
+            market_state.init_share_price
+            * (1 + rate * time_remaining.normalized_time) ** (1 / time_remaining.stretched_time)
+            - market_state.share_price
+        ) - market_state.bond_reserves
```
jalextowle, line 94: This logic should be abstracted (I realize that you didn't write it). The form is always the same except sometimes it's `market_state.share_reserves` or `market_state.share_reserves - d_shares`. Besides, having a helper function that computes the bond reserves given shares reserves, mu, c, and APR is super useful when writing unit tests.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -110,134 +110,141 @@ def calc_lp_out_given_tokens_in(
...
+            market_state.bond_reserves,
+            market_state.base_buffer,
+            market_state.init_share_price,
+            market_state.share_price,
+            market_state.lp_reserves,
             rate,
-            time_remaining,
-            stretched_time_remaining,
+            time_remaining.normalized_time,
+            time_remaining.stretched_time,
```
jalextowle, line 121: Ultimately, I'd think that we should log market state and time remaining as objects to reduce the logging bloat. We just need to implement `as_str` functions for this, right? Anyways, more of a note of something to think about (and a good first issue for other team members!).

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -110,134 +110,141 @@ def calc_lp_out_given_tokens_in(
...
         )
-        d_shares = d_base / share_price
-        lp_in = (d_shares * lp_reserves) / (share_reserves - base_buffer / share_price)
+        d_shares = d_base / market_state.share_price
+        lp_in = (d_shares * market_state.lp_reserves) / (market_state.share_reserves - market_state.base_buffer)
         # TODO: Move this calculation to a helper function.
-        d_bonds = (share_reserves - d_shares) / 2 * (
-            init_share_price * (1 + rate * time_remaining) ** (1 / stretched_time_remaining) - share_price
-        ) - bond_reserves
+        d_bonds = (market_state.share_reserves - d_shares) / 2 * (
```
jalextowle, line 189: Case in point of my earlier comment about why this logic should be abstracted. It is tightly coupled to pricing models, so perhaps it does make sense for the two models to live side by side 🤔.

src/elfpy/utils/price.py
```diff
@@ -1,49 +1,109 @@
...
-        Future Value = Present Value * (1 + r)^n
-    The equation converts from future value to present value at the appropriate discount rate,
-    which measures the opportunity cost of getting a dollar tomorrow instead of today.
-    discount rate = (1 + r)^n
-    spot price APR = 1 / (1 + r)^n
+    bond_reserves = (share_reserves / 2) * (
+        init_share_price * (1 + target_apr * time_remaining.normalized_time) ** (1 / time_remaining.stretched_time)
+        - share_price
+    )  # y = x/2 * (u * (1 + rt)**(1/T) - c)
+    return bond_reserves
```
jalextowle, line 50: Oh perfect, this is what I was looking for.

src/elfpy/utils/price.py
```diff
@@ -1,49 +1,109 @@
...
+
+    Returns
+    -------
+    float
+        The expected amount of base asset in the pool, calculated from the provided parameters
+    """
+    share_reserves = bond_reserves / (
+        init_share_price * (1 - target_apr * time_remaining.normalized_time) ** (1 / time_remaining.stretched_time)
+    )  # z = y / (u * (1 - rt)**(1/T))
+    return share_reserves
```
jalextowle, line 84: Same as above. It probably makes sense to stub this in the base pricing model and copy this to the YieldSpace model (YieldSpace and Hyperdrive can share this).

src/elfpy/utils/price.py
```diff
@@ -87,14 +147,10 @@ def calc_base_asset_reserves(
...
-    share_price: float = 1,
-):
-    """
-    Returns the reserve volumes and total supply
+    target_apr: float,
+    market: Market,
+) -> tuple[float, float]:
+    """Returns the reserve volumes and total supply
 
     The scaling factor ensures token_asset_reserves and base_asset_reserves add
```
jalextowle, line 155: Please change `token_asset_reserves` and `base_asset_reserves` to `bond_reserves` and `share_reserves`, respectively.

tests/utils/test_price_utils.py
```diff
@@ -946,3 +823,27 @@ def test_calc_k_const(self):
...
+        """Execute the test"""
+        self.run_calc_spot_price_from_apr_test()
+
+    def test_calc_base_asset_reserves(self):
+        """Execute the test"""
+        self.run_calc_base_asset_reserves_test()
+
+    def test_calc_liquidity(self):
+        """Execute the test"""
+        self.run_calc_liquidity_test()
```
jalextowle, line 849: I don't see the value in structuring our tests like this. Can you explain what the goal of this architecture is? Generally speaking, I see the value in parameterized test runners that can be composed to lead tests of emergent behavior, but I don't see that happening here.

jalextowle (commented):
It's looking good overall. There are a few comments I'd like to see addressed before I approve (in particular the comment about the unit testing changes in `test/utils/test_price_utils.py`).

src/elfpy/agent.py
jalextowle: Yep, that makes sense 👍

src/elfpy/pricing_models/base.py
jalextowle: Sure, that makes sense. I'm good with changing to this style.

tests/utils/test_price_utils.py
dpaiton: It's a boiler plate that has been useful elsewhere. For example, we can write a generic `calc_x_test` that receives a parameter, which is then looped in the `test_x` function. It's also nice for debugging because you can turn on/off tests easily. Elsewhere we use this structure to make it easier to turn on/off deletion of log files. You're correct that none of that is actually used here. I had a use set up in an earlier iteration of the refactor, but when we removed the element PM I no longer needed it.

src/elfpy/pricing_models/yieldspace.py
dpaiton: Yeah, there's a TODO for that right above the line

tests/utils/test_price_utils.py
jalextowle: Ok, that all makes sense.

dpaiton: e.g.:
``` 
    def test_calc_liquidity(self):
        """Execute the test"""
        self.run_calc_liquidity_test(sim_utils.get_pricing_model("hyperdrive"))
        self.run_calc_liquidity_test(sim_utils.get_pricing_model("yieldspace"))

```

dpaiton: also here, in test_trade, where we run the same base test function with different arguments:
```
    def test_single_long(self):
        """Tests the BaseUser class"""
        self.run_base_trade_test(agent_policies=["single_long"], config_file="config/example_config.toml")

    def test_single_short(self):
        """Tests the BaseUser class"""
        self.run_base_trade_test(agent_policies=["single_short"], config_file="config/example_config.toml")
```

jalextowle (approved):

PR #91, opened 2022-12-16
Rename some of the fields in wallet

jalextowle:


src/elfpy/wallet.py
```diff
@@ -16,21 +16,23 @@ class Wallet:
 
     # fungible
     address: int
-    base_in_wallet: float
-    lp_in_wallet: float = 0  # they're fungible!
+    base: float
+    lp: float = 0  # pylint: disable=invalid-name
```
dpaiton, line 20: what if you just made this a valid name? like liquidity?

src/elfpy/wallet.py
```diff
@@ -16,21 +16,23 @@ class Wallet:
 
     # fungible
```
dpaiton, line 17: Can you add a description of each of these variables in the docstring for this class?

src/elfpy/wallet.py
jalextowle: Liquidity doesn't really have the same meaning to be honest. The linting directive doesn't really bother me.

dpaiton (approved):
Change `lp` to `lp_tokens` and then it's good.

PR #92, opened 2022-12-17
Switched from using a share buffer to a base buffer

jalextowle:
The purpose of the base buffer is to store the amount of base that should be held in anticipation of paying off bonds that need to be redeemed. This base doesn't change based on the variable rate interest rate (the base required to pay off a bond is simply the face value of the bond). For this reason, we should be storing this buffer in units of base rather than shares since using shares would result in the value of the buffer increasing over time, which would systemically overcount the amount of reserves that should be set aside for bond obligations.


jrhea (approved):
lgtm, it would be good to update the AMM math doc to represent this

PR #100, opened 2023-01-04
Started the MVP simulation

jalextowle:
This PR adds a new simulation example with the following features:
1. Uses a Geometric Brownian Motion process to model the vault's APR.
2. Uses a random agent to open and close longs.


src/elfpy/utils/parse_config.py
```diff
@@ -60,7 +60,6 @@ def parse_simulation_config(config_dict):
         amm=AMMConfig(**config_dict["amm"]),
         simulator=SimulatorConfig(**config_dict["simulator"]),
     )
-    simulation_config.simulator.logging_level = text_to_logging_level(simulation_config.simulator.logging_level)
```
dpaiton, line 63: The default value for the SimulatorConfig dataclass logging level is a string, but the logging level needs to be an int. the `text_to_logging_level` function makes this conversion happen. I am probably missing something, but why are you able to remove this conversion here and elsewhere?

src/elfpy/utils/config.py
```diff
@@ -40,27 +40,38 @@ class AMMConfig:
...
+
+    # vault
     init_vault_age: float = field(default=0, metadata={"hint": "initial vault age"})
-    vault_apy: list[float] = field(
+    vault_apr: list[float] = field(
         default_factory=list, metadata={"hint": "the underlying (variable) vault apy at each time step"}
     )
+
+    # logging
+    verbose: bool = field(default=False, metadata={"hint": "verbosity level for logging"})
```
dpaiton, line 68: I believe we can remove the verbose config flag, it is no longer used. Somehow this was missed in earlier logging-related PRs.

src/elfpy/utils/parse_config.py
jalextowle: I can understand the confusion. The impetus of the change was that the previous code was assigning a string to the config and then assigned an int to the config, which was problematic from a typing perspective (and pyright complained about it). What I've done instead of this conversion is keep the configuration as a string always but pass the log level (string) into `test_to_logging_level` when I pass the log level into the logger. I'm going to abstract this logic to avoid the linter's code duplication complaints, so I will write a function that takes care of this.

examples/simulation.py
```diff
@@ -0,0 +1,208 @@
+"""A minimum viable simulation"""
+# stdlib
+import argparse
+from typing import Any, Optional
+
+# external imports
+import numpy as np
+from numpy.random import Generator
+from stochastic.processes import GeometricBrownianMotion
```
dpaiton, line 9: I am a wary about adding `stochastic` as a dependency for our repo. It seems popular (322 stars) but is maintained primarily by a single PhD student. I've been burned by this type of thing in the past. The [code](https://scipy-cookbook.readthedocs.io/items/BrownianMotion.html) to generate it from more core repos is pretty straight forward. What are your thoughts on doing it in-house using just scipy & numpy?

tests/pricing_models/test_get_max_long.py
```diff
@@ -0,0 +1,187 @@
...
+    """Dataclass for get_max_long test cases"""
+
+    market_state: MarketState
+    fee_percent: float
+    time_remaining: StretchedTime
+
+    __test__ = False  # pytest: don't test this class
+
+
+class TestGetMaxLong(unittest.TestCase):
```
dpaiton, line 31: It's not clear to me that this is actually testing that the method found the maximum long. It seems to only be testing that the solution is a valid trade. Is that correct?

dpaiton (approved):

PR #105, opened 2023-01-06
Refactor vault apr

dpaiton:


src/elfpy/utils/sim_utils.py
```diff
@@ -191,6 +183,56 @@ def get_pricing_model(model_name: str) -> PricingModel:
...
+                        high=config.market.vault_apr["high"],
+                        size=config.simulator.num_trading_days,
+                    ).tolist()
+                case "geometricbrownianmotion":
+                    # the n argument is number of steps, so the number of points is n+1
+                    vault_apr = (
+                        GeometricBrownianMotion(rng=rng)
+                        .sample(n=config.simulator.num_trading_days - 1, initial=config.market.vault_apr["initial"])
+                        .tolist()
+                    )
```
jalextowle, line 222: Nice!

src/elfpy/simulators.py
```diff
@@ -213,14 +214,17 @@ def run_simulation(self) -> None:
...
-                self.market.market_state.share_price += (
-                    self.random_variables.vault_apr[self.day]  # current day's apy
-                    / 365  # convert annual yield to daily
-                    * self.market.market_state.init_share_price  # APR, apply return to starting price (no compounding)
-                    # * self.market.share_price # APY, apply return to latest price (full compounding)
+                delta = MarketDeltas(
+                    d_share_price=(
+                        self.market.market_state.vault_apr  # current day's apy
+                        / 365  # convert annual yield to daily
+                        * self.market.market_state.init_share_price  # apply return to starting price (no compounding)
```
jalextowle, line 224: I think it probably makes sense to use compounding (or at least make this configurable). 

See the "Calculating Accrued Interest" section of this document: https://docs.compound.finance/v2/. Since Compound has so much activity, its users experience a close approximation of compounding every block. I spent some time reading through the Compound III contracts, and I believe this to be the case for those contracts too.

jalextowle (approved):
I'm going to approve because it doesn't seem like the sawtooth notebook is the main purpose of this PR. Having said this, I'd like to run it, and I'm still not entirely sure how I'm supposed to install the `elfpy` package in Googe Collab. I'd appreciate it if you could include instructions for how to do this in the README in this PR.

src/elfpy/simulators.py
dpaiton: This was the previously-existing formula; I didn't modify it in this PR. I guess I should add a flag to config.simulator? Or do you think there is no scenario where one might want a "no compounding" option?

jalextowle: This is a case in which I think having both options is good, so a flag would be good. One should use `self.market.market_state.share_price` and the other should use `init_share_price` (as is currently the case).

PR #107, opened 2023-01-09
simplify time

wakamex:
StretchedTime stores only one internal representation, in years (_years)
- days returns a function of _years
- normalized_time returns _years
- stretched_time returns a function of _days (could be changed to _years, that's just how current util function is set up)


sentilesdal:
closing as there have been more recent refactors

PR #108, opened 2023-01-09
Get the random long agent to work

jalextowle:
This PR puts the last puzzle piece in place for the random long simulation to be able to run for an entire term without crashing. In particular, it fixed a bug in the market's long accounting and it fixed a bug in the `RandomAgent` where the random agent wasn't using the `mint_time` field of `MarketAction` correctly.


sentilesdal (approved):
lgtm

.github/workflows/lint.yml
```diff
@@ -26,6 +26,6 @@ jobs:
       - name: Analyzing code with black
         run: python -m black --check --diff .
       - name: Analysing code with pylint
-        run: python -m pylint --rcfile=.github/workflows/.pylintrc $(git ls-files '*.py')
+        run: python -m pylint $(git ls-files '*.py')
```
dpaiton, line 29: can you keep the `rcfile` flag, just redirect it? I understand that the location now corresponds to the default, but I'd still like to keep this explicit so that anyone looking through our CI will be pointed to the file.

dpaiton (approved):
one small request, otherwise lgtm

PR #109, opened 2023-01-09
add agent.py in-code documentation

MazyGio:


dpaiton:
Can you add newlines between arguments & returns? I believe this is the convention we have most other places.

dpaiton (approved):

PR #110, opened 2023-01-10
Fixed rounding issues in the pricing model

jalextowle:
Prior to this PR, the pricing model performed all computations using floating point arithmetic. Since floating point numbers have a maximum range of 18 decimals, this led to situations in which pricing model calculations would be rounded up. These numerical errors would build up over multiple arithmetic operations, which led to sign flip errors.

The resolution to these errors is to add a minimum to input trade quantities, maximums differences to the reserves, and use the `Decimal` module for all calculations within the pricing model (while still accepting float inputs and returning float outputs).

### DIFF
--- a/src/elfpy/pricing_models/base.py
+++ b/src/elfpy/pricing_models/base.py
@@ -10,8 +10,7 @@ import elfpy.utils.price as price_utils
 
 # Set the Decimal precision to be higher than the default of 28. This ensures
 # that the pricing models can safely a lowest possible input of 1e-18 with an
-# opposing reserves of over 1 billion (although this should be viewed as a cap
-# to give the market room for interest to accrue over time).
+# reserves difference of up to 20 billion.
 decimal.getcontext().prec = 30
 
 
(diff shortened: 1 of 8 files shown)


src/elfpy/pricing_models/base.py
```diff
@@ -201,15 +208,46 @@ def calc_spot_price_from_reserves(
...
+        time_remaining: StretchedTime,
+    ) -> Decimal:
+        r"""
+        Calculates the spot price of base in terms of bonds.
+
+        The spot price is defined as:
+
+        .. math::
+            \begin{align}
+            p = (\frac{y + cz}{\mu z})^{-\tau}
```
jrhea, line 225: shouldn't the denominator be mu * x?

jrhea (changes requested):
I think the docs for calculate spot price from reserves has a mistake.

src/elfpy/pricing_models/base.py
jalextowle: Yeah, now that I look at this, I think the whole calculation is off. The numerator is also wrong (should at least be 2y + cz), but something else that seems off is that in the Allen Niemberg paper, they have $p = \frac{y}{\mu x}^t$, which would mean that our calculation should be: $p = (\frac{c \cdot (y + s)}{\mu z})^t \approx (\frac{c \cdot (2y + cz)}{\mu z})^t$. Let me know if this looks right to you, and I'll fix it. We also need to start using LP tokens in the price calculations rather than $s \approx 2y + cz$ but I'll create an issue for that.

jrhea: ya the reason it is to the power of t is b/c they look at it in terms of swapping base for PT.  I/we prefer to look at it the other way b/c it results in spot prices < 1

jalextowle: I'm going to resolve this since I updated the comment. As discussed offline, adding an extra `c` into the numerator makes the `without_fee_or_slippage` diverge further from the `without_fee` calculation, so it's a less accurate approximation. Thus, I'm keeping the spot price calculation the same.

src/elfpy/pricing_models/base.py
```diff
@@ -341,22 +379,28 @@ def check_input_assertions(
...
         """Applies a set of assertions to the input of a trading function."""
 
-        assert quantity.amount > 0, (
-            "pricing_models.check_input_assertions: ERROR: " f"expected quantity.amount > 0, not {quantity.amount}!"
+        assert quantity.amount >= 1e-18, (
+            "pricing_models.check_input_assertions: ERROR: "
+            f"expected quantity.amount >= 1e-18, not {quantity.amount}!"
         )
-        assert market_state.share_reserves >= 0, (
+        assert market_state.share_reserves >= 1e-18, (
```
dpaiton, line 386: I think it makes sense for us to use an enum for this since we repeat it everywhere and it has a meaning ("basically zero")

src/elfpy/pricing_models/base.py
```diff
@@ -341,22 +382,28 @@ def check_input_assertions(
...
             "pricing_models.check_input_assertions: ERROR: "
-            f"expected share_reserves >= 0, not {market_state.share_reserves}!"
+            f"expected share_reserves >= 1e-18, not {market_state.share_reserves}!"
```
dpaiton, line 391: Why can the bond_reserves == 0 but not share_reserves?

jalextowle: It's an artifact of the way that we bootstrap the markets (we initialize the market with some shares and the bonds are 0 for a small period of time). To be honest, I don't really see the rationale for the current method of bootstrapping instead of just having the AMM solve for the virtual bond reserves required to hit a target APR, but I didn't want to rock the boat.

src/elfpy/utils/price.py
```diff
@@ -127,25 +128,28 @@ def calc_spot_price_from_apr(apr: float, time_remaining: StretchedTime):
     return 1 / (1 + apr * time_remaining.normalized_time)  # price = 1 / (1 + r * t)
 
 
-# TODO: This should be updated to use StretchedTime.
-def calc_k_const(market_state: MarketState, time_elapsed):
+def calc_k_const(market_state: MarketState, time_remaining: StretchedTime):
```
dpaiton, line 131: can you add a return type hint to this signature?

src/elfpy/pricing_models/base.py
dpaiton: I see, makes sense. I agree it's not ideal, but it's the [boat we're in](https://youtu.be/_uHXS36LKNE).

dpaiton (approved):

PR #111, opened 2023-01-10
fix max short

wakamex:


src/elfpy/agent.py
```diff
@@ -53,18 +53,19 @@ def action(self, market: Market, pricing_model: PricingModel) -> list[MarketActi
...
         """
         if market.market_state.share_reserves == 0:
             return 0
-        max_pt_short = (
-            market.market_state.share_reserves * market.market_state.share_price / market.get_spot_price(pricing_model)
+        trade_results = pricing_model.calc_in_given_out(
+            out=Quantity(amount=market.market_state.share_reserves, unit=TokenType.PT),
+            market_state=market.market_state,
+            fee_percent=market.fee_percent,
+            time_remaining=market.position_duration,
```
jalextowle, line 66: It doesn't seem like this method ensures that the agent has sufficient base to cover the max short. In other words, it's entirely possible that this would result in the agent having negative base after making a max short (it will provide the same result regardless of the capital held in the agent's wallet).

Furthermore, this doesn't ensure that the market's invariants are satisfied after making this trade. If you look at the implementation of `get_max_long` in `src/elfpy/pricing_models/base.py`, you'll see some of the invariants that I'm talking about. Since this is a short, you won't have to worry about the rate going negative; however, you will need to worry about the buffer invariants.

wakamex: ya i figure those are done elsewhere, since this resides in the agent's logic, it's just looking at things from his perspective

wakamex: here's how it was used as a piece of the agent's logic (and nowhere else):

`self.create_agent_action(action_type="open_short", trade_amount=min(max_pt_short, self.pt_to_short))`
https://github.com/wakamex/elf-simulations/blob/7754095a7c17bf271e27df98d3167781476c29f6/src/elfpy/strategies/single_short.py#L48

is it used somewhere else now?

jalextowle: The problem in your example is that just because the `self.pt_to_short` is less than the `max_pt_short`, it doesn't mean that the agent has sufficient `base` to cover the maximum loss of the short (which is the amount that will be debited from the agent's wallet when the trade is put on). It shouldn't be hard to fix this, but it will need to be fixed.

You're right that the invariants can be checked elsewhere (since you can take the minimum of a market-level `get_max_short` and this `get_max_short`), but we'll need to add a separate pricing model level function that calculates the maximum short that can be put on given a `MarketState`. It should look very similar to `get_max_long`.

wakamex:
not needed anymore

PR #117, opened 2023-01-11
adds override test and reorganizes some of the config helper functions

dpaiton:
- Adds new test to make sure that config override works as expected
- fixes some order of operations issues with `vault_apr` & `num_trading_days` override
- moves config override out of sim_utils and into config_utils
- minor syntax improvements


jalextowle (approved):
I had one ask, but feel free to ignore if you want

PR #120, opened 2023-01-12
Fix the short accounting

jalextowle:
The primary objective of this PR was to fix the short accounting. I believe that this has been accomplished (for an initial version of the short accounting); however, the logic still needs to be well tested. This will be one of the focuses next week, so I am leaving that out of this PR.

An ancillary change made during this PR was to add pyright as part of our linting CI. As of right now, this has been marked as `continue-on-error: true`, but it should ultimately be enforced.

### DIFF
--- a/examples/example_main.py
+++ b/examples/example_main.py
@@ -36,8 +36,8 @@ class CustomShorter(Agent):
 
     def action(self, market: Market) -> list[Any]:
         """Implement a custom user strategy"""
-        block_position_list = list(self.wallet.shorts.values())
-        has_opened_short = bool(any((x < -1 for x in block_position_list)))
+        shorts = list(self.wallet.shorts.values())
+        has_opened_short = bool(any((short.balance > 0 for short in shorts)))
         can_open_short = self.get_max_pt_short(market) >= self.pt_to_short
         vault_apr = market.market_state.vault_apr
         action_list = []
--- a/src/elfpy/agent.py
+++ b/src/elfpy/agent.py
@@ -9,7 +9,7 @@ import logging
 import numpy as np
 
 from elfpy.utils.outputs import float_to_string
-from elfpy.wallet import Wallet
+from elfpy.wallet import Long, Wallet
 from elfpy.types import MarketAction, MarketActionType
 
 if TYPE_CHECKING:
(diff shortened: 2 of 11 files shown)


examples/example_main.py
```diff
@@ -36,8 +36,8 @@ def __init__(self, wallet_address: int, budget: int = 10_000) -> None:
 
     def action(self, market: Market) -> list[Any]:
         """Implement a custom user strategy"""
-        block_position_list = list(self.wallet.shorts.values())
-        has_opened_short = bool(any((x < -1 for x in block_position_list)))
+        shorts = list(self.wallet.shorts.values())
+        has_opened_short = bool(any((short.balance > 0 for short in shorts)))
```
dpaiton, line 40: Cool I was just looking at this and I have no idea what that logic was doing before lol. I fixed it differently in my current workflow, but this is good I think.

src/elfpy/wallet.py
```diff
@@ -25,12 +60,10 @@ class Wallet:
         The base assets that held by the trader.
     lp_tokens : float
         The LP tokens held by the trader.
-    longs : dict
+    longs : Dict[float, Long]
```
sentilesdal, line 63: why is this `Dict[float, Long]`?  Shouldn't it be num or str for the key?

sentilesdal: ah nvm, i see that the mint time is a float

src/elfpy/agent.py
```diff
@@ -106,36 +106,68 @@ def update_wallet(self, wallet_deltas: Wallet, market: Market) -> None:
...
-                        if mint_time in self.wallet[key]:  #  entry already exists for this mint_time, so add to it
-                            self.wallet[key][mint_time] += amount
+                        if mint_time in self.wallet.longs:  #  entry already exists for this mint_time, so add to it
+                            self.wallet.longs[mint_time].balance += long.balance
                         else:
-                            self.wallet[key].update({mint_time: amount})
-                    if self.wallet[key][mint_time] == 0:
-                        del self.wallet[key][mint_time]
+                            self.wallet.longs.update({mint_time: long})
+                    if self.wallet.longs[mint_time].balance == 0:
```
sentilesdal: as a protection against rounding errors, maybe have this be `balance <= 0`

src/elfpy/wallet.py
```diff
@@ -12,6 +12,41 @@
...
 
 
+@dataclass
+class Long:
+    """An open long position.
+
+    Arguments
+    ---------
+    balance : float
+        The amount of longs that are open.
```
sentilesdal, line 22: this is a weird statement, is the balance in base or in # of bonds?

src/elfpy/wallet.py
```diff
@@ -12,6 +12,41 @@
...
+
+
+@dataclass
+class Short:
+    """An open short position.
+
+    Arguments
+    ---------
+    balance : float
+        The amount of shorts that are open.
```
sentilesdal, line 38: same here, weird sentence.  I assume these are in bonds.  margin below is in base?

sentilesdal (approved):
accounting looks right to me

PR #121, opened 2023-01-14
Implemented `get_max_short`

jalextowle:
Implemented `get_max_short` variants for `PricingModel` and `Agent`.


src/elfpy/pricing_models/base.py
```diff
@@ -438,13 +435,99 @@ def get_max_long(
...
+                    last_maybe_max_long = (maybe_max_long, d_bonds)
                     if bond_percent == 1:
-                        return maybe_max_long
-                    last_maybe_max_long = maybe_max_long
+                        return last_maybe_max_long
                     bond_percent += step_size
 
         return last_maybe_max_long
 
+    def get_max_short(
```
wakamex, line 445: shouldn't the pricing model get_max_short be exact, if you assume an agent unconstrained by base in their wallet?

wakamex: is it because this allows us to check your invariants? if we can verify those are not needed, then we could switch to the closed form solution, correct? although that's not much of an improvement, since you use it as your starting point here anyway. what you have here is simply more general and covers more ground. no need to move away from it.

src/elfpy/pricing_models/base.py
```diff
@@ -375,17 +373,17 @@ def get_max_long(
         float
             The maximum amount of base that can be used to purchase bonds.
```
wakamex, line 374: we're returning a tuple here right? of `(maybe_max_long, d_bonds)` why is this necessary? the agent ignores the latter, right?

wakamex (approved):
i'm fine merging this now, wanna start using it to template some market tests.

my first 2 comments kinda answer themselves, and the third one is just a confusion about why max_long returns a tuple when it seems we only ever use one value. maybe it's for symmetry with max_short, where we do use both values?

PR #123, opened 2023-01-17
Test Flat + Curve

sentilesdal:
- rename files
- add hyperdrive test files
- add comments
- remove unnecessary paranthesis
- fix typo
- fix bug
- fix bug
- changes to test files
- add file to generate expected results


src/elfpy/pricing_models/base.py
```diff
@@ -205,7 +205,10 @@ def calc_spot_price_from_reserves(
             "pricing_models.calc_spot_price_from_reserves: ERROR: "
             f"expected share_reserves > 0, not {market_state.share_reserves}!",
         )
+        # s = y + c*z
```
jalextowle, line 208: Please add a TODO here. In general $s \neq y + cz$ and we'll want to update this soon to use `s = market_state.lp_reserves`

src/elfpy/pricing_models/hyperdrive.py
```diff
@@ -126,7 +126,7 @@ def calc_in_given_out(
             )
             market_result = MarketTradeResult(
                 d_base=-out.amount,
-                d_bonds=curve.market_result.d_bonds,
+                d_bonds=flat + curve.market_result.d_bonds,
```
jalextowle, line 129: We don't want to add the flat component to the market bonds as this introduces a convexity problem. The previous implementation was correct.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -353,12 +353,12 @@ def calc_in_given_out(
         # We precompute the YieldSpace constant k using the current reserves and
         # share price:
         #
-        # k = (c / μ) * (μ * z)**(1 - t) + (2y + cz)**(1 - t)
+        # k = (c / μ) * (μ * z)**(1 - tau) + (2y + cz)**(1 - tau)
```
jalextowle, line 356: You might as well use the unicode symbol for tau since we already do this for mu.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -353,12 +353,12 @@ def calc_in_given_out(
...
         k = price_utils.calc_k_const(market_state, time_elapsed)
         if out.unit == TokenType.BASE:
             in_reserves = market_state.bond_reserves + total_reserves
             out_reserves = market_state.share_reserves
-            d_shares = out.amount / market_state.share_price
+            d_shares = out.amount
```
jalextowle, line 361: Out is denominated in base not shares (just like how our smart contract will accept base rather than shares). The previous implementation was correct.

src/elfpy/pricing_models/hyperdrive.py
```diff
@@ -113,7 +113,7 @@ def calc_in_given_out(
             out=Quantity(amount=out.amount * time_remaining.normalized_time, unit=out.unit),
             market_state=market_state,
             fee_percent=fee_percent,
-            time_remaining=StretchedTime(days=365, time_stretch=time_remaining.time_stretch),
+            time_remaining=time_remaining,
```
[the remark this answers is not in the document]
sentilesdal: Ah right, this is what we were talking about last week.  Thanks for the clarification.

src/elfpy/pricing_models/hyperdrive.py
sentilesdal: Ok thanks

src/elfpy/pricing_models/yieldspace.py
sentilesdal: Ok look at this one again.  The next line of code, line 371 has:
```
              without_fee_or_slippage = (1 / spot_price) * market_state.share_price * d_shares
 ```
 
 since the output is base, we multiply by c since `x = c * z`

Comment (jalextowle) [reply src/elfpy/pricing_models/yieldspace.py:361]: Let's chat about this after all-hands.

File: src/elfpy/pricing_models/hyperdrive.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (sentilesdal) [reply src/elfpy/pricing_models/hyperdrive.py:116]: undid this

File: src/elfpy/pricing_models/base.py (hunk 1 above)
Comment (sentilesdal) [reply src/elfpy/pricing_models/base.py:208]: added TODO

File: src/elfpy/pricing_models/hyperdrive.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (sentilesdal) [reply src/elfpy/pricing_models/hyperdrive.py:116]: also added a TODO to change `days=365` to the length of the term during setup.

File: src/elfpy/pricing_models/hyperdrive.py (hunk 2 above)
Comment (sentilesdal) [reply src/elfpy/pricing_models/hyperdrive.py:129]: undid this.

File: src/elfpy/pricing_models/yieldspace.py (hunk 3 above)
Comment (sentilesdal) [reply src/elfpy/pricing_models/yieldspace.py:356]: I made this change, but it was cumbersome and I can see grepping etc being cumbersome in the future.  If you have no strong feelings I could to a search and replace all to get rid of greek symbols with alphabet spellings instead.

Comment (sentilesdal) [reply src/elfpy/pricing_models/yieldspace.py:356]: in a follow up PR i mean

Comment (dpaiton) [reply src/elfpy/pricing_models/yieldspace.py:356]: agreed, I'd rather us spell it out (`mu`, `tau`, etc)

Comment (jalextowle) [reply src/elfpy/pricing_models/yieldspace.py]: I prefer $\mu$ and $\tau$, but I don't care enough to argue about it.

File: src/elfpy/pricing_models/hyperdrive.py [hunk 6]
```diff
@@ -118,6 +118,7 @@ def calc_in_given_out(
             out=Quantity(amount=float(Decimal(out.amount) * Decimal(time_remaining.normalized_time)), unit=out.unit),
             market_state=market_state,
             fee_percent=fee_percent,
+            # TODO: don't hardcode days to 365, initialize to term length
```
jalextowle, line 121: IMO, this change should be part of this effort. We should have tests with non-365 day terms that capture the problems caused by our current normalization (again, this won't manifest on this line).

src/elfpy/pricing_models/base.py
```diff
@@ -321,9 +321,14 @@ def _calc_spot_price_from_reserves_high_precision(
             "pricing_models.calc_spot_price_from_reserves: ERROR: "
             f"expected share_reserves > 0, not {market_state.share_reserves}!",
         )
+
+        # TODO: in general s != y + c*z, we'll want to update this to have s = lp_reserves
+        # s = y + c*z
         total_reserves = Decimal(market_state.bond_reserves) + Decimal(market_state.share_price) * Decimal(
             market_state.share_reserves
         )
+        # p = ((y + s)/(u*z))^(-tau)
```
jalextowle, line 330: Let's use `mu` or $\mu$ instead of `u`.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -357,7 +357,7 @@ def calc_in_given_out(
         # We precompute the YieldSpace constant k using the current reserves and
         # share price:
         #
-        # k = (c / μ) * (μ * z)**(1 - t) + (2y + cz)**(1 - t)
+        # k = (c / μ) * (μ * z)**(1 - τ) + (2y + cz)**(1 - τ)
```
jalextowle, line 360: Let's pick $\tau$ or `tau` in this PR. I don't really care, but would be good to set the record straight. Let's do the same for picking between $\mu$ and `mu`.

tests/pricing_models/hyperdrive/hyperdrive_generate_expected_results.py
```diff
@@ -0,0 +1,256 @@
+# pylint: skip-file
```
jalextowle, line 1: I'm against merging unlinted files into the repo. It's fine to keep around a useful tool, but it needs to meet the repo's standards.

jalextowle: This file needs to be cleaned up before I'll approve.

I also just want to make sure that this is not replacing having your eyes on the inputs and outputs to make sure they are reasonable. I used a calculator (`python -i`) when I was writing the original pricing model tests, so it's totally fine to use another tool; however, we just want to make sure that we're carefully considering what we're putting into our test cases as we go along. I'm not suggesting that you aren't doing this, but I just think it's worth calling out.

tests/pricing_models/yieldspace/test_calc_in_given_out.py
```diff
@@ -155,21 +155,19 @@ def test_calc_in_given_out_success(self):
...
-                        without_fee_or_slippage=97.55458141947516,
-                        # d_z' = 1/u * (u/c*(k - (2*y + c*z - d_y)**(1-t)))**(1/(1-t)) - z
-                        # d_z' = 1/1 * (1/1*(302929.51067963685 - (2*100000 + 100000 - 100)**(1-0.0225358440315970471499308329778)))**(1/(1-0.0225358440315970471499308329778)) - 100000
-                        #         = 97.55601990513969
-                        without_fee=97.55601990513969,
-                        # fee is 20% of discount before slippage = (100-97.55458141947516)*0.2 = 0.48908371610496887
-                        yieldspace_fee=0.48908371610496887,
-                        # with_fee = d_z' + fee = 97.55601990513969 + 0.4887960189720616 = 98.04481592411175
-                        yieldspace_with_fee=98.04510362124466,
-                    )
```
jalextowle, line 172: Let's make this change everywhere in this file and `test_calc_out_given_in`.

tests/pricing_models/hyperdrive/hd_test_calc_in_given_out.py
```diff
@@ -0,0 +1,770 @@
...
+
+    out: Quantity
+    market_state: MarketState
+    fee_percent: float
+    time_remaining: StretchedTime
+
+    __test__ = False  # pytest: don't test this class
+
+
+class TestCalcInGivenOut(unittest.TestCase):
```
jalextowle, line 57: We're duplicating a lot of code between these test files. My preference is that we use common inputs and provide the different outputs as a dictionary keyed by the model's name. That way we keep everything in one file and all of the pricing models benefit when we add cases.

tests/pricing_models/hyperdrive/hd_test_calc_in_given_out.py
```diff
@@ -0,0 +1,770 @@
...
+                expected_result.hyperdrive_fee,
+                err_msg=f"test {test_case.test_id} unexpected hyperdrive fee",
+            )
+            np.testing.assert_almost_equal(
+                trade_result.breakdown.with_fee,
+                expected_result.hyperdrive_with_fee,
+                err_msg=f"test {test_case.test_id} unexpected hyperdrive with_fee",
+            )
+
+    # # TODO: This should be refactored to be a test for check_input_assertions and check_output_assertions
```
jalextowle, line 588: This will need to be uncommented before it's merged.

jalextowle (changes requested):
I'll need to do a final review after this.

src/elfpy/pricing_models/hyperdrive.py
dpaiton: Agreed, although IMO the test PR should not include new features so I'd say make that its own PR

tests/pricing_models/hyperdrive/hyperdrive_generate_expected_results.py
sentilesdal: Yeah totally.  I'm not actually running anything through a calculator, but I'm visually inspecting things to make sure they look approximately sane.  I think a lot of the value for these tests come from the fact that the output is generated by a function that has all the math inlined, so we can inspect that the implementation code is doing what we want it to do.

dpaiton (approved):
Looks good after you address Alex's remaining concerns (add the failure tests, fix `mu` etc, & move the test helper file out of the repo).

src/elfpy/pricing_models/base.py
sentilesdal: I went with `mu`

src/elfpy/pricing_models/yieldspace.py
sentilesdal: I went with `mu` and `tau`.  I cleaned a LOT of this up:
`τ`, `t`, `T`, `tau` -> `tau`.
`T`, `t_stretch` -> `t_stretch`
`t` or `t_d` is now only used for normalized time
`μ`, `u` -> `mu`

tests/pricing_models/hyperdrive/hd_test_calc_in_given_out.py
sentilesdal: Added failure tests

src/elfpy/pricing_models/hyperdrive.py
jalextowle: I'm okay with testing PRs that fix the problems they test; however, I'm fine with this approach in this case.

src/elfpy/pricing_models/yieldspace.py
```diff
@@ -303,7 +303,7 @@ def calc_in_given_out(
             \begin{cases}
             c (\frac{1}{\mu} (\frac{k - (2y + cz - \Delta y)^{1-t}}{\frac{c}{\mu}})^{\frac{1}{1-t}} - z),
```
jalextowle, line 304: While we're fixing the use of time in comments, let's do it everywhere. For example this equation and any others you see that still use `t` instead of `\tau`.

tests/pricing_models/yieldspace_test_calc_out_given_in.py
```diff
@@ -136,12 +136,12 @@ def test_calc_out_given_in_success(self):
...
                     #
-                    #   k = (c / μ) * (μ * (z + d_z)) ** (1 - t) + (2 * y + c * z - d_y') ** (1 - t)
-                    #     = 100_100 ** (1 - T) + (300_000 - d_y') ** (1 - T)
+                    #   k = (c / mu) * (mu * (z + d_z)) ** (1 - tau) + (2 * y + c * z - d_y') ** (1 - tau)
+                    #     = 100_100 ** () + (300_000 - d_y') ** (1 - tau)
                     #
                     # Solving for d_y, we get the following calculation:
                     #
-                    #   d_y' = 300_000 - (k - 100_100 ** (1 - T)) ** (1 / (1 - T))
+                    #   d_y' = 300_000 - (k - 100_100 ** ()) ** (1 / (1 - tau))
```
jalextowle, line 144: Nit: `(1 - T)` should have been updated to `(1 - tau)` but it's empty here.

tests/pricing_models/yieldspace_test_calc_out_given_in.py
```diff
@@ -206,12 +206,12 @@ def test_calc_out_given_in_success(self):
                     # We want to solve for the amount of bonds out given the
                     # amount of shares coming in, so we set up the problem as:
                     #
-                    #   k = (c / μ) * (μ * (z + d_z)) ** (1 - t) + (2 * y + c * z - d_y') ** (1 - t)
-                    #     = 110_000 ** (1 - T) + (300_000 - d_y') ** (1 - T)
+                    #   k = (c / mu) * (mu * (z + d_z)) ** (1 - tau) + (2 * y + c * z - d_y') ** (1 - tau)
+                    #     = 110_000 ** () + (300_000 - d_y') ** (1 - tau)
```
jalextowle, line 210: I'd recommend that you look through the diffs and find all cases in which `(1 - T)` => `(1 - tau)` and all other changes like this.

tests/pricing_models/hyperdrive_test_calc_in_given_out.py
```diff
@@ -0,0 +1,736 @@
...
+                ),
+                # in_base=97.58374088542769
+                # in_pt=102.54051519598579
+                TestResultCalcInGivenOutSuccess(
+                    without_fee_or_slippage=102.53971546251678,
+                    without_fee=102.54051519598579,
+                    fee=0.25397154625167895,
+                    with_fee=102.79448674223747,
+                ),
+            ),  # end of test 1
```
jalextowle, line 88: This definitely doesn't need to be fixed in this PR, but I don't like these comments personally. I'm of the opinion that they just add clutter and don't really help to distinguish between the cases. Having said this, I know you're just following the old pattern, so no need to address this here.

jalextowle (commented):
Thanks for the hard work on this! This is really starting to come together.

A few more comments here. I am okay with approving once these are addressed, but the tests need to be merged (as one of my earlier comments asked for) in this PR or a follow-up PR. It is really difficult to compare the outputs from different pricing models when they live in different files. If they live together, this issue is trivial _and_ we will significantly reduce code duplication and maintenance.

tests/pricing_models/yieldspace_test_calc_out_given_in.py
sentilesdal: oops! good catch

tests/pricing_models/yieldspace_test_calc_out_given_in.py
```diff
@@ -206,12 +206,12 @@ def test_calc_out_given_in_success(self):
                     # We want to solve for the amount of bonds out given the
                     # amount of shares coming in, so we set up the problem as:
                     #
-                    #   k = (c / μ) * (μ * (z + d_z)) ** (1 - t) + (2 * y + c * z - d_y') ** (1 - t)
-                    #     = 110_000 ** (1 - T) + (300_000 - d_y') ** (1 - T)
+                    #   k = (c / mu) * (mu * (z + d_z)) ** (1 - tau) + (2 * y + c * z - d_y') ** (1 - tau)
+                    #     = 110_000 ** (t - tau + (300_000 - d_y') ** (1 - tau)
```
jalextowle, line 210: Nit: This comment isn't right. It should be `(1 - tau)`. I hate to be so nitpicky, but it needs to be done right before we merge it in. If you read through your own diff, you should be able to catch problems like this before review.

jalextowle: For things like this, I'd recommend reviewing the changes before you even commit as that way you won't waste your own cycles waiting for review and won't waste the reviewers cycles either.

jalextowle (approved):
I am approving because it's ready after my last comment is addressed. Having said this, it does need to be addressed, and I'd like you to make sure the comments look correct before merging.

PR #125, opened 2023-01-18
market tests and implement partial shorts

wakamex:
add tests of market functions (open/close long/short) and update market accounting to work with partial shorts


wakamex:
lgtm, could move `validate_custom_parameters` into a util function if desired

wakamex:
nevermind, going to add market tests here

src/elfpy/markets.py
```diff
@@ -109,22 +109,22 @@ def trade_and_update(self, agent_action: MarketAction) -> Wallet:
...
             market_deltas, agent_deltas = self.close_long(
                 wallet_address=agent_action.wallet_address,
-                trade_amount=agent_action.trade_amount,
+                trade_amount=agent_action.trade_amount,  # in bonds: that's the thing in your wallet you want to sell
                 mint_time=agent_action.mint_time,
             )
         elif agent_action.action_type == MarketActionType.OPEN_SHORT:  # sell PT to open short
             market_deltas, agent_deltas = self.open_short(
-                wallet_address=agent_action.wallet_address,
+                wallet_address=agent_action.wallet_address,  # in bonds: that's the thing you want to short
```
dpaiton, line 122: I think you meant this comment to be on `trade_amount`

src/elfpy/markets.py
```diff
@@ -109,22 +109,22 @@ def trade_and_update(self, agent_action: MarketAction) -> Wallet:
...
                 trade_amount=agent_action.trade_amount,
             )
         elif agent_action.action_type == MarketActionType.CLOSE_SHORT:  # buy PT to close short
             market_deltas, agent_deltas = self.close_short(
-                wallet_address=agent_action.wallet_address,
+                wallet_address=agent_action.wallet_address,  # in bonds: that's the thing you owe, and need to buy back
```
dpaiton, line 127: I think you meant this comment to be on `trade_amount`

src/elfpy/markets.py
```diff
@@ -299,16 +299,16 @@ def close_short(
             d_token_asset=trade_result.market_result.d_bonds,
             d_bond_buffer=-trade_amount,
         )
+        d_worst_case_scenario = trade_amount  # in the worst case for the short, p=1 and they owe the face value
```
dpaiton, line 302: Please just use the `trade_amount` variable instead of introducing this variable, and use comments & docstrings to explain how/why/when it is the worst_case_scenario.

src/elfpy/markets.py
```diff
@@ -370,7 +370,7 @@ def open_long(
     def close_long(
         self,
         wallet_address: int,
-        trade_amount: float,
+        trade_amount: float,  # in bonds
```
dpaiton, line 373: No need to do it in this PR, but the desire to add the unit comments seems like a good enough argument to use the `types.Quantity` dataclass here, which has a unit attribute.

dpaiton: @wakamex could you add a TODO in the docstrings to remind us?

src/elfpy/types.py
```diff
@@ -153,11 +153,11 @@ def __str__(self):
                 if value != 0:
                     output_string += f" {key}: "
                     if isinstance(value, float):
-                        output_string += f"{float_to_string(value)}"
+                        output_string += f"{value}"
```
dpaiton, line 156: If you are leaning away from using this in logs it then I support removing `float_to_string` all-together. I think it's better to use the native fstring formatting when needed anyway. Feel free to mark this as a todo (e.g. in clickup or a github issue) and hold off for a follow-up PR.

src/elfpy/utils/outputs.py
```diff
@@ -114,7 +114,7 @@ def close_logging(delete_logs=True):
     if delete_logs:
         for handler in logging.getLogger().handlers:
             handler.close()
-            if isinstance(handler, logging.FileHandler):
+            if hasattr(handler, "baseFilename") and not isinstance(handler, logging.StreamHandler):
                 if os.path.exists(handler.baseFilename):
```
dpaiton, line 118: Reminder to change this to `getattr` to resolve pyright concern.

tests/pricing_models/test_calc_in_given_out.py
```diff
@@ -2,6 +2,7 @@
 Testing for the calc_in_given_out of the pricing models.
 """
 
+# pylint: disable=too-many-lines
```
dpaiton, line 5: looks like you can delete this line -- Matt added the pylint disable in his own PR.

pyproject.toml
```diff
@@ -34,4 +34,6 @@ max-line-length = "120"
 
 [tool.black]
 line-length = "120"
-extend-exclude = "\\.ipynb"
\ No newline at end of file
+extend-exclude = "\\.ipynb"
+
+[tool.pyright]
```
dpaiton, line 39: I think it wants a newline at the end of the file.

src/elfpy/markets.py
```diff
@@ -196,9 +197,17 @@ def open_short(
         trade_amount: float,
     ) -> tuple[MarketDeltas, Wallet]:
         """
-        take trade spec & turn it into trade details
-        compute wallet update spec with specific details
-        will be conditional on the pricing model
+        ##############################
```
dpaiton, line 200: I don't think all the `#` are necessary -- it's in a docstring already

src/elfpy/types.py
wakamex: removed it in this PR (only used in like 2 places still)

src/elfpy/markets.py
```diff
@@ -247,9 +251,15 @@ def close_short(
         mint_time: float,
     ) -> tuple[MarketDeltas, Wallet]:
         """
-        take trade spec & turn it into trade details
-        compute wallet update spec with specific details
-            will be conditional on the pricing model
+        ###########################
```
dpaiton, line 254: See above regarding `#`

tests/test_market.py
```diff
@@ -0,0 +1,431 @@
...
+                    np.testing.assert_almost_equal(
+                        agent_balance_in_pts,
+                        expected_balance,  # passed in as pt's
+                        err_msg=(
+                            f"agent #{custom_agent_index}'s {account} does not equal {expected_balance}"
+                            f", instead we have (base,pt)=({agent_balance_in_base},{agent_balance_in_pts})"
+                        ),
+                    )
+
+    # I'd make this a static method but you can't call assertEqual on a static method
```
dpaiton, line 100: I don't think the comment about it being a static method is useful in the repo long-term. I'd say take it out if it's all the same to you. The function name should be more descriptive / useful, like `assert_equal_and_log` or something. Type hints should be on all arguments. We (apparently) have much lower standards for docstrings and typehints in tests -- they're not really used anywhere. So I won't ask that you add typehints in this whole file. But in a given function if you're going to do it for one you should just do it for all of them. For me, and my ocd. Thanks!

tests/test_market.py
```diff
@@ -0,0 +1,423 @@
...
+                    np.testing.assert_almost_equal(
+                        agent_balance_in_pts,
+                        expected_balance,  # passed in as pt's
+                        err_msg=(
+                            f"agent #{custom_agent_index}'s {account} does not equal {expected_balance}"
+                            f", instead we have (base,pt)=({agent_balance_in_base},{agent_balance_in_pts})"
+                        ),
+                    )
+
+    def assert_equal_and_log(self, thing: str, expected: Any, actual: Any):
```
dpaiton, line 101: Can you change `thing` to a more descriptive variable name? I am not sure what it is supposed to be... the name of what is being compared?

tests/test_market.py
```diff
@@ -0,0 +1,423 @@
...
+        """Compare actual thing to expected thing"""
+        logging.debug("=== %s ===\nexpected %s: %s\nactual %s: %s", thing.upper(), thing, expected, thing, actual)
+        self.assertEqual(expected, actual, f"{thing} do not match")
+
+    def compare_deltas(self, actual_deltas: Deltas, expected_deltas: Deltas):
+        """Compare actual deltas to expected deltas"""
+        self.assert_equal_and_log("market deltas", expected_deltas.market_deltas, actual_deltas.market_deltas)
+        self.assert_equal_and_log("agent deltas", expected_deltas.agent_deltas, actual_deltas.agent_deltas)
+
+    def compare_wallets(self, actual_wallet: Wallet, expected_wallet: Wallet):
```
dpaiton, line 111: This is defined and then never used (just referenced in a commented line). I think it would be better to just comment out line 113: `self.assert_equal_and_log("wallet", expected_wallet, actual_wallet)` below and delete this function.

tests/test_market.py
```diff
@@ -0,0 +1,423 @@
...
+        """Test market trades result in the expected wallet balances"""
+        simulator = self.set_up_test(agent_policies=[agent_policy])
+        agent = simulator.agents[1]
+        market_deltas, agent_deltas = simulator.market.open_short(
+            wallet_address=1,
+            trade_amount=agent.amount_to_trade,  # in bonds: that's the thing you want to short
+        )
+        # peek inside the agent's wallet and see how many bonds they have
+        amount_of_bonds_sold = agent_deltas.shorts[0].balance
+        # sell pthose bonds to close the short (partial amount if selected)
```
dpaiton, line 169: is `pthose` a typo?

tests/test_market.py
```diff
@@ -0,0 +1,423 @@
...
+        trade_amount = amount_of_bonds_sold * partial if partial else amount_of_bonds_sold
```
dpaiton, line 170: I think it would be better to just set the default of `partial` to `1` and then you don't need the conditional.

Beyond that, checks for `None` should be done as `if partial is not None`. See here:
https://peps.python.org/pep-0008/#programming-recommendations

tests/test_market.py
```diff
@@ -0,0 +1,423 @@
...
+        actual_deltas = Deltas(market_deltas=market_deltas, agent_deltas=agent_deltas)
+        self.compare_deltas(actual_deltas=actual_deltas, expected_deltas=expected_deltas)
+        # TODO: do we want to test the final value in the wallet?
+        # we're not actually updating wallets here, but a full simulation test would
+        # we could test the final wallet value there much more easily
+        # self.compare_wallets(actual_wallet=agent.wallet, expected_wallet=expected_wallet)
+        output_utils.close_logging(delete_logs=delete_logs)
+
+
+class MarketTestsOneFunction(BaseMarketTest):
```
dpaiton, line 187: We have been including the equations that result in expected values (`trade_result`, `fees_paid`) in the test docs. Can you do that here?

tests/test_market.py
```diff
@@ -0,0 +1,423 @@
...
+    # TODO: sweep?
+    # possible sweep
+    # agent_policies = [f"single_long:amount_to_trade={amount}" for amount in [100, 10_000, 1_000_000, 100_000_000]]
+
+
+class MarketTestsOneTimestep(BaseMarketTest):
+    """Tests of market trades that execute the simulation for one timestep only"""
+
+    def test_100_long(self):
+        """self-explanatory"""
```
dpaiton, line 419: Please include a docstring that explains the purpose of the test.

tests/test_parameters.py
```diff
@@ -49,7 +50,7 @@ def run_custom_parameters_test(self, agent_policies, expected_result, delete_log
                 custom_agent_index = all_agent_index - number_of_init_agents  # identify which custom agent we are on
                 expected_result_dict = expected_result[custom_agent_index]
                 for key, value in expected_result_dict.items():  # for each custom parameter to check
-                    np.testing.assert_equal(
+                    np.testing.assert_almost_equal(
```
dpaiton, line 53: What changed such that this test result can no longer be equal?

tests/utils_for_tests.py
```diff
@@ -8,12 +8,48 @@
...
 import logging
 from importlib import import_module
+import numpy as np
 
 from elfpy.simulators import Simulator
 from elfpy.utils import sim_utils
 import elfpy.utils.parse_config as config_utils
 
 
+def float_to_string(value, precision=3, min_digits=0, debug=False):
```
dpaiton, line 18: If you think this function would be useful as a potential output utility, then I'd suggest putting it in the output utils file. I don't really see a reason to put it here, especially considering that it is not being called. Also, keep in mind that if you delete it and later find a use for it you can always look in the git history to find it and revive it.

dpaiton:
Overall it's a great test suite, and was clearly no small feat. Thanks a lot! The suggestions are hopefully all quick fixes -- DM me if you want to go over any of them.

tests/test_market.py
wakamex: ya it's just a display string

tests/test_market.py
wakamex: setting parameter to false avoids this issue, i believe

tests/test_market.py
wakamex: something like lines 189-195 in test_get_max? since we're testing that elsewhere, i don't see the point of adding it here

tests/test_parameters.py
wakamex: not sure, set it back. these values are simply assigned, so should be no rounding issues.

tests/utils_for_tests.py
wakamex: i put this in utils_for_tests for now, which also lives outside of the package. is that sufficient for now?

tests/test_market.py
dpaiton: Just asking for you to note where the numbers came from

dpaiton (approved):
looks great, thanks Mihai & Mazy!

PR #126, opened 2023-01-19
Simulator Initialization Refactor

jalextowle:
As a bonus, this PR fixes all of the outstanding pyright errors in the codebase and makes pyright a mandatory CI job.


src/elfpy/simulators.py
```diff
@@ -36,23 +36,25 @@ def __init__(
         self,
```
dpaiton, line 36: Adding the `agents` arg on init means the user has to pass two empty dictionaries to construct a Simulator that doesn't have any agents. It also seems like a likely point of confusion, we would have to be careful with documentation to explain the difference between the `init_agents` and `agents` args. I think we should be in agreement here as it sets expecations for the typical user workflow, so let's discuss. Here are two ideas for alternatives:

1. What do you think about passing only `init_agents` on Simulator initialization, and then creating a helper function `add_agents` that the user can call to add additional agents. This is more transparent than the previous workflow (which was `simulator.agents = get_example_agents(...)`), and addresses my above two complaints.

2. We could also remove the concept of `init_agents` entirely. Simulator is constructed without any agents passed.  The workflow would be then to always construct a Simulator without agents, and then manually add agents & execute trades as you see fit. Wrapper functions, like`run_simulation`, would include a check that the Simulator has agents. This way a user can be more purposeful about varying the number of agents in any given moment. This supports the notion that `collect_and_execute_trades` and `run_simulation` are wrapper functions, but not required for running simulations. One could make loops over days & blocks in an external script, and execute trades with varying sets of agents on any given day. The typical workflow (e.g. as set up in `example_main.py`) would then be to create the simulator, create `init_agent`, add `init_agent` to `simulator.agents`, execute trades, add more agents to `simulator.agents`, call `run_simulation`.

src/elfpy/types.py
```diff
@@ -257,16 +263,18 @@ class RandomSimulationVariables:
...
+    vault_apr: list = field(metadata=to_description("yield bearing source APR"))
+    init_vault_age: float = field(metadata=to_description("fraction of a year since the vault was opened"))
+    init_share_price: float = field(
+        default=0, metadata=to_description("initial market share price for the vault asset")
+    )
 
     def __post_init__(self):
         """init_share_price is a function of other random variables"""
-        if self.init_share_price is None:
+        if self.init_share_price == 0:
```
dpaiton, line 277: Why did you not like the `None` default? After this change, what would a user do if they actually want it to be 0?

src/elfpy/utils/sim_utils.py
```diff
@@ -197,28 +197,29 @@ def setup_vault_apr(config, rng):
         list of apr values that is the same length as num_trading_days
     """
     if isinstance(config.market.vault_apr, dict):  # dictionary specifies parameters for the callable
-        allowable_keys = ["constant", "uniform", "geometricbrownianmotion"]
```
dpaiton, line 200: much better, thanks

src/elfpy/simulators.py
jalextowle: I'm not a fan of (1). What do you think about adding an `initialize` method that does what the constructor does in this PR and otherwise adopting solution (2)? This would make sure that the agents list is empty to avoid any possible conflicts with the more manual workflow.

dpaiton: yup I like that. If it's just adding agents then maybe `initialize_agents` as a better name?

src/elfpy/types.py
dpaiton: None can be included in typing using the `Optional` type. https://docs.python.org/3/library/typing.html#typing.Optional

Can you use that and allow the None default?

I would rather not initialize it to a value out of principle, because it is opaque as to whether that value means "not initialized" or whether it is a reasonable initialization.

jalextowle: The issue with this is that we then need to check that it isn’t None everywhere that it is used (this is what by the vanilla approach not being typesafe). I don’t think the principle is worth it with this in mind, but let me know if you disagree.

examples/example_main.py
```diff
@@ -56,14 +55,9 @@ def action(self, market: Market) -> list[Any]:
         return action_list
 
 
-def get_example_agents(num_new_agents: int, num_existing_agents: int = 0) -> list[Agent]:
+def get_example_agents(new_agents: int, existing_agents: int = 0) -> list[Agent]:
     """Instantiate a set of custom agents"""
-    agents = []
-    for wallet_address in range(num_existing_agents, num_new_agents + 1):
-        agent = CustomShorter(wallet_address)
-        agent.log_status_report()
```
dpaiton, line 64: I think we should keep this step in.

src/elfpy/types.py
```diff
@@ -257,12 +263,16 @@ class RandomSimulationVariables:
...
     # dataclasses can have many attributes
     # pylint: disable=too-many-instance-attributes
-    target_liquidity: float = field(metadata="total size of the market pool (bonds + shares)")
-    target_pool_apr: float = field(metadata="desired fixed apr for as a decimal")
-    fee_percent: float = field(metadata="percent to charge for LPer fees")
-    vault_apr: list = field(metadata="yield bearing source APR")
-    init_vault_age: float = field(metadata="fraction of a year since the vault was opened")
-    init_share_price: float = field(default=None, metadata="initial market share price for the vault asset")
+    target_liquidity: float = field(metadata=to_description("total size of the market pool (bonds + shares)"))
+    target_pool_apr: float = field(metadata=to_description("desired fixed apr for as a decimal"))
```
dpaiton, line 267: Might as well fix the rest of the `metadata` entries in this file, since you're creating `to_description`

src/elfpy/utils/sim_utils.py
```diff
@@ -24,6 +26,61 @@
...
+        The random number generator that should be used in these simulations.
+        We pass this in to avoid restarting the generator after using it to
+        setting up the agents.
+    config : Config
+        The simulator config.
+    agents : list[Agent]
+        The agents to that should be used in the simulator.
+    """
+    # Sample the random simulation arguments.
+    random_sim_vars = get_random_variables(config, rng)
```
dpaiton, line 45: It looks like you moved `get_random_variables` out of this function and into `simulators.py`

src/elfpy/simulators.py
```diff
@@ -1,26 +1,104 @@
...
+from elfpy.types import MarketDeltas, RandomSimulationVariables, SimulationState
 
 if TYPE_CHECKING:
     from elfpy.agent import Agent
     from elfpy.markets import Market
-    from elfpy.pricing_models.base import PricingModel
     from elfpy.utils.config import Config
 
 
+def setup_vault_apr(config, rng):
```
dpaiton, line 22: What is the reason for putting `setup_vault_apr` and `get_random_variables` here instead of in `sim_utils`?

dpaiton (changes requested):
Solid improvement to the workflow, thanks! I posted a few minor requests & questions before approval.

src/elfpy/simulators.py
jalextowle: Circular imports.

src/elfpy/types.py
jalextowle: These are the only entries that break pyright, but sure I'll make them consistent.

tests/test_utils.py
```diff
@@ -9,8 +9,6 @@
 import logging
 from importlib import import_module
 
-import numpy as np
```
dpaiton, line 12: The original file name was intentional -- we have tests for `utils` that exist in `src`, so the file was called `utils_for_tests` to be differentiated from a file that would test the project utils.

jalextowle: I got that, but `test_utils` is a pretty standard file name (and this name sounds bad). My personal preference would be for it to live in `elfpy`, but I'm not going to fight about it and will change it back.

src/elfpy/simulators.py
dpaiton: If that is a result of typing you can resolve circular imports by placing the import statements behind a `TYPE_CHECKING` flag, which is already done for other imports in `simulators.py`. I would prefer this for a few reasons: `simulators.py` can be limited in scope to just the simulator class; all of the `get` functions can live together as utils; and imports are more consistent between the chief class objects and helper functions.

dpaiton: That being said, if the problem is not resolved by the TYPE_CHECKING flag then I'm fine with punting this to a later PR

dpaiton (approved):
Approving this, but check out the circular import thing before you merge please.

PR #128, opened 2023-01-19
test cleanup

wakamex:
allows me to easily dump in market tests

- create agents with custom parameters, like easily varying their `amount_to_lp` and `pt_to_short`
- remove code duplication by merging base_lp_test into base_run_test with more parameters
- fix bug with init_agents passed as reference then relying on its value not changing
- give agents a name so you can count which ones are init_lp for instance, needed to skip over them
- add 3.11 python testing, I was getting errors but it was passing on here
- add `close_logging` method to two tests where it was needed, shutting down logging and closing handlers (`test_trade` and `test_sim`)
- use adder method for adding logging handler, so pytest doesn't complain
```
./code/elfpy/src/elfpy/utils/outputs.py:101: ResourceWarning: unclosed file <_io.TextIOWrapper name='/code/elfpy/.logging/test_trades.log' mode='a' encoding='UTF-8'>

  logging.getLogger().handlers = [
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
- fix deterministic mode
- [move repetitive testing into test_utils](https://github.com/element-fi/elf-simulations/pull/125/commits/5348af7f47c95518d960e017fac5d6e4b9bad036)


tests/test_market.py
```diff
@@ -0,0 +1,95 @@
...
+        output_utils.setup_logging(log_filename=".logging/test_trades.log", log_level=logging.DEBUG)
+        # load default config
+        override_dict = {
+            "pricing_model_name": "Yieldspace",
+            "target_liquidity": 10e6,
+            "fee_percent": 0.1,
+            "target_pool_apr": 0.05,
+            "vault_apr": {"type": "constant", "value": 0.05},
+            # minimal simulation steps, we only care to investigate the first day's trades
+            "num_trading_days": 10,
```
dpaiton, line 38: running for 10 days with logging level set to DEBUG could take some time. I recommend lowering this to the minimum number of days that is needed to accomplish the testing task (probably 3?)

tests/test_sim.py
```diff
@@ -34,6 +34,17 @@ def setup_logging(log_level=logging.DEBUG):
         log_filename = ".logging/test_sim.log"
         output_utils.setup_logging(log_filename, log_level=log_level)
 
+    @staticmethod
+    def close_logging(delete_logs=True):
```
dpaiton, line 38: import this from `utils/outputs.py`

src/elfpy/utils/testing.py
```diff
@@ -0,0 +1,95 @@
+""" Implements functions that are useful for testing """
```
dpaiton, line 1: change this file to be located in tests, e.g. `tests/wrappers_for_testing.py`

tests/test_parameters.py
```diff
@@ -0,0 +1,101 @@
...
+        number_of_init_agents = 0  # count number of init agents so we can skip over them
+        for all_agent_index, agent in simulator.agents.items():  # loop over all agents
+            if agent.name == "init_lp":
+                number_of_init_agents += 1
+            else:  # only for custom agents, loop across them and check their parameters
+                custom_agent_index = all_agent_index - number_of_init_agents  # identify which custom agent we are on
+                expected_result_dict = expected_result[custom_agent_index]
+                for key, value in expected_result_dict.items():  # for each custom parameter to check
+                    np.testing.assert_almost_equal(
+                        getattr(agent, key),
```
dpaiton, line 54: is there a reason why you would not expect these to be exactly equal?

dpaiton (approved):

tests/test_parameters.py
wakamex: only python rounding. i think default is 1e-7, probably could be lower?

wakamex: wait no there's no rounding here

PR #129, opened 2023-01-19
Make the simulator more usable for fixtures

jalextowle:
Previously, fixture writers needed to invoke the agent machinery in an unnatural way to get the benefits of using the simulator for tests with a fixed set of trades and agents. Now, these consumers can simply call `execute_trades` without worrying about setting up agents to perform the trades they need executed.


dpaiton (approved):
these are super helpful funcs, thanks!

PR #132, opened 2023-01-23
Combine calc_in_given_out tests

sentilesdal:
combines tests for yieldspace and hyperdrive so the results are easier to compare


src/elfpy/pricing_models/base.py
```diff
@@ -89,6 +89,11 @@ def model_name(self) -> str:
         """Unique name given to the model, can be based on member variable states"""
         raise NotImplementedError
 
+    @abstractmethod
+    def model_type(self) -> str:
+        """Unique identifier given to the model, should be lower snake_cased name"""
+        raise NotImplementedError
+
```
jalextowle, line 96: If the model type is always the lower cased version of the name, why don't we just implement this as:

```
return self.model_name().lower()
```

tests/pricing_models/test_calc_in_given_out.py
```diff
@@ -0,0 +1,1097 @@
...
+                # fee is 10% of discount before slippage = (100-97.55458141947516)*0.1 = 0.24454185805248443
+                fee=0.24454185805248443,
+                # with_fee = d_z' + fee = 97.55601990513969 + 0.24454185805248443 = 97.80056176319217
+                with_fee=97.80056176319218,
+            ),
+            hyperdrive=TestResultCalcInGivenOutSuccess(
+                without_fee_or_slippage=97.58591137152354,
+                without_fee=97.5866001243412,
+                fee=0.24140886284764632,
+                with_fee=97.82800898718885,
```
jalextowle, line 438: I think we'll ultimately consider this result to be incorrect since this discrepancy arises from our improper use of a normalizing constant for time.

jalextowle (approved):
LGTM

tests/pricing_models/test_calc_in_given_out.py
sentilesdal: ok cool, when we make the change to time, we'll have to fix the tests.

PR #135, opened 2023-01-24
Add calc_out_given_in tests for hyperdrive

sentilesdal:
Adds tests for hyperdrive's `calc_out_given_in` function.  I also went through and added comments to the failure tests and tweaked a few values that were causing the failure for the wrong reason.


tests/pricing_models/test_calc_in_given_out.py
```diff
@@ -2,9 +2,11 @@
 Testing for the calc_in_given_out of the pricing models.
 """
 
+# pylint: disable=too-many-lines
```
dpaiton, line 5: accidentally have this twice

tests/pricing_models/test_dataclasses.py
```diff
@@ -43,8 +44,6 @@ def __getitem__(self, key):
     yieldspace: TestResultCalcInGivenOutSuccess
     hyperdrive: TestResultCalcInGivenOutSuccess
 
-    __test__ = False  # pytest: don't test this class
-
 
 @dataclass
 class TestCaseCalcInGivenOutFailure:
```
jalextowle, line 49: Since we have a types module for these tests, we may as well make `in_` and `out` quantity and delete the duplicated types.

tests/pricing_models/test_calc_out_given_in.py
```diff
@@ -0,0 +1,1611 @@
...
+                        # dy' = 2*y + c*z - (k - (c / u) * (u * (z + delta_z*t))**(1 - tau_full))**(1 / (1 - tau_full))
+                        #       + c * delta_z * (1 - t)
+                        # dy' = 199900.0 + 100050.0 - (229443.78725685307 - 1.0 * (1 * (100050.0 + 100.0*0.5))**(0.9549283119368059))**(1.0471990279267966)
+                        #       + 1 * 100.0 * 0.5
+                        without_fee=102.5357591957436,
+                        # fee = ((1 / p) - 1) * phi * c * d_z
+                        # fee = (1.050731163045899 - 1) * 0.01 * 1 * 100.0
+                        fee=0.025365581522949543,
+                        with_fee=102.51039361422065,
+                    ),
```
jalextowle, line 139: Your math and explanations look good. The fact that this is different than the example above in all cases except where `time_remaining = 365 days` indicates that we have an issue with time in Hyperdrive. We discussed this previously, and I think we need to remedy the situation before we give the sims to maker.

tests/pricing_models/test_calc_out_given_in.py
```diff
@@ -0,0 +1,1611 @@
...
+                ),
+                fee_percent=0.01,
+                time_remaining=StretchedTime(days=91.25, time_stretch=1),
+                exception_type=AssertionError,
+            ),
+            TestCaseCalcOutGivenInFailure(
+                in_=Quantity(amount=100, unit=TokenType.PT),
+                market_state=MarketState(
+                    # reserves waaaay unbalanced
+                    share_reserves=30_000_000_000,
```
jalextowle, line 1583: I'm a little surprised this isn't a decimal error. Good to know.

jalextowle (approved):
LGTM

tests/pricing_models/test_dataclasses.py
sentilesdal: will follow up with this

PR #136, opened 2023-01-26
improve init_lp, expand initialization testing

wakamex:
replace `test_init_only` with `test_compare_agent_to_calc_liquidity` that subsumes the test (calling the same helper function) while checking against direct initialization through `calc_liquidity` from the pricing model.
sweep across $100-$1 billion target liquidity and 1% to 200% rates
change `get_init_lp_agent` in `sim_utils.py` to use iterative solution that is more accurate
clean up repeated calculations in pricing models


src/elfpy/utils/sim_utils.py
```diff
@@ -112,34 +112,38 @@ def get_init_lp_agent(
         Agent class that will perform the lp initialization action
     """
     # Wrapper functions are expected to have a lot of arguments
-    # pylint: disable=too-many-arguments
-    # get the reserve amounts for a small target liquidity to achieve a target pool APR
+    # pylint: disable=too-many-locals
```
sentilesdal, line 115: don't think we should disable this.  refactor instead.

sentilesdal (commented):
glad we got this tested!  I think we need a few comments on that this code is doing, it feels dense and I can't just skim this and understand what's going on.

src/elfpy/pricing_models/base.py
```diff
@@ -238,7 +239,7 @@ def calc_liquidity(
         # update variables by rescaling the original estimates
         bond_reserves = bond_reserves * scaling_factor
         share_reserves = share_reserves * scaling_factor
-        return (share_reserves, bond_reserves)
+        return share_reserves, bond_reserves
```
dpaiton, line 242: Why remove the parenthesis? I believe python automatically groups multiple returns into a tuple, so the two statements are equivalent, although including the parenthesis is more explicit.

src/elfpy/utils/sim_utils.py
```diff
@@ -112,34 +112,38 @@ def get_init_lp_agent(
...
-    short_amount = init_bond_reserves
-    second_base_to_lp = target_liquidity - init_share_reserves
-    # budget is the full amount for LP & short
-    budget = first_base_to_lp + short_amount + second_base_to_lp
-    # construct the init_lp agent with desired budget, lp, and short amounts
-    init_lp_agent = import_module("elfpy.policies.init_lp").Policy(
+    delta_shares = init_liquidity
+    delta_base = 0
+    trade_result = None
+    for _ in range(11):
```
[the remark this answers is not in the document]
dpaiton: I'd rather see the while loop + max_iterations solution than a function argument.

tests/test_trade.py
```diff
@@ -78,6 +87,54 @@ def test_init_only(self):
...
+                )
+                # assign the results of the init_lp agent to explicit variables
+                share_reserves_new = simulator.market.market_state.share_reserves
+                share_reserves_old, bond_reserves_old = simulator.market.pricing_model.calc_liquidity(
+                    market_state=simulator.market.market_state,  # used only for share_price and init_share_price
+                    target_liquidity=target_liquidity,
+                    target_apr=target_pool_apr,
+                    position_duration=simulator.market.position_duration,
+                )
+                market_old = Market(
```
dpaiton, line 109: nit: `market_old` is only meaningful for us because we recall how this used to be done. I think a more descriptive variable name would be helpful here. e.g. `market_no_lp` vs `market_with_lp`

src/elfpy/pricing_models/base.py
wakamex: i hate unnecessary stuff, so it triggerred me

wakamex: "tuples are not formed by the parentheses, but rather by use of the comma" https://docs.python.org/3/reference/expressions.html

wakamex: [ThiefMaster](https://stackoverflow.com/users/298479/thiefmaster) on [stackoverflow](https://stackoverflow.com/questions/4978567/should-a-return-statement-have-parentheses) says

> return value is the "correct" way - return is a language construct, not a function.
> If you want to return a tuple, use `return your, values, here`
> There's no need for any parenthesis (tuples are created by the , "operator", not the ())

wakamex: regular expressions!
```
grep -rs "^ **return[^(]*," .                                                                        /code/elfpy
./tests/utils_for_tests.py:    return policy_name, kwargs
./src/elfpy/utils/outputs.py:    return fig, axes, grid_spec
./src/elfpy/pricing_models/base.py:        return share_reserves, bond_reserves
./src/elfpy/pricing_models/yieldspace.py:        return lp_out, d_base, d_bonds
./src/elfpy/pricing_models/yieldspace.py:        return lp_in, d_base, d_bonds
./src/elfpy/pricing_models/yieldspace.py:        return lp_in, d_base, d_bonds
./src/elfpy/markets.py:        return market_deltas, wallet_deltas
./src/elfpy/markets.py:        return market_deltas, agent_deltas
./src/elfpy/markets.py:        return market_deltas, agent_deltas
./src/elfpy/markets.py:        return market_deltas, agent_deltas
./src/elfpy/markets.py:        return market_deltas, agent_deltas
./src/elfpy/markets.py:        return market_deltas, agent_deltas
```
I just manually changed one of the above examples, so I'm kinda cheating. but I like this look :)
also just found an errant wallet_deltas!

wakamex:
@dpaiton's thorough explanation of why the `get_init_lp_agent` iterative approach converges in sim_utils.py:

> The following estimates initial market reserves (composed of `share_reserves` and `bond_reserves`=0)  such that the reserve levels after a subsequent short will equal `init_share_reserves`.
>
> The first iteration estimates `delta_shares`, which is proportional to the slippage for the market with share reserves set to an initial estimate. Our first estimate is `share_reserves=init_share_reserves`, which we know is wrong but close.
>
> In future iterations, we set `share_reserves = initial_share_reserves + delta_shares`. We again estimate the slippage and get a new value for delta_shares.
>
> The estimation error on `share_reserves` is our convergence criteria. It is the difference between the previous `delta_shares` and the new `delta_shares`; call it `delta_delta_shares`.
>
> This difference (`delta_delta_shares`) monotonically decreases because each iteration is adjusting `delta_shares` according to the slippage on the remainder. Since slippage will always be less than the reserve levels, the slippage due to `delta_delta_shares` will be less than `delta_delta_shares`, and thus the new difference is lower.

wakamex:
as a code comment, I'm planning on including:
```
# to hit the desired reserve ratio, we open a short which increases bond reserves exactly by the amount we want
# this short trade decreases share reserves by an exact amount that we know (delta_shares)
# since this is undesired, we add delta_shares to the initial liquidity amount (first_base_lp)
# however, this new initial LP amount changes the result of the short trade due to a change in slippage
# this change in slippage is relatively small, and added to the estimate at each iteration, converging quickly
# for detailed convergence process: https://github.com/element-fi/elf-simulations/pull/136#issuecomment-1405922764
```

src/elfpy/utils/sim_utils.py
```diff
@@ -112,34 +112,38 @@ def get_init_lp_agent(
...
+            time_remaining=market.position_duration,
+        )
+        delta_base = trade_result.user_result.d_base
+        delta_shares = delta_base / market.market_state.share_price
+    max_loss = init_bond_reserves - trade_result.user_result.d_base if trade_result is not None else 0
+    init_base_reserves = init_share_reserves * market.market_state.share_price
+    first_base_to_lp = init_base_reserves + delta_base  # add back delta_base to immunize effect of short
+    second_base_to_lp = target_liquidity - init_base_reserves  # remember delta_base is immunized
+    budget = first_base_to_lp + max_loss + second_base_to_lp  # full amount needed to trade
+    init_lp_agent = import_module("elfpy.policies.init_lp").Policy(  # construct the agent with desired amounts
```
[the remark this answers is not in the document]
wakamex: are you saying use `from elfpy.policies.init_lp import Policy as InitializeLpPolicy` at the top?

src/elfpy/utils/sim_utils.py
wakamex: brb creating local_variables dataclass

I guess the idea is this is a hint that the function is doing too much, right? How could we split it up? I was thinking of putting the iteration into a function, but I don't think that actually eliminates any local variables.

I know. make them all global 🙌

src/elfpy/utils/sim_utils.py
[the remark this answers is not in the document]
wakamex: went with dylan's while + iteration count

src/elfpy/markets.py
```diff
@@ -233,13 +233,13 @@ def open_short(
         )
         # amount to cover the worst case scenario where p=1. this amount is 1-p. see logic above.
         max_loss = trade_amount - trade_result.user_result.d_base
-        wallet_deltas = Wallet(
+        agent_deltas = Wallet(
```
dpaiton, line 236: If you want to make this change (fine with me) then you should do it everywhere. Here is what I found with grep:
`./src/elfpy/markets.py:        wallet_deltas = Wallet(
./src/elfpy/markets.py:        return market_deltas, wallet_deltas
./src/elfpy/simulators.py:                wallet_deltas = self.market.trade_and_update(trade)
./src/elfpy/simulators.py:                agent.update_wallet(wallet_deltas, self.market)
./src/elfpy/simulators.py:                    wallet_deltas.__dict__,
./src/elfpy/agent.py:    update_wallet(self, wallet_deltas: Wallet, market: Market) -> None
./src/elfpy/agent.py:    def update_wallet(self, wallet_deltas: Wallet, market: Market) -> None:
./src/elfpy/agent.py:        wallet_deltas : Wallet
./src/elfpy/agent.py:        for key, value_or_dict in wallet_deltas.__dict__.items():`

src/elfpy/pricing_models/base.py
dpaiton: Ok, I'm convinced. Just so I understand -- that grep is showing that we won't be using parenthesis to return tuples anywhere in the repo, correct?

src/elfpy/pricing_models/hyperdrive.py
```diff
@@ -92,24 +92,24 @@ def calc_in_given_out(
...
-        # Redeem the matured bonds 1:1 and simulate these updates hitting the
-        # reserves.
+        # TODO: This is somewhat strange since these updates never actually hit the reserves.
+        # Redeem the matured bonds 1:1 and simulate these updates hitting the reserves.
         if out.unit == TokenType.BASE:
-            market_state.share_reserves -= float(
-                Decimal(out.amount) * (1 - Decimal(time_remaining.normalized_time)) / Decimal(market_state.share_price)
-            )
-            market_state.bond_reserves += float(Decimal(out.amount) * (1 - Decimal(time_remaining.normalized_time)))
+            market_state.share_reserves -= float(d_shares)
```
dpaiton, line 108: much cleaner, thanks!

src/elfpy/pricing_models/base.py
wakamex: that was showing the existing cases where we were already using no-parenthesis return, which is higher than the alternative. i think there were 4 cases where we were using parenthesis, and 2 of them were like `(0,0)`. I've changed those all over now.

src/elfpy/markets.py
wakamex: regular expressions to the rescue! changed the first 5. I'm gonna keep the last 4, cus those are inside the agent, and the agent is managing deltas on its wallet, not on itself. in the other classes, there are `market_deltas` and `agent_deltas` which both refer to wallets, so we need that extra qualifier.

src/elfpy/utils/sim_utils.py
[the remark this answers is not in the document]
sentilesdal: oh yeah that'll work

src/elfpy/utils/sim_utils.py
```diff
@@ -112,34 +112,65 @@ def get_init_lp_agent(
...
+    logging.debug(
+        ("\nstopping at iteration %g because delta_delta_shares is <= %g"),
+        iteration,
+        absolute_tolerance,
+    )
+    delta_base = delta_shares * market.market_state.share_price
+    max_loss_in_base = init_bond_reserves - delta_base
+    init_base_reserves = init_share_reserves * market.market_state.share_price
+    first_base_to_lp = init_base_reserves + delta_base  # add back delta_base to immunize effect of short
+    second_base_to_lp = target_liquidity - init_base_reserves  # remember delta_base is immunized, doesn't need adding
```
sentilesdal, line 171: shouldn't this be 
```second_base_to_lp = target_liquidity - first_base_to_lp - max_loss_in_base```

Comment (sentilesdal) [reply src/elfpy/utils/sim_utils.py:171]: the first action by the agent should send `first_base_to_lp` to the market and the second action should send `max_loss_in_base` to the market no?

Comment (sentilesdal) [reply src/elfpy/utils/sim_utils.py:171]: if the liquidity adds up correctly then something is wrong in our market accounting.  Let's add a TODO if that's the case and follow up with another PR.

File: src/elfpy/utils/sim_utils.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (dpaiton) [reply src/elfpy/utils/sim_utils.py:173]: This pattern lets us specify policy types as strings, which allows end-users to easily create an arbitrary number of policies and specify which ones to use as a list of strings in the config.

I'm also not sure what your proposed alternative is @sentilesdal. Happy to discuss further in voices. If you have a better method I'm eager to hear it.

Either way -- I suggest @wakamex leaves this alone for this PR and we create a new issue with the proposed alternative.

File: src/elfpy/utils/sim_utils.py (hunk 8 above)
Comment (wakamex) [reply src/elfpy/utils/sim_utils.py:171]: the second sends `max_loss_in_base` into the protocol, but keeps it in the margin account outside of the pool reserves

File: tests/test_trade.py [hunk 9]
```diff
@@ -78,6 +87,54 @@ def test_init_only(self):
         """Tests base LP setups"""
         self.run_base_trade_test(agent_policies=[], target_liquidity=1e6, target_pool_apr=0.05)
 
+    def test_compare_agent_to_calc_liquidity(self):
+        """Compare agent init as above to old calc_liquidity method"""
+        for target_liquidity in (1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9):
+            for target_pool_apr in (0.01, 0.03, 0.05, 0.10, 0.25, 0.5, 1, 2):  # breaks at 500%
+                print(f"running test with {target_liquidity=} and {target_pool_apr=}")
```
dpaiton, line 94: Can you change this to a log statement?

src/elfpy/markets.py
dpaiton: following up after reviewing the rest of the PR -- looks like it's good! Resolve if you agree.

src/elfpy/utils/sim_utils.py
wakamex: don't need to use `first_base_to_lp` because the end result of the first LP's accounts is `init_base_reserves` in the pool, not `first_base_to_lp` which is made up of `init_base_reserves + delta_base`. the first LP does put both components into the market, but the second piece (`delta_base`) is there only to offset the effect of the short, so after the short, the market holds `init_base_reserves` base.

src/elfpy/utils/sim_utils.py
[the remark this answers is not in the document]
wakamex: ya I guess that change makes it cleaner, but locks init_lp as the only initialization agent. leaving it like this makes it more wide open to having different initialization agents in the future? I guess depends on whether we envision using different agents in the future, or just one to initialize the market. I'll leave it like this for now.

wakamex (commented):
I approve

tests/test_trade.py
```diff
@@ -78,6 +97,58 @@ def test_init_only(self):
...
+    def test_compare_agent_to_calc_liquidity(self):
+        """Compare agent init as above to old calc_liquidity method"""
+        for target_liquidity in (1e2, 1e3, 1e4, 1e5, 1e6, 1e7, 1e8, 1e9):
+            for target_pool_apr in (0.01, 0.03, 0.05, 0.10, 0.25, 0.5, 1, 2):  # breaks at 500%
+                logging.debug(
+                    "running test with target_liquidity=%g and target_pool_apr=%g",
+                    target_liquidity,
+                    target_pool_apr,
+                )
+                simulator = self.run_base_trade_test(
```
dpaiton, line 109: Took me a bit to realize what is going on here -- can you add a comment that explains how you're using this? e.g. `# run_base_trade_test initializes a market with an LP agent`

tests/test_trade.py
```diff
@@ -78,6 +97,58 @@ def test_init_only(self):
...
+                    agent_policies=[],
+                    target_liquidity=target_liquidity,
+                    target_pool_apr=target_pool_apr,
+                    init_only=True,
+                )
+                # assign the results of the init_lp agent to explicit variables
+                share_reserves_new = simulator.market.market_state.share_reserves
```
dpaiton, line 116: "new" and "old" is only meaningful to someone who has been working on this repo from the start. Can you change these to more descriptive modifiers? Like "agent" and "direct" as you use elsewhere.

src/elfpy/utils/sim_utils.py
wakamex: Dylan resolved this by combining calculations to avoid creating variables

src/elfpy/utils/sim_utils.py
wakamex: let me know if you think there's something missing to this logic, we can add the TODO and/or investigate further. gonna resolve and merge for now.

tests/test_trade.py
```diff
@@ -78,6 +97,58 @@ def test_init_only(self):
...
+                        vault_apr=simulator.market.market_state.vault_apr,
+                        share_price=simulator.market.market_state.share_price,
+                        init_share_price=simulator.market.market_state.init_share_price,
+                    ),
+                    position_duration=simulator.market.position_duration,
+                )
+                total_liquidity_old = market_direct.pricing_model.calc_total_liquidity_from_reserves_and_price(
+                    market_state=market_direct.market_state, share_price=market_direct.market_state.share_price
+                )
+                total_liquidity_new = share_reserves_new * simulator.market.market_state.share_price
```
dpaiton, line 140: I think it makes more sense to use `pricing_model.calc_total_liquidity_from_reserves_and_price()` here, which will hopefully increase the likelihood that this test still passes as we add pricing models.

dpaiton (approved):
Few last nits to fix up and then it's ready to go.

src/elfpy/utils/sim_utils.py
sentilesdal: yeah but `max_loss_in_base` still increases the total liquidity of the protocol.  the smart contract will have that much more base flow into it and the agent will spend that much base, so we should account for it if we aren't.  Add a TODO to investigate.

sentilesdal (approved):

src/elfpy/utils/sim_utils.py
wakamex: added

PR #139, opened 2023-01-31
Add the generated html to the docs folder

sentilesdal:
This is temporary until we improve our ability to do this in the cloud.  Currently vercel only has python 3.9 right now so this is kindve a hack until we can either support building for 3.9 or they add 3.10


dpaiton (approved):
didn't read it.

PR #140, opened 2023-01-31
fix yieldspace tests

sentilesdal:
the yieldspace calc out given in tests were not running because the file name didn't start with 'test'.  This means that changes in #136 that broke tests weren't caught.  They showed as broken in #135 when the filename was changed.  This PR fixes the equations so the tests pass.


dpaiton (approved):
lgtm

PR #147, opened 2023-02-02
Improve docs

Padraic-O-Mhuiris:


docs/source/installation.rst
```diff
@@ -0,0 +1,8 @@
+Quick Install
+=================================
+
+Blah
```
dpaiton, line 4: Is this rendered in the docs? Would rather not have the blah blah blah show up anywhere :-P

dpaiton: You can just point them to the readme for now

Padraic-O-Mhuiris: It was just temp - I'll clean it up

dpaiton (approved):
Thanks, this is a huge improvement!

PR #148, opened 2023-02-02
Random agents & poisson vault apr

dpaiton:
Adds random agents & Poisson distributed vault apr shifts. Also relaxes some asserts (see comment below) and improves log formatting.


dpaiton:
Demo can be seen by running `examples/notebooks/random_agents.ipynb`

### Summary
This PR demonstrates the ability to model random agent behavior. It also adds functionality to simulate a vault APR that steps up or down in a Poisson distributed fashion.

### Poisson-distributed vault APR
To control the vault APR, I created a custom poisson generator. I'll move this into `src/elfpy` in a follow-up PR. The poisson generator takes in parameters that can vary the number of "events" per year. An example event would be the vault apr changing. HERE is a plot demonstrating that it works:

image

The top plot shows the events. The event array is binary, where 1 represents an event and 0 represents the absence of an event. In the plot, each spike is an event. We set the average spikes to be 4 events per year (i.e. avg 3 months between events).

The bottom plot validates that the generating process is poisson. We simulate 10,000 years and generate events per year. We then histogram the normalized number of events occurring in a given year (black bars). The blue line is a [poisson fit](https://en.wikipedia.org/wiki/Poisson_distribution#:~:text=Definitions-,Probability%20mass%20function,-A%20discrete%20random) for the same mean (lambda). Pretty good fit!

This is then piped into a generator that actually generates the vault_apr. That generator has the following arguments:
```
vault_apr_init = 0.01 # Initial vault APR
vault_apr_jump_size = 0.06 # Scale of the vault APR change (vault_apr (+/-)= jump_size)
vault_jumps_per_year = 4 # The average number of jumps per year
vault_apr_jump_direction = "random" # The direction of a rate change. Can be 'up', 'down', or 'random'.
vault_apr_allow_negative = False # Whether or not the vault apr can go negative
```
These arguments match Jonny's specification, typed up in the [clickup](https://app.clickup.com/t/8669e6zmm). Following this spec, here is an example output:
image

Not too interesting! Note that the `vault_apr_jump_direction` is set to `"random"`, which means a coin flip happens for the direction. It seems that the coin landed on "down" multiple times, but the apr couldn't go any lower. Let's allow negative rates and rerun:
image

Ok, more interesting. We can also set `vault_apr_jump_direction` to `"up"` to keep it positive:
image

Wow. Looks great. You can modify the behavior in the notebook if you want more control.

### Random agents

Agents randomly open longs & shorts for a random amount of money. They also randomly close the positions. When they close, they close the full position, but it is simple to modify the trade_amount to be partial closure.

It's a bit of a mess even with 5 agents, but here is the long/short behavior:
image

jrhea (changes requested):
suggested changes:

- let's go ahead and set `avg_events_per_interval = vault_jumps_per_year`
- when `vault_apr_jump_direction = "random"` and `vault_apr_allow_negative = False` the Vault APR doesn't seem to change much bc it bottoms out at zero:
image
we could either resample, or make a weighted coinflip that is less likely to do down the closer it is to zero and less like to go up the closer it is to 6. (I think i like the latter option)
- change the title of this plot to Random Agents
image
Also why does it only go up?  I configured the vault APR to be random and i would think it would look like the previous plot above
- Market Pool APR Plot
image
The problem with it starting at 0 and immediately jumping to like 2.2% ends up hiding the wiggles in the APR bc it forces the plot to be zoomed out.  Perhaps if we add a param to `output_utils.plot_pool_apr(simulator)` that trims the initial data point?
- Market Spot Price Plot (same issue as Market Pool APR Plot)
image

dpaiton:
- [x] let's go ahead and set avg_events_per_interval = vault_jumps_per_year
- [x] when vault_apr_jump_direction = "random" and vault_apr_allow_negative = False the Vault APR doesn't seem to change much bc it bottoms out at zero

------
 I fixed this by modifying the function to take lower & upper bound arguments. You can set these to extreme values like +/- 1e9 to have it act like an "unbounded" apr. There is also a new direction, so the options are ["up", "down", "random", or "random_weighted"]. The random weighted option uses a linear PDF that maps any given apr to a probability `p` for the "up" direction and `1-p` for the "down" direction. These are all displayed in this plot:
 
plots
------

- [ x] change the title of this plot to Random Agents
- [ ] Also why does it only go up? I configured the vault APR to be random and i would think it would look like the previous plot above
- [ x] Market Pool APR Plot
- [ x] Market Spot Price Plot (same issue as Market Pool APR Plot)
- [ x] Market Spot Price Plot (same issue as Market Pool APR Plot)

jrhea (approved):
lgtm

PR #163, opened 2023-02-05
remove html (build in ci)

sentilesdal:
We can now remove all the html files in the docs/ directory and build them directly in Vercel instead.  In the Vercel project settings for the repo we now run `./vercel-install.sh` for the install command, `./vercel-build.sh` for the build command and change the output dir to `docs/html`


vercel-install.sh
```diff
@@ -0,0 +1,10 @@
+python3 --version
+which python3
+python3 -m ensurepip
+python3 -m pip install --upgrade pip
+# repository dependencies
+python3 -m pip install -r requirements.txt
```
Padraic-O-Mhuiris, line 6: I don't think this is needed

vercel-install.sh
```diff
@@ -0,0 +1,10 @@
...
+# extra dependencies for the docs
+python3 -m pip install -r requirements-dev.txt
+# intstall elfpy package
+python3 -m pip install -e .
```
Padraic-O-Mhuiris, line 10: I don't think this is needed - just for docs

vercel-build.sh
```diff
@@ -0,0 +1,4 @@
+rm -rf build html docs/source/autoapi
+cd docs
+make html
+cp -r build/html/ ../static/
```
Padraic-O-Mhuiris, line 4: You need to cleanup all generated files before `make` is ran, and create `static`. Also add `/static` and `/docs/build/ to gitignore or delete build after it's ran

Padraic-O-Mhuiris (changes requested):

Padraic-O-Mhuiris (approved):

PR #164, opened 2023-02-05
change lp token base value calc

sentilesdal:
Changes the calculation of the approximate base value of the lp tokens

### DIFF
--- a/src/elfpy/wallet.py
+++ b/src/elfpy/wallet.py
@@ -113,7 +113,13 @@ class Wallet:
 
         .. todo:: TODO: return a dataclass instead of dict to avoid having to check keys & the get_state_keys func
         """
-        lp_token_value = market.remove_liquidity(self.address, self.lp_tokens)[1].base if self.lp_tokens > 0 else 0.0
+        lp_token_value = (
+            market.market_state.share_reserves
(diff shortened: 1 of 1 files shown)


src/elfpy/wallet.py
```diff
@@ -113,7 +113,13 @@ def get_state(self, market: Market) -> dict:
...
         .. todo:: TODO: return a dataclass instead of dict to avoid having to check keys & the get_state_keys func
         """
-        lp_token_value = market.remove_liquidity(self.address, self.lp_tokens)[1].base if self.lp_tokens > 0 else 0.0
+        lp_token_value = (
+            market.market_state.share_reserves
+            * market.market_state.share_price
+            * (self.lp_tokens / market.market_state.lp_reserves)
+            if self.lp_tokens > 0
+            else 0.0
+        )
```
jalextowle, line 122: Like Jonny was saying on the call, we could have an overestimate and an underestimate. I'll approve now though since we can play that by ear.

jalextowle (approved):

src/elfpy/wallet.py
sentilesdal: yeah sounds good

PR #172, opened 2023-02-06
allow PNL notebook to run from main branch on colab

wakamex:
allow PNL notebook to run from main branch on colab


jrhea (approved):
works for me now

PR #190, opened 2023-02-08
improve lp pnl calculation

wakamex:
* include bond reserves in LP PnL
* add precision check when market reserves are updated in the apply_delta() function
* remove precision checks from 
* remove tests for market reserves from test_calc_in_given_out: negative trade amount, zero trade amount, share reserves < 1 wei, bond reserves < 1 wei
* remove tests for market reserves from test_calc_out_given_in: negative share reserves, negative base reserves, share reserves < 1 wei, bond reserves < 1 wei
* add names to test_calc_in_given_out and test_calc_out_given_in failure test cases so you don't have to manually count code blocks to find test case 14 (copied from existing code comments)
* add test_max_short.py which tests that attempting to 500 and 5000 PTs against liquidity of $200 is scaled down properly, with and without init_lp


src/elfpy/pricing_models/base.py
```diff
@@ -481,8 +481,10 @@ def get_max_short(
         last_maybe_max_short = 0, 0
         bond_percent = 1
         num_iters = 25
+        iteration = 0
         market_state_post_trade = copy.copy(market_state)
         for step_size in [1 / (2 ** (x + 1)) for x in range(num_iters)]:
+            iteration += 1
```
sentilesdal, line 487: what does this do?  remove it?

src/elfpy/pricing_models/base.py
```diff
@@ -568,14 +571,6 @@ def check_input_assertions(
             "pricing_models.check_input_assertions: ERROR: "
             f"expected quantity.amount >= {WEI}, not {quantity.amount}!"
         )
-        assert market_state.share_reserves >= WEI, (
```
sentilesdal, line 571: these seem like good checks to have no?

src/elfpy/simulators.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         self.config.check_vault_apr()
         # NOTE: lint error false positives: This message may report object members that are created dynamically,
         # but exist at the time they are accessed.
-        self.config.freeze()  # pylint: disable=no-member # type: ignore
+        self.config.freeze()  # type: ignore
```
sentilesdal, line 43: we no longer need the pylint disabled?  if so that's cool

sentilesdal (commented):
oops just noticed this is still a draft, stopping my review now

src/elfpy/pricing_models/base.py
wakamex: yeah, but they don't work in a world where we end up with negative reserves, then catch them in `apply_deltas` and set reserves to 0 when the negative value is within a precision threshold. I guess we could set that amount to 1 WEI instead of 0? that way it would be closer to solidity, which is probably going to keep this check?

src/elfpy/simulators.py
wakamex: it seems to be the case, but only in a few cases, which I can't quite figure out why lol. like in the main package, it seems to be not needed. but in the test classes, it is.

examples/notebooks/midwit.ipynb
```diff
@@ -0,0 +1,756 @@
+{
```
dpaiton, line 1: pls give this file a meaningful name

src/elfpy/pricing_models/base.py
```diff
@@ -327,6 +328,9 @@ def _calc_spot_price_from_reserves_high_precision(
             market_state.share_reserves
         )
         # p = ((y + s)/(mu*z))^(-tau) = ((2y + cz)/(mu*z))^(-tau)
+        assert market_state.init_share_price != 0, "init_share_price cannot be zero"
```
dpaiton, line 331: I think it makes more sense to put these in the external function call `calc_spot_price_from_reserves`

src/elfpy/pricing_models/base.py
```diff
@@ -525,6 +529,7 @@ def get_max_short(
                     or market_state_post_trade.share_price * market_state_post_trade.share_reserves
                     < market_state_post_trade.base_buffer
                     or market_state_post_trade.bond_reserves < market_state_post_trade.bond_buffer
+                    or market_state_post_trade.share_reserves < 0
```
dpaiton, line 532: Why are you adding this? Adding a test given the TODO above (line 520) makes me nervous. Do we need all of these conditionals? If you have it in your head, can you add a comment on what these conditionals are achieving? What's the logic summary for this block?

src/elfpy/pricing_models/base.py
dpaiton: We really should have these rounding conditionals & checks located in one place. It's dangerous to sprinkle them throughout the code -- you can get values transforming that you don't expect. Debugging becomes increasingly difficult. I _think_ that is what Mihai is suggesting, but I just want to make it clear from my end.

dpaiton: But on the other hand we need to have full coverage with these model assertions.

src/elfpy/pricing_models/hyperdrive.py
```diff
@@ -112,8 +112,17 @@ def calc_in_given_out(
         d_bonds = out_amount * (1 - normalized_time)
         d_shares = d_bonds / share_price
 
-        # TODO: Verify that this is needed.
-        market_state = copy.copy(market_state)
+        # this is needed to not modify the actual market state, as this is a mock only
+        market_state = copy.copy(
```
dpaiton, line 116: I don't think you need `copy.copy` if you're instantiating a new object.

src/elfpy/simulators.py
```diff
@@ -186,9 +186,16 @@ def execute_trades(self, trades: list[tuple[int, list[MarketAction]]]) -> None:
         trades : list[tuple[int, list[MarketAction]]]
             A list of agent trades. These will be executed in order.
         """
+        number_of_trades = sum((len(agent_trades) for _, agent_trades in trades))
+        if number_of_trades == 0:
+            return
+        logging.info(
```
dpaiton, line 192: I think this should be `logging.debug`. I know it's not consistent now, but I'd like the `info` logs to be daily or _maybe_ per block sumaries. Definitely not per trade.

src/elfpy/simulators.py
```diff
@@ -186,9 +186,16 @@ def execute_trades(self, trades: list[tuple[int, list[MarketAction]]]) -> None:
...
+            "Executing trades at self.day=%g and self.daily_block_number=%g", self.day, self.daily_block_number
+        )
         for agent_id, agent_trades in trades:
             agent = self.agents[agent_id]
             for trade in agent_trades:
+                print(f"  trade #{self.run_trade_number} for agent #{agent_id:03.0f}: {trade=}")
```
dpaiton, line 198: accidental print statement

src/elfpy/utils/price.py
```diff
@@ -22,8 +23,7 @@ def calc_apr_from_spot_price(price: float, time_remaining: StretchedTime):
         APR (decimal) calculated from the provided parameters
     """
     assert price > 0, (
-        "utils.price.calc_apr_from_spot_price: ERROR: "
-        f"Price argument should be greater or equal to zero, not {price}"
+        "utils.price.calc_apr_from_spot_price: ERROR: " f"Price argument should be greater than zero, not {price}"
```
dpaiton, line 26: Some weirdness here with the strings

src/elfpy/wallet.py
```diff
@@ -113,13 +113,10 @@ def get_state(self, market: Market) -> dict:
...
         .. todo:: TODO: return a dataclass instead of dict to avoid having to check keys & the get_state_keys func
         """
-        lp_token_value = (
-            market.market_state.share_reserves
-            * market.market_state.share_price
-            * (self.lp_tokens / market.market_state.lp_reserves)
-            if self.lp_tokens > 0
-            else 0.0
-        )
+        share_of_pool = self.lp_tokens / market.market_state.lp_reserves if market.market_state.lp_reserves > 0 else 0.0
```
dpaiton, line 116: I think it would be more clear to put this whole assignment in a loop, since it results in `lp_token_value` = 0. i.e.:
```
if market.market_state.lp_reserves > 0:
    # compute lp_token_value
else:
    lp_token_value = 0
```

dpaiton:
Can you explain what the changes are in `calc_*_given_*`? What did you do, why did you have to do it, have we lost any test coverage?

src/elfpy/pricing_models/base.py
wakamex: this was when I thought there was an error with get_max_short (there isn't). I noticed it's not doing checks for share_reserves, so I added it 🤷

wakamex: since we don't know which of those checks are necessary anyway, adding another one doesn't change that. I'm not attempting to do a thorough check of whether these are necessary here. If leaving this additional check out makes you more comfortable, I can do that.

src/elfpy/pricing_models/hyperdrive.py
wakamex: true, I was doing it without, then half-reverted. i'll revert fully.

src/elfpy/simulators.py
wakamex: this is not per trade. it's right before the for loop that loops across trades. so leave it as info? i'm okay either way, but I thought it's a nice thing to have to delineate the passage of time in logs. it also only gets printed when there's a trade.

wakamex:
merged separately in #203 and #228

src/elfpy/pricing_models/base.py
wakamex: that function has been refactored

src/elfpy/simulators.py
wakamex: left this out, since `log_market_step_string` is sufficient

PR #193, opened 2023-02-09
check for available shares instead of bonds for get_max_short

wakamex:
need this to get a non-zero max short when bond reserves are zero, but there's plenty of share reserves to facilitate a short (the bootstrap from 0% scenario, without an init_lp)

### DIFF
--- a/src/elfpy/pricing_models/base.py
+++ b/src/elfpy/pricing_models/base.py
@@ -469,8 +469,8 @@ class PricingModel(ABC):
         float
             The maximum amount of bonds that can be shorted.
         """
-        available_bonds = market_state.bond_reserves - market_state.bond_buffer
-        if available_bonds <= 0:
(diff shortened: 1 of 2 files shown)


dpaiton (approved):
good catch, thanks!

src/elfpy/pricing_models/base.py
```diff
@@ -469,8 +469,8 @@ def get_max_short(
         float
             The maximum amount of bonds that can be shorted.
         """
-        available_bonds = market_state.bond_reserves - market_state.bond_buffer
-        if available_bonds <= 0:
+        available_shares = market_state.share_reserves - market_state.base_buffer / market_state.share_price
```
sentilesdal, line 472: yeah nice catch!

sentilesdal (approved):

PR #196, opened 2023-02-09
add extra step for get_max_short

sentilesdal:


src/elfpy/pricing_models/base.py
```diff
@@ -525,6 +526,22 @@ def get_max_short(
                         return last_maybe_max_short
                     bond_percent += step_size
 
+        # do one more iteration at the last step size if the bisection method was stuck approaching
+        # a max_short value that used more reserve shares than the market had available.
+        if market_state_post_trade.share_reserves < market_state.base_buffer:
+            last_step_size = 1 / (2 ** (num_iters + 1))
+            bond_percent -= last_step_size
```
wakamex, line 533: how do we know that subtraction is the correct next step? couldn't it vary, depending on where you end up, or which direction you're approaching from? I guess in this case we're only approaching from the "too big" direction so a decrease is always... getting closer? but isn't that dangerous? shouldn't we save the `last_direction_used` and reverse that?

wakamex: oh I guess if the statement by definition is "you've shorted too much" so the answer is always in the negative direction... "short 1 step less"

wakamex:
would we want to start the agent's estimate at something like `min(share_reserves,wallet.shares)`, instead of the market max short? in the case where his wallet is much smaller, we're currently wasting like the first 10 iterations by starting at a way-too-high value. but it's not like this algo ever finishes early, so we'll always use all 25 iterations.  this seems like a better starting estimate tho.

src/elfpy/pricing_models/base.py
sentilesdal: Yeah exactly.

wakamex (approved):
lgtm, leaving improved starting point for later

PR #197, opened 2023-02-09
Fixed the APR accounting

jalextowle:
I made changes to the APR accounting when fixing the flat+curve model by adding a normalizing constant. It looks like those changes were dropped; however, the changes are required for the APR accounting to work correctly (otherwise we aren't computing APR since it wouldn't be annualized). Running `pytest -s -k test_example` shows that the low-level accounting functions are working properly. I inspected a breaking test in `test_trade` and found that the initial bootstrapping of the market using the initial LP agent works gives the expected APR. The problem lies elsewhere in the logic of the test or the initial LP agent.

An easier way to fix this may be to abandon the init LP agent entirely and use an approach that is more similar to that used in the contract (using `Market`'s `initialize function).

### DIFF
--- a/src/elfpy/pricing_models/base.py
+++ b/src/elfpy/pricing_models/base.py
@@ -19,6 +19,7 @@ from elfpy.types import (
     TradeResult,
 )
 import elfpy.utils.price as price_utils
+import elfpy.utils.time as time_utils
 
 # Set the Decimal precision to be higher than the default of 28. This ensures
 # that the pricing models can safely a lowest possible input of 1e-18 with an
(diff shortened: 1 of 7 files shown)


dpaiton:
Adds a test for calculating spot price from utils & the pricing model
adds comments to explain the annualized_time step
general logging, documentation, variable name improvements
fixes bug in calculating LP liquidation to now use annualized time instead of normalized time remaining

src/elfpy/pricing_models/base.py
```diff
@@ -160,10 +163,14 @@ def calc_share_reserves(
...
+        .. todo:: Write a test for this function
         """
-        # TODO: Write a test for this function
+
+        # Only want to renormalize time for APR ("annual", so hard coded to 365)
+        # Don't want to renormalize stretched time
+        annualized_time = time_utils.norm_days(time_remaining.days, 365)
         share_reserves = bond_reserves / (
-            init_share_price * (1 - target_apr * time_remaining.normalized_time) ** (1 / time_remaining.stretched_time)
+            init_share_price * (1 - target_apr * annualized_time) ** (1 / time_remaining.stretched_time)
```
jrhea, line 173: i like annualized time

jrhea (approved):
lgtm

PR #198, opened 2023-02-10
Test notebooks

dpaiton:


wakamex:
would there be any benefit in using [jupytext](https://github.com/mwouts/jupytext) for this task?

sentilesdal (approved):
this looks good.  glad to see the notebooks get cleaned up!

PR #200, opened 2023-02-11
standardize inputs and reserves checks

wakamex:
* add precision check when market reserves are updated in `apply_delta` which is called in `update_market` which is called in `trade_and_update` at the end of every trade
* remove precision checks from pricing models
* remove tests for market reserves from `test_calc_in_given_out`: share reserves < 1 wei, bond reserves < 1 wei
* add failure test for share_reserves==0 in test_calc_in_given_out checked only for yieldspace (it's valid in hyperdrive 😈)
* remove tests for market reserves from `test_calc_out_given_in`: negative share reserves, negative base reserves, share reserves < 1 wei, bond reserves < 1 wei
* add names to `test_calc_in_given_out` and `test_calc_out_given_in` failure test cases so you don't have to manually count code blocks to find test case 14 (copied from existing code comments)
* make checks for time less strict to allow time=1.0 and eliminate `TODO: convert this to a check for 1>=time and fix tests as necessary`
* add basic logging to `test_notebook` so I can see which notebook is failing

### DIFF
--- a/tests/test_notebooks.py
+++ b/tests/test_notebooks.py
@@ -11,10 +11,13 @@ import pathlib
 import ast
 import tempfile
 from contextlib import redirect_stdout, redirect_stderr
+import logging
 
 import astunparse
 from IPython.core.inputtransformer2 import TransformerManager
 import nbformat
+
+import elfpy.utils.outputs as output_utils  # utilities for file outputs
 
 
 class TestNotebook(unittest.TestCase):
@@ -24,6 +27,7 @@ class TestNotebook(unittest.TestCase):
         """Tests notebooks in the `examples/notebooks` folder to ensure that they run without error"""
         # pylint: disable=too-many-locals
         # pylint: disable=too-many-branches
+        output_utils.setup_logging(log_filename=".logging/test_notebooks.log", log_level=logging.DEBUG)
         isp = TransformerManager()  # module for converting jupyter cell into source code
         notebook_location = os.path.join(
             os.path.dirname(pathlib.Path(__file__).parent.resolve()),
@@ -32,6 +36,7 @@ class TestNotebook(unittest.TestCase):
         for file in os.listdir(notebook_location):
             if not file.endswith(".ipynb"):
                 continue
+            logging.debug("Testing notebook: %s", file)
             # Read the notebook cell by cell, grab the code & add it to a string
             notebook = nbformat.read(os.path.join(notebook_location, file), as_version=4)
             file_source = ""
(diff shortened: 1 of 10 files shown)


wakamex:
== in yieldspace.py ==
amount_in > 0			COVERED, more restrictive (>= WEI vs. > 0)
shares >= 0			missing, covered in apply_deltas
bond >= 0			missing, covered in apply_deltas
base buffer >= -1e-8		missing, covered in apply_deltas
** missing bond buffer **	missing, covered in apply_deltas
lps >= 0			missing, covered in apply_deltas
rate >= 0			missing, not needed since redundant w/ price and not in market_state
norm time >= 0			missing, ADDED
stretched time >= 0		COVERED
share_price >= 1		COVERED

== in check_input_assertions ==
amount >= WEI
share_price >= 1
reserves_difference < MAX_RESERVES_DIFFERENCE
trade_fee_percent >= 0
redemption_fee_percent >= 0
stretched_time >= 0

tests/test_notebooks.py
```diff
@@ -32,6 +36,7 @@ def test_notebook_execution(self):
         for file in os.listdir(notebook_location):
             if not file.endswith(".ipynb"):
                 continue
+            logging.debug("Testing notebook: %s", file)
```
dpaiton, line 39: great, thanks for adding this

dpaiton:
> Member

Can you add this as a comment somewhere? I think `check_input_assertions` makes sense.

wakamex:
each of those items is an assert in yieldspace

covered = this DOES exist in `check_input_assertions`
missing = this does NOT exist in `check_input_assertions`

the missing cases we need to do something about

so what are we doing about them?

the first 5 are all covered by checking for reserve_field > -PRECISION_THRESHOLD (bond buffer didn't have an assert in yieldspace, I just noted it here for completeness, and to show that our previous checks were lacking. new ones are more thorough, since they exhaustively check every key in the market_state class. so if we add a new type of reserve or buffer, it will be automatically checked)

normalized time had an assert in yieldspace, but not in `check_input_assertions`, so I added it. I'm not sure if it's 100% redundant with the stretched time check, but they'll both be covered even if we change how we use them in the future.

I'm fine not checking rate, since there's a 1:1 mapping from price to rates. prices from 0 to 1 map to rates from infinity to 0. so we assrt we're in that range, inclusive of endpoints, just by checking price. relationship between price and rate is never going to change.

wakamex:
closed by #203

PR #203, opened 2023-02-14
check inputs in liquidity trades

wakamex:
1. add check inputs in liquidity trades (add liquidity and remove liquidity)
2. this causes a bunch of failures, so fix this by loosening asserts. which ones?
```
checks loosened:
bond > 1 wei => bond >= 0
shares > 1 wei => shares >= 0
allow time == 1 for norm and stretched:
1 > stretched_time >= 0
=> 1 >= stretched_time >= 0
=> 1 >= norm_time >= 0
```
why is this okay?
the first two check that bond and share reserves are >=0, a value lower than previous by 1e-18 (1 WEI). this is well within our precision threshold. it's needed because in the percision threshold check, we take a negative value within the threshold and set it to 0, which would fail the `> 1 wei` check.
allowing time == 1 is needed for trades on day 0 and had been planned to be added. this removes the TODO for it.
3. remove asserts from `yieldspace.py` because they're all fully covered in `check_input_assertions` now
4. make assert in `types.py` more descriptive
5. fix tests in test_calc_in/out_given_out/in
remove 2 tests for `reserves < 1 wei` no longer needed, since we don't check for it, as above
move `share_reserves == 0` into a new category that fails in yieldspace only
this passes in hyperdrive because before executing the trade, hyperdrive adds the matured portion to share reserves, so there's now something to execute against
6. add new `share_reserves == 0` as a success case for hyperdrive only
7. this requires adding model-specific testing
8. this requires changing the `TestResultCalcInGivenOutSuccessByModel` dataclass to make the model-specific test cases optional

### DIFF
--- a/src/elfpy/markets.py
+++ b/src/elfpy/markets.py
@@ -472,6 +472,13 @@ class Market:
             rate = 0
         else:
             rate = self.apr
+        # sanity check inputs
+        self.pricing_model.check_input_assertions(
+            quantity=Quantity(amount=trade_amount, unit=TokenType.PT),  # temporary Quantity object just for this check
+            market_state=self.market_state,
+            time_remaining=self.position_duration,
+        )
+        # perform the trade
         lp_out, d_base_reserves, d_token_reserves = self.pricing_model.calc_lp_out_given_tokens_in(
             d_base=trade_amount,
             rate=rate,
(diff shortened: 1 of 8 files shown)


wakamex:
checks loosened:
bond > 1 wei => bond > 0
shares > 1 wei => shares > 0
init_share_price >= share_price >= 1
 => init_share_price > 0
 => share_price > 0
allow time == 1 for norm and stretched:
1 > stretched_time >= 0
 => 1 >= stretched_time >= 0
 => 1 >= norm_time >= 0

dpaiton (approved):
Great, thanks for getting this in!

PR #207, opened 2023-02-15
propagates changes to remove open_share_price argument from create_ag…

dpaiton:
create_agent_action no longer has an open_share_price argument. If the action is close short then it is derived from the share price.
MarketAction has a better default for open_share_price, and only requires it on close_short
if it is not provided, market.close_short fails instead of propagating a default value


wakamex (approved):
lgtm

PR #209, opened 2023-02-15
Add functions to calculate bonds or base at target apr

sentilesdal:
Add `calc_base_for_target_apr` and `calc_bond_for_target_apr` which can be used to calculate the min/max base/bond in/out depending on the context at a target apr.


villanuevawill:
Should be added to the GET_MAX_LONG function

tests/pricing_models/test_calc_in_given_out.py
```diff
@@ -4,7 +4,7 @@
 import decimal
 import unittest
 import numpy as np
-from test_dataclasses import (
+from dataclass_testcases import (
```
dpaiton, line 7: https://media.giphy.com/media/LyJ6KPlrFdKnK/giphy.gif

tests/pricing_models/test_pricing_model_utils.py
```diff
@@ -316,3 +538,11 @@ def test_calc_liquidity(self):
         """Execute the test"""
         self.run_calc_liquidity_test(sim_utils.get_pricing_model("hyperdrive"))
         self.run_calc_liquidity_test(sim_utils.get_pricing_model("yieldspace"))
+
+    def test_calc_bond_for_target_apr(self):
+        """Execute the test"""
+        self.run_calc_bond_for_target_apr(sim_utils.get_pricing_model("hyperdrive"))
```
dpaiton, line 544: should this also have a yieldspace version?

tests/pricing_models/test_pricing_model_utils.py
```diff
@@ -316,3 +538,11 @@ def test_calc_liquidity(self):
...
+
+    def test_calc_base_for_target_apr(self):
+        """Execute the test"""
+        self.run_calc_base_for_target_apr(sim_utils.get_pricing_model("hyperdrive"))
```
dpaiton, line 548: should this also have a yieldspace version?

tests/pricing_models/dataclass_testcases.py
```diff
@@ -16,8 +16,6 @@ class TestCaseCalcInGivenOutSuccess:
     days_remaining: float
     time_stretch_apy: float
 
-    __test__ = False  # pytest: don't test this class
```
dpaiton, line 19: 👍

src/elfpy/pricing_models/base.py
```diff
@@ -164,14 +164,106 @@ def calc_share_reserves(
...
+        # delta_y / delta_z = p = 1 - r
+        # delta_z = delta_y / (1 - r)
+        # delta_x = c * delta_y / (1 - r)
+
+        # Only want to renormalize time for APR ("annual", so hard coded to 365)
+        # Don't want to renormalize stretched time
+        annualized_time = time_utils.norm_days(position_duration.days, 365)
+        base_in = share_price * bond / (1 - target_apr * annualized_time)
+
+        assert base_in >= 0, "base value negative"
```
dpaiton, line 222: would this happen as a result of inappropriate input values? If so, shouldn't the assert be on the input ranges? If this could happen with reasonable values, isn't it better to put bounds on `base_in` such that it is always valid. Seems weird to put an assert on a computed value.

src/elfpy/pricing_models/base.py
```diff
@@ -164,14 +164,106 @@ def calc_share_reserves(
...
+        # delta_y / delta_z = p = 1 - r
+        # delta_y = delta_z * (1 - r)
+        # delta_y = (delta_x / c) * (1 - r)
+
+        # Only want to renormalize time for APR ("annual", so hard coded to 365)
+        # Don't want to renormalize stretched time
+        annualized_time = time_utils.norm_days(position_duration.days, 365)
+        bond_out = (base / share_price) * (1 - target_apr * annualized_time)
+
+        assert bond_out >= 0, "bond value negative"
```
dpaiton, line 264: see above on base_in

tests/pricing_models/test_calc_in_given_out.py
sentilesdal: renaming... :P

tests/pricing_models/test_pricing_model_utils.py
sentilesdal: its pricing model agnostic, will leave a comment

tests/pricing_models/test_pricing_model_utils.py
sentilesdal: pricing model agnostic, will leave a comment

src/elfpy/pricing_models/base.py
sentilesdal: This was by no means meant to be exhaustive.  Check the other functions in base.py.  If we have exhaustive asserts in our base level calc methods, we are adding a lot of steps to run the code.  This one assert does a minimum level of sanity checking for the output which is nice I think.  I actually could be convinced to take the assert _out_ and throw an error in the test instead, but this made testing a little easier.

src/elfpy/pricing_models/base.py
sentilesdal: see answer above

src/elfpy/pricing_models/base.py
dpaiton: I see, ok I'm good with it.

dpaiton (approved):
good besides fixes marked by CI

PR #210, opened 2023-02-16
Wakamex/smart agents2

wakamex:


wakamex:
moved scratchpad to https://github.com/wakamex/elf-simulations/pull/5

PR #214, opened 2023-02-17
Add new simulation state that uses relational dataframes & stores trade logs

dpaiton:
First pass, hidden under a feature flag

### DIFF
--- a/src/elfpy/types.py
+++ b/src/elfpy/types.py
@@ -4,10 +4,11 @@ from __future__ import annotations  # types will be strings by default in 3.11
 import logging
 from functools import wraps
 from typing import TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


src/elfpy/types.py
```diff
@@ -588,3 +592,140 @@ def check_vault_apr(self) -> None:
                 + f"{self.num_trading_days},"
                 + f" not {len(self.vault_apr)}"
             )
+
+
+@dataclass
+class RunSimVariables:
```
sentilesdal, line 598: Let's land as is.  But types.py is gonna get huge.  Let's move these into a simulator/types.py folder/file

sentilesdal (approved):
yup looks good

PR #224, opened 2023-02-17
improve market.trade_and_update and market.update_market

dpaiton:


src/elfpy/types.py
```diff
@@ -271,10 +271,10 @@ def apply_delta(self, delta: MarketDeltas) -> None:
         self.lp_reserves += delta.d_lp_reserves
         self.share_price += delta.d_share_price
 
+        # TODO: issue #146
         # this is an imperfect solution to rounding errors, but it works for now
         # ideally we'd find a more thorough solution than just catching errors
-        # when they are. issue #146 tracks this.
-        # .. todo::  #146
+        # when they are.
```
sentilesdal, line 277: yes, when they are indeed.

sentilesdal (approved):

Issue #212, opened 2023-02-17
move market initialization steps into src


dpaiton:
market initialization is set up correctly in hyperdrive.ipynb -- need to update helper functions in sim_utils to initialize correctly.  Currently sim_utils market initialization only works with yieldspace. The adjustment of target_apr so that it is scaled by market.position_duraiton.days / 365 is also in sim_utils, and should be in a more core spot. Maybe in config?

PR #226, opened 2023-02-17
updates market initialization

dpaiton:
* fixes sim_utils market initialization to use new scheme that supports hyperdrive & yieldspace
* fixes bug with frozen agent actions; they should not be frozen because we adjust mint_time
* fixes bug in example_main, single short where open_short actions did not include mint_time
* updates example notebooks to use market initialization in sim_utils
* fixes bug with mint_time_default in agent.create_agent_action
* simplifies init_lp agent to only add liquidity, since markets are assumed to have the correct apr
* Rewrites test_trade to more closely match the pattern used elsewhere, and expands the tests


dpaiton:
addresses #212

tests/test_trade.py
```diff
@@ -22,139 +22,136 @@ class SingleTradeTests(unittest.TestCase):
...
+                            ),
+                            target_liquidity,
+                            target_pool_apr,
+                            num_position_days,
+                            pricing_model_name,
+                            simulator.market.market_state,
+                        )
+                        # assign the results of the init_lp agent to explicit variables
+                        # market_state is used only for share_price and init_share_price
+                        share_reserves_direct, bond_reserves_direct = simulator.market.pricing_model.calc_liquidity(
```
sentilesdal, line 61: this is the exact method that is used in market_state, so we aren't really verifying values.  In cases like this it's better to handroll and do a hardcoded value in case the underlying function is changed and messed up at some point.

sentilesdal: or rather, this and init market both call calc_bond_reserves.

sentilesdal (approved):
awesome, glad to get this done.

PR #234, opened 2023-02-20
create narrative (static output containing experiment over simulations, from .py file)

wakamex:


wakamex:
closing in favor of using executed notebooks to share results, as per https://github.com/delvtech/elf-simulations/pull/490

PR #235, opened 2023-02-21
Locate types near their class

dpaiton:
This PR was primarily motivated by needing to restructure `markets.py` to be in its own folder so that it's easy to group alternative markets (e.g. a borrow market). However, that alone could have been done without any of this other stuff. 🤷 

Objects stored in `types.py` were associated with a variety of different core classes in `src`. Additionally, the file structure of the repo did not support grouping multiple markets. This is now fixed by reorganizing the file & folder structure, and moving types out of `types.py` and into the files for classes they are associated with. The primary utility here is to get stuff closer together so it is more intuitive for someone navigating imports & the repo itself. 

old and busted:
```
src/elfpy/
    policies/
        init_lp.py
        ...
    pricing_models/
        ...
    utils/
        ...
    agent.py
    markets.py
    simulators.py
    types.py
    wallet.py
```

new hotness:
```
elfpy/
    agents/
        policies/
            init_lp.py
            ...
        agent.py
        wallet.py
    markets/
        hyperdrive.py
    pricing_models/
        ...
    utils/
        ...
    simulators/
        simulators.py
        trades.py
    types.py
```

additionally, only the following classes remain in types:
```
freezable
to_description
TokenType
Quantity
```

**Notes:**
* Imports suck now. Doing `from module import object` in python is effectively importing everything in `module`, then deleting everything except `object`. This is problematic with the new organization because, e.g., `types.py` imports `MarketAction` from `hyperdrive.py` and `hyperdrive.py` imports `freezable` from `types.py`. The problem is solved by changing up the import syntax, as can be seen in the file diffs. It's less clean because we have to use `module.object` instead of just `object` throughout the file. This is a quick fix -- I _think_ we can implement a better fix by coordinating module namespaces using the `__init__.py` files. But I'd rather put that in a different PR because I'm not 100% sure how it works.
* `hyperdrive.py` is now considerably bigger. However, right now it seems to be ok. I don't see it growing much more. And `types.py` was getting to big, with a clear path to getting increasingly bigger. 
* I could see an argument for moving `pricing_models` into `markets` -- I'll do it if reviewers agree.


sentilesdal (approved):
looks like mostly mechanincal changes, approving.  Definitely a step in the right direction!

PR #239, opened 2023-02-22
Small Refactorings

sentilesdal:
Renames a few things to make them more generic in the base market.  
Moves trades.py to pricing_models since that's where that originates from.  Simulators shouldn't have it since Markets needs to import it and it shouldn't know about the simulator.


dpaiton (approved):
squeaky clean

PR #240, opened 2023-02-22
Adding flexibility to the testing framework

villanuevawill:
I noticed a number of issues surrounding the way we are currently doing tests and did a pass through and cleanup.
TODO:
- [ ] rewrite logic


villanuevawill:
@wakamex @dpaiton if you all can take a quick look it would be great

dpaiton (changes requested):
Almost there! Just have to rewrite all of them to have matched coverage.

PR #242, opened 2023-02-22
Fix Annualization in Calculations

sentilesdal:
- don't annualize twice
- add annualized time to lp token calc equations


dpaiton (approved):
yay bug fixes!!

PR #247, opened 2023-02-23
make bororw market inherit from base market

wakamex:


dpaiton (approved):
print sucks do better

PR #253, opened 2023-02-26
Integrated smart contracts into sim repo with jupyter notebook & ape.

villanuevawill:
Done. Readme attached.


villanuevawill:
image

notebook/apeworx.ipynb
```diff
@@ -0,0 +1,455 @@
+{
```
dpaiton, line 1: You'll need to "Clear all outputs", save, then push up again to pass CI.

notebook/contracts/IERC20.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: MIT
```
dpaiton, line 1: Can you move your `notebook` folder into `examples/notebooks` pls?

villanuevawill: Unfortunately it won't work this way.

villanuevawill: Current tooling needs everything in a notebook folder. We can update their core repo, but I think it makes more sense to have our notebooks in a notebook folder anyways.

sentilesdal (approved):
lgtm

PR #260, opened 2023-02-27
Add addLiquidity tests

sentilesdal:
Submitted as a draft right now.  Writing these tests exposed that we aren't doing the accounting the same way as we are in the smart contracts: `lpShares = (dz * l) / (z + a_s - a_l)` where `a_s` and `a_l` are the short and long adjustments.  In order to calculate these we need to keep track of the long and short base volumes, amounts outstanding and average maturity times.

### DIFF
--- a/elfpy/markets/hyperdrive.py
+++ b/elfpy/markets/hyperdrive.py
@@ -40,6 +40,7 @@ class MarketActionType(Enum):
 @dataclass
 class MarketDeltas(base_market.MarketDeltas):
     r"""Specifies changes to values in the market"""
+    # pylint: disable=too-many-instance-attributes
     d_base_asset: float = 0
     d_bond_asset: float = 0
     d_base_buffer: float = 0
@@ -319,7 +320,7 @@ class Market(base_market.Market[MarketState, MarketDeltas]):
                 )
             else:
                 raise ValueError(f'ERROR: Unknown trade type "{agent_action.action_type}".')
-        except Exception:
+        except AssertionError:
             logging.debug(
                 "TRADE FAILED %s\npre_trade_market = %s",
                 agent_action,
@@ -856,7 +857,7 @@ class Market(base_market.Market[MarketState, MarketDeltas]):
 
         return (base_amount - (1 - normalized_time_remaining) * bond_amount) / normalized_time_remaining
 
-    def update_weighted_average(
+    def update_weighted_average(  # pylint: disable=too-many-arguments
         self,
         average: float,
         total_weight: float,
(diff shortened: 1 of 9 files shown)


elfpy/markets/hyperdrive.py
```diff
@@ -414,23 +473,52 @@ def close_short(
...
+        # Make sure the trade is valid
         self.pricing_model.check_output_assertions(trade_result=trade_result)
+
+        # TODO: add accounting for withdrawal shares
+
         # Return the market and wallet deltas.
+        base_volume = self.calculate_base_volume(
+            trade_result.market_result.d_base, bond_amount, time_remaining.normalized_time
+        )
+        print("\nclose_short")
```
dpaiton, line 502: remove prints

elfpy/markets/hyperdrive.py
```diff
@@ -463,13 +550,8 @@ def open_long(
         tuple[MarketDeltas, wallet.Wallet]
             The deltas that should be applied to the market and agent
         """
-        if trade_amount > self.market_state.bond_reserves:
```
dpaiton, line 466: why are you removing this? I think we need it for the open long tests

elfpy/markets/hyperdrive.py
```diff
@@ -599,7 +740,7 @@ def add_liquidity(
             time_remaining=self.position_duration,
         )
         # perform the trade
-        lp_out, d_base_reserves, d_bond_reserves = self.pricing_model.calc_lp_out_given_tokens_in(
+        lp_out, d_base_reserves, d_bond_reserves = self.calc_lp_out_given_tokens_in(
```
dpaiton, line 743: can you also remove this from the pricing model so we don't have it in two places?

tests/test_max_short.py
```diff
@@ -6,11 +6,17 @@
...
+from pytest import skip
+
 import utils_for_tests as test_utils  # utilities for testing
 
 import elfpy.simulators.simulators as simulators
 import elfpy.utils.outputs as output_utils  # utilities for file outputs
 
+# TODO: rewrite these tests wthout the simulator, and actually verify values are correct or at least
+# sensical.
+skip(msg="These tests aren't actually testing anything.", allow_module_level=True)
```
dpaiton, line 18: you said this is working now right? can we remove the skip then?

dpaiton (approved):
Looks good -- I think we need to sync up on when/why we use newlines to separate code. Pep8 recommends "sparing" use. I'm mostly just concerned with consistency. Anyways, we can talk about it another time. Thanks for this! It was clearly a lot of work. Appreciate it.

tests/test_max_short.py
sentilesdal: just tried again, all failing.  something to do with the simulator.  this will hopefully get covered soon by the solidity tests

elfpy/markets/hyperdrive.py
sentilesdal: i think i may have done this during a rebase.  i'll it back in.

PR #261, opened 2023-02-27
remove calc_liquidity and use initialize_market

dpaiton:


tests/test_market.py
```diff
@@ -424,3 +426,160 @@ def test_apr(self):
...
             self.assertAlmostEqual(market.fixed_apr, target_apr, 12)
+
+    def test_market_init(self):
+        """Unit tests for the pricing model calc_liquidity function
+
+        Example check for the test:
+            test 1: 5M target_liquidity; 5% APR;
+            6mo remaining; 22.186877016851916 time_stretch (targets 5% APR);
+            1 init share price; 1 share price
+                l = target_liquidity = 5_000_000
```
wakamex, line 437: why the indent?

tests/test_trade.py
```diff
@@ -57,19 +57,13 @@ def test_market_init_apr_and_liquidity(self):
...
-                        )
+                        # TODO: Redo this to compute the direct reserves exactly instead of calling initialize_market
+                        #       this ensures that the underlying function is not changed
                         market_direct = Market(
                             pricing_model=simulator.market.pricing_model,
                             market_state=MarketState(
-                                share_reserves=share_reserves_direct,
-                                bond_reserves=bond_reserves_direct,
+                                share_reserves=0,  # set later
+                                bond_reserves=0,  # set later
```
wakamex, line 66: can do this without passing anything in now, as per [elfpy/utils/sim_utils.py](https://github.com/element-fi/elf-simulations/pull/261/files#diff-8cae30f02fc039d47994eac409e2dd37e9ff54d6dcf4a27378488173198ceb34)

wakamex (commented):
we only have initialize_market for hyperdrive, right? so this approach doesn't work with yieldspace? or should we put this in `base.py`?

tests/test_market.py
dpaiton: fixed

dpaiton:
> we only have initialize_market for hyperdrive, right? so this approach doesn't work with yieldspace? or should we put this in `base.py`?

Market init is the same for both hyperdrive & yieldspace (they both use the hyperdrive market type), but for coverage sake I went ahead and included yieldspace in the test. I also made a borrow market test, but it's kind of silly because the borrow market doesn't need to init with any reserves presently. I didn't put the signature in `base.py` because it would have different arguments for different markets.

wakamex:
we currently test the borrow market in `test_borrow.py` do we want to merge these borrow market tests into that file, or vice versa?

tests/test_market.py
```diff
@@ -424,3 +426,160 @@ def test_apr(self):
...
+                "init_share_price": 1.3,  # original share price when pool started
+                "share_price": 1.3,  # share price of the LP in the yield source
+                "expected_share_reserves": 7_692_307.692307692,
+                "expected_bond_reserves": 6_486_058.016848019,
+            },
+        ]
+        # Loop through the test cases & pricing model
+        for test_index, test_case in enumerate(test_cases):
+            # Check if this test case is supposed to fail
+            if "is_error_case" in test_case and test_case["is_error_case"]:
```
sentilesdal: There are no error cases, I'd say remove this or add some?   `initialize_market` takes `contribution` and `target_apr` so we should make sure that zero liquidity, zero apr error cases are covered.

sentilesdal (approved):
nice cleanup, thanks!

tests/test_market.py
sentilesdal: gonna mark this as resolved and do this myself so we can land

PR #262, opened 2023-02-28
Reorganize time utils to be in a dedicated folder in elfpy

dpaiton:


elfpy/time/time.py
```diff
@@ -33,18 +50,6 @@ def normalized_time(self):
...
-            f"\t{self.normalized_time=},\n"
-            f"\t{self.stretched_time=},\n"
-            f"\t{self.time_stretch=},\n"
-            f"\t{self.normalizing_constant=},\n"
-            ")"
-        )
-        return output_string
-
 
 def current_datetime() -> datetime:
```
wakamex, line 54: i feel like we could remove all the datetime functions, convert from time to datetime in post-op if desired

wakamex (approved):
lgtm

PR #266, opened 2023-03-01
SC Deploy in ape

villanuevawill:
- Apeworx deploys contracts directly in notebooks
- Reorganized to use existing examples folder
- Configs added
- Import github library (when no longer private no need for npm link)
- README updated


dpaiton (approved):
🎉

README.md
```diff
@@ -56,18 +56,20 @@ docker run -it --name elf-simulations-dev --rm --volume $(pwd):/app/ --net=host
...
 
-First, open the hyperdrive contracts repo on the ape-jupyter-sc-integration branch. Follow instructions for nstalling forge, then run anvil.
-Next, deploy contracts by running `yarn testnet`
+run:
+```
+anvil
+```
 
-Then in the sim repo run:
+npm link the hyperdrive repo, then run:
```
sentilesdal, line 66: I'd prefer that we use `pip` to do this, or straight up symlink or git clone instead of introducing npm and node_modules.

sentilesdal (commented):
Can we avoid adding node_modules to this repo just for this?

README.md
villanuevawill: pip wouldn't work, symlink is fine, git clone is part of npm link

PR #270, opened 2023-03-01
pricing model assert before trying to find root of negative number

wakamex:


elfpy/pricing_models/yieldspace.py
```diff
@@ -329,9 +329,9 @@ def calc_in_given_out(
             # d_y' = (k - (c / mu) * (mu * (z - d_z))**(1 - tau))**(1 / (1 - tau)) - (2y + cz)
             #
             # without_fee = d_y'
-            without_fee = (k - scale * (init_share_price * (out_reserves - d_shares)) ** time_elapsed) ** (
-                1 / time_elapsed
-            ) - in_reserves
+            base_of_exponent = init_share_price * (out_reserves - d_shares)
+            assert base_of_exponent > 0, f"yieldspace.calc_in_given_out() ERROR: {base_of_exponent=} <= 0"
```
dpaiton, line 333: I've been trying to move our error checking in the direction of raising errors instead of asserting. For example, this would be:
```
if base_of_exponent > 0:
    raise ValueError(f"ERROR: {base_of_exponent=} <= 0")
```

No need to include the location (`yieldspace.calc_in_given_out()`) because that'll be in the stack trace already. This also makes it easier to change up the error type down the road if we decide we want to catch it.

dpaiton (approved):

Issue #272, opened 2023-03-01
make logging safer without relative paths


wakamex:
currently unsafe to use relative paths, if the assumption that notebooks are run from the same working directory doesn't hold. this would not hold in the following cases (cases 1 and 3 have occured to me):
- `test_notebooks.py` executes
- someone moves a notebook
- a notebook is converted to .py (using something like jupytext) then calls it from a different folder

suggested solutions:
- store the base folder in elfpy's `__init__.py` then making logging relative to that
- make all logging relative to current working directory (we'd end up with multiple logging folders)

PR #273, opened 2023-03-01
make logging safer

wakamex:
* don't need to specify folder, all logging happens in .logging in base folder
* can pass in with or without .log at the end of the log_filename

fixes issue #272


wakamex (commented):
lgtm

sentilesdal (approved):

PR #284, opened 2023-03-07
make pricing model errors more explicit

wakamex:
* make pricing model errors more explicit


wakamex:
this will need to be updated slightly after the incoming changes to the pricing model in another PR

wakamex:
more explicit errors make it harder to grok the math, not needed

PR #286, opened 2023-03-08
renames calc_bond_reserves to calc_initial_bond_reserves

dpaiton:
…ropriate calc_bond_reserves


elfpy/pricing_models/base.py
```diff
@@ -124,11 +120,51 @@ def calc_bond_reserves(
...
+        float
+            The expected amount of bonds (token asset) in the pool, given the inputs
+
+        .. todo:: Test this function
+        """
+        # Only want to renormalize time for APR ("annual", so hard coded to 365)
+        annualized_time = time.norm_days(time_remaining.days, 365)
+        # (1 + r * t) ** (1 / tau)
+        interest_factor = (1 + target_apr * annualized_time) ** (1 / time_remaining.stretched_time)
+        # mu * z * (1 + apr * t) ** (1 / tau) - l
```
wakamex, line 164: where does this formula come from?
i'm lost as to what interest_factor is meant to accomplish
if annualized_time is a fraction of a year, then the amount of interest at the end of the period would be `(1 + r * t) ** t` but that's APY (with compounding) not APR. not sure why stretched time would factor in.

wakamex (approved):
lgtm

PR #288, opened 2023-03-08
fix market get_max_short

wakamex:
the market get max short function currently under-estimates the max max, but otherwise works :)

fixes: #440


wakamex:
needs tests to confirm this works, and some additional tests to check that shorting $1 more than calculated indeed does not pass

wakamex:
closing for now, not sure if it has any relevance. saving to my fork to review later, if needed.

PR #293, opened 2023-03-10
Refactor Checkpointing Logic

sentilesdal:
Cleaning up the checkpointing logic a bit here, refactoring into methods so that open/close long/short methods in the hyperdrive market can stay clean.  I also experimented with pulling all the checkpointing code into it's own file but I think that's a little overkill for now.  going to land this and revisit.


wakamex (approved):
lgtm

dpaiton (approved):

PR #296, opened 2023-03-14
fixes close long test to include burning total supply

dpaiton:


dpaiton:
closing as it is addressed in #297

PR #298, opened 2023-03-14
moves random agent from hyperdrive into its own policy

dpaiton:


elfpy/agents/policies/random_agent.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def __init__(self, rng: numpyGenerator, trade_chance: float, wallet_address: int, budget: int = 10_000) -> None:
+        """Adds custom attributes"""
+        self.trade_chance = trade_chance
+        self.rng = rng
+        super().__init__(wallet_address, budget)
+
+    def get_available_actions(
+        self,
+        disallowed_actions: "list[hyperdrive_actions.MarketActionType] | None" = None,
```
sentilesdal, line 27: prefer non quoted types.

elfpy/agents/policies/random_agent.py
```diff
@@ -0,0 +1,152 @@
...
+    ) -> "list[hyperdrive_actions.MarketActionType]":
```
sentilesdal, line 28: same here, won't comment on the rest.

sentilesdal (changes requested):

examples/notebooks/hyperdrive.ipynb
```diff
@@ -137,122 +135,47 @@
...
+        "class RandomAgent(random_agent.Policy):\n",
+        "    \"\"\"Agent that randomly opens or closes longs or shorts\n",
+        "\n",
+        "    Customized from the policy in that one can force the agent to only open longs or shorts\n",
         "    \"\"\"\n",
         "\n",
         "    def __init__(self, rng: Generator, trade_chance: float, wallet_address: int, budget: int = 10_000) -> None:\n",
         "        \"\"\"Add custom stuff then call basic policy init\"\"\"\n",
         "        self.trade_long = True  # default to allow easy overriding\n",
         "        self.trade_short = True  # default to allow easy overriding\n",
```
wakamex, line 147: why not add these to the random agent class? it already has some customizability in terms of `trade_chance` this would just add the ability for directional randomnes. that would save you from having to sub-class anything in this notebook, i think.

wakamex: doing so would eliminate the instructive benefit of showing how to sub-class an agent tho 🤔

wakamex (approved):
lgtm, just left a comment above

elfpy/agents/policies/random_agent.py
dpaiton: We get a lint error if we use non quoted types, and I think quoted types are better than lint ignore statements

dpaiton: nvm -- just needed to import annotations. Fixed.

examples/notebooks/hyperdrive.ipynb
dpaiton: If we find that we are subclassing for the same reason a lot down the road we can add it in.

sentilesdal (approved):

PR #302, opened 2023-03-14
use Decimal throughout codebase

wakamex:
* check that round trips have slippage
* use Decimal throughout the codebase, wth the following exceptions
 - config
 - block time (defined manually, never affected by pricing model)
 - mint_time (this is a reflection of block time)
 - weighted averages (because the amounts are weighted by time, a float)
*


wakamex:
closed by #334

PR #307, opened 2023-03-21
Get trade tape

dpaiton:


elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+class AssetIdPrefix(IntEnum):
+    r"""The asset ID is used to encode the trade type in a transaction receipt"""
+    LONG = 0
+    SHORT = 1
+    WITHDRAWAL_SHARE = 2
+
+
+def encode_asset_id(prefix: int, timestamp: int) -> int:
```
sentilesdal, line 28: nice! but `encode_asset_id` and `decode_asset_id` definitely don't belong in apeworx_integrations.py.  definitely more of a hyperdrive_market.py thing.  They are very specific to the implementation of the hyperdrive smart contract

elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+    tuple[int, int]
+        identifier, timestamp
+    """
+    prefix_mask = (1 << 248) - 1
+    prefix = asset_id >> 248  # shr 248 bits
+    timestamp = asset_id & prefix_mask  # apply the prefix mask
+    return prefix, timestamp
+
+
+def get_transaction_trade_event(tx_receipt: Receipt) -> ContractLog:
```
sentilesdal, line 79: we should be absolutely clear that this is just for the hyperdrive contract.  could incidentally work for others, but not for sure.

elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+        logging.debug("hyperdrive_pool_state=%s", pool_state)
+    return pool_state, tx_receipt
+
+
+def ape_close_position(
+    trade_prefix: AssetIdPrefix,
+    hyperdrive_contract: ContractInstance,
+    agent_address: AccountAPI,
+    bond_amount: int,
+    maturity_time: int,
```
sentilesdal, line 174: is this in seconds or years?  add to the comment, its gonna be hard to keep track :P

elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+                maturity_time,
+                trade_bond_amount,
+                min_output,
+                agent_address,
+                as_underlying,
+            )
+        else:
+            raise ValueError(f"{trade_prefix=} must be 0 (long) or 1 (short).")
+        # Return the updated pool state & transaction result
+        transfer_single_event = get_transaction_trade_event(tx_receipt)
```
sentilesdal, line 240: oh i see, ok maybe we rename it to just `get_transfer_single_event` and then we can use it wherever (like here) where we know that TransferSingle has what we want

elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+            trade_bond_amount = bond_amount
+        else:
+            trade_bond_amount = agent_balance
+            logging.warning(
+                "bond_amount=%g is greater than or equal to the Hyperdrive pool balance=%g",
+                bond_amount,
+                trade_bond_amount,
+            )
+        # Close the position
+        min_output = 0
```
sentilesdal, line 219: ok so we allowing for maximum slippage here.  I would make this a function argument that has `min_output = 0` for now.

elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+    Returns
+    -------
+    Tuple[dict[str, Any], ape_ethereum.transactions.Receipt]
+        A tuple containing an update dictionary for the Hyperdrive pool state
+        as well as the Ape-ethereum transaction `receipt
+        <https://docs.apeworx.io/ape/stable/methoddocs/api.html#ape.api.transactions.ReceiptAPI>`_.
+    """
+    with ape.accounts.use_sender(agent_address):  # sender for contract calls
+        if trade_prefix == AssetIdPrefix.LONG:  # open a long
+            min_output = 0
```
sentilesdal, line 136: move as function argument as well.  i have a better comment regarding same thing in next function.

elfpy/utils/apeworx_integrations.py
```diff
@@ -0,0 +1,250 @@
...
+            as_underlying = False  # mockHyperdriveTestNet does not support as_underlying=True
+            tx_receipt = hyperdrive_contract.openLong(  # type: ignore
+                trade_amount,  # base
+                min_output,
+                agent_address,
+                as_underlying,
+            )
+        elif trade_prefix == AssetIdPrefix.SHORT:  # open a short
+            max_deposit = trade_amount
```
sentilesdal, line 145: this feels... strange.  the `max_deposit` is the `trade_amount`?  this is probably a temporary thing for now?  add a comment.  i guess this works as a calc_in_given_out kinda thing?  makes me think we'll probably want ape_open_short and ape_open_long since the latter has a `min_out` and the former will have a `max_in`

sentilesdal (commented):
very nice!  happy to talk about comments IRL

sentilesdal (approved):
shipit!

elfpy/markets/borrow.py
```diff
@@ -179,6 +179,11 @@ def borrow_rate(self) -> float:
         """The borrow rate is the lending rate multiplied by the spread ratio"""
         return self.market_state.lending_rate * self.market_state.spread_ratio
 
+    @property
+    def latest_checkpoint_time(self) -> float:
```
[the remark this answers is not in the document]
dpaiton: yeah -- lint was upset the borrow wasn't overriding it bc it is specified in the child class as not implemented

PR #308, opened 2023-03-23
convert float to a float so accidentally creating an agent with a bud…

wakamex:
…get of int type, like 0 instead of 0.0, does not cause problems down the line, like with assertEqual failing because it receives an unexpected type


wakamex:
not necessary, I found another bigger issue that was causing this problem

PR #314, opened 2023-03-29
add remove liquidity tests

sentilesdal:
This PR adds the equivalent remove liquidity tests in python for hyperdrive.  It also contains a fair amount of the code needed to complete the fair withdrawal shares, though some will need to be added when closing longs/shorts.

### DIFF
--- a/elfpy/markets/hyperdrive/hyperdrive_market.py
+++ b/elfpy/markets/hyperdrive/hyperdrive_market.py
@@ -735,7 +735,7 @@ class Market(
         market_deltas.withdraw_interest -= float(recovered_interest)
 
         # Withdraw for the user
-        base_proceeds = self._withdraw(recovered_margin + recovered_interest, as_underlying)
+        base_proceeds = self._withdraw(float(recovered_margin + recovered_interest), as_underlying)
         # TODO: figure out how to keep track of hyperdrive's base asset amount.  market_deltas has
         # a d_base_asset, but that is used to update the share_reserves :/.
         # market_deltas.d_base_asset -= base_proceeds
--- a/elfpy/markets/hyperdrive/hyperdrive_market.py
+++ b/elfpy/markets/hyperdrive/hyperdrive_market.py
@@ -6,7 +6,6 @@ import logging
 from collections import defaultdict
 from dataclasses import dataclass, field
 from enum import IntEnum
-from typing import Union
 
 import numpy as np
 
@@ -15,7 +14,6 @@ import elfpy.errors.errors as errors
 import elfpy.markets.base as base_market
 import elfpy.markets.hyperdrive.hyperdrive_actions as hyperdrive_actions
 import elfpy.pricing_models.hyperdrive as hyperdrive_pm
-import elfpy.pricing_models.yieldspace as yieldspace_pm
 import elfpy.time as time
 import elfpy.types as types
 import elfpy.utils.price as price_utils
--- a/elfpy/markets/hyperdrive/hyperdrive_market.py
+++ b/elfpy/markets/hyperdrive/hyperdrive_market.py
@@ -768,3 +768,44 @@ class Market(
         # Get the amount of base to transfer.
         amount_withdrawn = shares * self.market_state.share_price
         return amount_withdrawn
+
+    def _free_margin(
+        self, freed_capital: float, max_capital: float, interest: float
+    ) -> hyperdrive_actions.MarketDeltas:
+        """
+        Moves capital into the withdraw pool and marks shares ready for withdraw.
+
+        Parameters
+        ----------
+        freed_capital: float
+            The amount of capital to add to the withdraw pool, must not be more than the max capital.
+        max_capital: float
+            The margin which the LP used to back the position which is being closed.
+        interest: float
+            The interest earned by this margin position, fixed interest for LP shorts and variable for longs.
+
+        Returns
+        -------
+        hyperdrive_actions.MarketDeltas
+            the capital, interest and shares added to the withdraw pool
+        """
+
+        # If we don't have capital to free then simply return zero
+        withdraw_share_supply = self.market_state.total_supply_withdraw_shares
+        withdraw_shares_ready_to_withdraw = self.market_state.withdraw_shares_ready_to_withdraw
+        if withdraw_share_supply <= withdraw_shares_ready_to_withdraw:
+            return 0, 0
+
+        # If we have more capital freed than needed we adjust down all values
+        withdraw_pool_deltas = hyperdrive_actions.MarketDeltas()
+        if max_capital + withdraw_shares_ready_to_withdraw > withdraw_share_supply:
+            # in this case we want max_capital * adjustment + withdraw_shares_ready_to_withdraw = withdraw_share_supply
+            # so adjustment = (withdraw_share_supply - withdraw_shares_ready_to_withdraw) / max_capital
+            # we adjust max_capital and do corresponding reduction in freed_capital and interest
+            adjustment = withdraw_share_supply - withdraw_shares_ready_to_withdraw / max_capital
+            withdraw_pool_deltas.withdraw_shares_ready_to_withdraw = max_capital * adjustment
+            withdraw_pool_deltas.withdraw_capital = freed_capital * adjustment
+            withdraw_pool_deltas.withdraw_interest = interest * adjustment
+
+        # Finally return the amount used by this action and the caller can update reserves.
+        return withdraw_pool_deltas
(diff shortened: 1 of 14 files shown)


elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -89,37 +87,37 @@ def __setitem__(self, key, value):
...
     # trading buffers
-    base_buffer: float = field(default=0.0)
-    bond_buffer: float = field(default=0.0)
+    base_buffer: float = 0.0
+    bond_buffer: float = 0.0
 
     # share price
-    variable_apr: float = field(default=0.0)
+    variable_apr: float = 0.0
     share_price: float = field(default=1.0)
```
dpaiton, line 102: might as well delete field for this & init_share_price

dpaiton:
Can you tighten up the extra newlines? [PEP8](https://peps.python.org/pep-0008/#blank-lines) recommends only using blank lines sparingly in functions to indicate logical sections. I prefer to use [block comments](https://peps.python.org/pep-0008/#block-comments) for this purpose, and not both.

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -549,13 +547,13 @@ def add_liquidity(
...
     ) -> tuple[hyperdrive_actions.MarketDeltas, wallet.Wallet]:
         """Computes new deltas for bond & share reserves after liquidity is removed"""
         self.apply_checkpoint(self.latest_checkpoint_time, self.market_state.share_price)
         market_deltas, agent_deltas = hyperdrive_actions.calc_remove_liquidity(
             wallet_address=agent_wallet.address,
-            bond_amount=bond_amount,
+            lp_shares=lp_shares,
             market=self,
         )
         self.market_state.apply_delta(market_deltas)
```
dpaiton, line 559: I noticed that some times we use `self.update_market(market_deltas)` (e.g. in `redeem_withdraw_shares`) and sometimes we use `self.market_state.apply_delta(market_deltas` (e.g. here). I think we should always use `self.update_market()` since it is doing the same thing with additional checks.

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -642,3 +640,169 @@ def apply_checkpoint(self, checkpoint_time: float, share_price: float) -> float:
...
+        # Update the pool state
+        # Note - Will revert here if not enough margin has been reclaimed by checkpoints or by
+        #  position closes
+        market_deltas.withdraw_shares_ready_to_withdraw -= shares
+        market_deltas.withdraw_capital -= recovered_margin
+        market_deltas.withdraw_interest -= recovered_interest
+
+        # Withdraw for the user
+        base_proceeds = self._withdraw(recovered_margin + recovered_interest, as_underlying)
+        # TODO: figure out how to keep track of hyperdrive's base asset amount.  market_deltas has
```
dpaiton, line 730: Can't you add `base_reserves` to `hyperdrive_market.MarketState`?

dpaiton: I assume the reason why you can't compute `base_reserves = share_reserves * share_price` is because the share price is variable and so you'd want to do track update the base reserves each time withdraw_shares are redeemed?

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -642,3 +640,169 @@ def apply_checkpoint(self, checkpoint_time: float, share_price: float) -> float:
...
+        as_underlying: bool
+            If true, the user is paid in underlying, if false the contract transfers in yield source
+            directly. Note - for some paths one choice may be disabled or blocked.
+
+        Returns
+        -------
+        tuple[hyperdrive_actions.MarketDeltas, wallet.Wallet]
+
+        """
+        market_deltas = hyperdrive_actions.MarketDeltas()
```
dpaiton, line 703: I'm thinking this should all be done with Decimal to match our use of it in the other `calc_` functions? Of course we might change it in a follow-up, but might as well stick to the pattern for now. This will help identify where changes need to be made later anyway.

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -642,3 +640,169 @@ def apply_checkpoint(self, checkpoint_time: float, share_price: float) -> float:
...
+        wallet_deltas = wallet.Wallet(address=0)
```
dpaiton, line 704: Can you add a TODO here to not use a wallet. You can reference issue #315

dpaiton: using address=0 is not chill bc that address is more commonly used for the init_lp agent

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -642,3 +640,169 @@ def apply_checkpoint(self, checkpoint_time: float, share_price: float) -> float:
...
+
+        # Get the amount of base to transfer.
+        amount_withdrawn = shares * self.market_state.share_price
+        return amount_withdrawn
+
+    def _free_margin(
+        self, freed_capital: float, max_capital: float, interest: float
+    ) -> hyperdrive_actions.MarketDeltas:
+        """
+        Moves capital into the withdraw pool and marks shares ready for withdraw.
```
dpaiton, line 773: Can you put the brief summary doc on the same line as the """, per [pep8](https://peps.python.org/pep-0008/#documentation-strings) and [the pep followed by sphinx numpydocs](https://peps.python.org/pep-0257/#multi-line-docstrings). You can optionally add a longer-form description below. This applies elsewhere as well -- I'll just comment here but I mean everywhere.

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -642,3 +640,169 @@ def apply_checkpoint(self, checkpoint_time: float, share_price: float) -> float:
...
+            The amount of capital to add to the withdraw pool, must not be more than the max capital.
+        max_capital: float
+            The margin which the LP used to back the position which is being closed.
+        interest: float
+            The interest earned by this margin position, fixed interest for LP shorts and variable for longs.
+
+        Returns
+        -------
+        hyperdrive_actions.MarketDeltas
+            the capital, interest and shares added to the withdraw pool
```
dpaiton, line 787: I think this should be a complete sentence

elfpy/markets/hyperdrive/hyperdrive_market.py
```diff
@@ -642,3 +640,169 @@ def apply_checkpoint(self, checkpoint_time: float, share_price: float) -> float:
...
+        Moves capital into the withdraw pool and marks shares ready for withdraw.
+
+        Parameters
+        ----------
+        freed_capital: float
+            The amount of capital to add to the withdraw pool, must not be more than the max capital.
+        max_capital: float
+            The margin which the LP used to back the position which is being closed.
+        interest: float
+            The interest earned by this margin position, fixed interest for LP shorts and variable for longs.
```
dpaiton, line 782: is this supposed to be "(add & remove) LP and shorts" or is "LP shorts" a thing? Is it the opposing trade to an agent opening a long?

elfpy/pricing_models/hyperdrive.py
```diff
@@ -324,3 +324,39 @@ def calc_out_given_in(
                 with_fee=float(flat_with_fee + Decimal(curve.breakdown.with_fee)),
             ),
         )
+
+    def calc_tokens_out_given_lp_in(
+        self, lp_in: float, market_state: hyperdrive_market.MarketState
+    ) -> tuple[float, float]:
+        """
+        Calculates the amount of base shares and bonds released from burning a a specified amount of
```
dpaiton, line 332: a a

elfpy/markets/hyperdrive/hyperdrive_market.py
sentilesdal: yeah it's not as simple as that. share_reserves are in shares, not base.  I think the only base would come from the withdraw pool.  there are spots in the solidity where they use the base token balance of the contract, so we'll need the same thing but gotta figure out where that balance comes from first.

elfpy/markets/hyperdrive/hyperdrive_market.py
sentilesdal: I don't think I've touched any of that logic in this PR.  I'll do a cleanup PR, I made a [clickup ticket](https://app.clickup.com/t/866a078gp) for this.  Note though that for remove_liquidity we can't use update_market because it can make reserves zero, which update_market protects against.

elfpy/markets/hyperdrive/hyperdrive_market.py
sentilesdal: LP was a typo there, copy pasta'd from the solidity.  it now reads 'fixed interest for shorts and variable for longs.'

elfpy/markets/hyperdrive/hyperdrive_market.py
dpaiton: ok sounds good

dpaiton (approved):
looks great thanks a lot!

Issue #254, opened 2023-02-26
logging does not get deleted after tests


dpaiton:
Somewhere along the line this stopped working... need to investigate.

PR #321, opened 2023-03-31
add bot liquidity 🤖 💦

wakamex:
* allow add/remove liquidity commands for random bot
* improve ape config
* add delete_previous_logs parameter to setup_logging. default to False. closes issue #254.

fixes: #254


elfpy/agents/policies/random_agent.py
```diff
@@ -85,6 +88,42 @@ def open_long_with_random_amount(self, market) -> list[types.Trade]:
             )
         ]
 
+    def add_liquidity_with_random_amount(self) -> list[types.Trade]:
+        """Add liquidity with a random allowable amount"""
+        # take a guess at the trade amount, which should be about 10% of the agent’s budget
+        initial_trade_amount = self.rng.normal(loc=self.budget * 0.1, scale=self.budget * 0.01)
+        # WEI <= trade_amount <= max_short
```
dpaiton, line 95: This  comment does not reflect the code

elfpy/utils/apeworx_integrations.py
```diff
@@ -35,13 +35,24 @@ def get_transfer_single_event(tx_receipt: Receipt) -> ContractLog:
         <https://docs.apeworx.io/ape/stable/methoddocs/types.html#ape.types.ContractLog>`_
         ) event, excluding periferal events.
     """
-    single_events: list[ContractLog] = []
-    for tx_event in tx_receipt.events:
-        if tx_event.event_name == "TransferSingle":
-            single_events.append(tx_event)
-    if len(single_events) != 1:
-        raise ValueError(f'The transaction receipt should have one "TransferSingle" event, not {len(single_events)}.')
-    return single_events[0]
```
dpaiton, line 44: I think the new logic avoids the undesired condition of multiple single events (`!=1`). You'll want to keep that one in, we don't want to just automatically return the [-1] index.

elfpy/utils/apeworx_integrations.py
```diff
@@ -35,13 +35,24 @@ def get_transfer_single_event(tx_receipt: Receipt) -> ContractLog:
...
+    if single_events := [tx_event for tx_event in tx_receipt.events if tx_event.event_name == "TransferSingle"]:
+        return single_events[-1]  # return last item
+    raise ValueError(f'The transaction receipt should have one "TransferSingle" event, not {len(single_events)}.')
+
+
+def get_pool_state(tx_receipt, hyperdrive_contract):
```
dpaiton, line 43: Need type defs & proper docstring

elfpy/utils/outputs.py
```diff
@@ -434,6 +435,9 @@ def setup_logging(
             log_dir = os.path.join(base_folder, ".logging")
         if not os.path.exists(log_dir):  # create log_dir if necessary
             os.makedirs(log_dir)
+        # delete the log file if it exists
+        if delete_previous_logs and os.path.exists(os.path.join(log_dir, log_name)):
```
dpaiton, line 439: Does this fix https://github.com/element-fi/elf-simulations/issues/254?

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -224,7 +222,7 @@
     ")\n",
     "ax.set_xlabel(\"Trade number\")\n",
     "ax.set_ylabel(\"Outstanding balance\")\n",
-    "ax.set_title(\"Random longs & shorts\")\n",
+    "ax.set_title(f\"Bots in Python (n={len(sim_trades_df)})\")\n",
```
dpaiton, line 225: "Random bots" pls

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -512,7 +527,7 @@
     "ax.step(range(1, len(trades_df)), trades_df[\"longsOutstanding_\"].iloc[1:] / 1e18, label=\"Longs\")\n",
     "ax.set_xlabel(\"Trade number\")\n",
     "ax.set_ylabel(\"Outstanding balance\")\n",
-    "ax.set_title(\"Random longs & shorts\")\n",
+    "ax.set_title(f\"Bots on Solidity (n={len(trades_df)})\")\n",
```
dpaiton, line 530: Random bots

examples/notebooks/simple_ape.ipynb
```diff
@@ -0,0 +1,235 @@
+{
```
dpaiton, line 1: Why are you adding this notebook?

dpaiton (changes requested):
my plots in that random bots notebook wouldn't show liquidity modifications
can you also add plots to verify that the lp token volumes in the sim agents matches the solidity agents?

elfpy/utils/outputs.py
wakamex: ya, it actually deletes at the start of a run, which I think is better (all optional, set to off by default)
found it useful when re-running repeatedly and checking logs, to avoid having to delete it manually in between each check

examples/notebooks/simple_ape.ipynb
wakamex: found it useful to use something like this when debugging ape. bots_on_solidity used to take 30 seconds to get to the ape part (it's faster now). but it's still nice to have something that is a simple test of ape. both for tessting, and also to show users a clean ape implementation, if they want to test their own setup, or take it in a different direction than we did with bots_on_solidity.

elfpy/utils/apeworx_integrations.py
wakamex: added another filter to exclude id=0 events only when we have more than one to choose from. this seems safe, since id is the field we're decoding, so we want the entry that's non-empty, if possible. don't see any edge cases anymore. diff here: https://github.com/element-fi/elf-simulations/pull/321/commits/9b41e095e44069907b4606a1efc9d0f2871a21d3

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -237,86 +254,90 @@
...
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "11fa3b43-48ac-4781-8285-5d6484e7c839",
+   "id": "a328b5f3",
    "metadata": {},
    "outputs": [],
    "source": [
-    "provider = ape.networks.parse_network_choice(\"ethereum:local:foundry\").__enter__()\n",
+    "provider = ape.networks.parse_network_choice(\"ethereum:local:foundry\", provider_settings={\"port\":8546}).__enter__()\n",
```
wakamex, line 307: for those of us running local nodes

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -237,86 +254,90 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8670c8fc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_spot_price_and_apr(pool_info, hyper_config, hyperdrive_math):\n",
```
wakamex, line 319: roll this into `get_pool_state` in `apeworx_integrations.py`?

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -359,19 +380,35 @@
     "fixed_math_address = sol_agents[\"agent_0\"].deploy(project.MockFixedPointMath)\n",
     "fixed_math = project.MockFixedPointMath.at(fixed_math_address)\n",
     "\n",
-    "base_ERC20.mint(int(config.target_liquidity * 10**18), sender=sol_agents[\"agent_0\"])\n",
+    "hyperdrive_math_address = sol_agents[\"agent_0\"].deploy(project.MockHyperdriveMath)\n",
+    "hyperdrive_math = project.MockHyperdriveMath.at(hyperdrive_math_address)\n",
+    "\n",
+    "base_ERC20.mint(int(config.target_liquidity * 2 * 10**18), sender=sol_agents[\"agent_0\"])\n",
```
wakamex, line 386: I thought this helped avoid a crash, but I may have been hallucinating. will need to test more thoroughly. it makes more sense that running out of Eth instead of ERC20 would potentially cause a crash.

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -359,19 +380,35 @@
...
     "time_stretch = int(1 / simulator.market.time_stretch_constant * 10**18)\n",
+    "print(f\"{time_stretch=}\")\n",
+    "actual_time_stretch = 1 / (time_stretch / 10**18)\n",
+    "print(f\"{actual_time_stretch=}\")\n",
     "curve_fee = int(config.trade_fee_percent * 10**18)\n",
     "flat_fee = int(config.redemption_fee_percent * 10**18)\n",
-    "gov_fee = 0\n",
+    "gov_fee = int(config.governance_fee_percent * 10**18)\n",
     "\n",
+    "print(\"=== SETUP ===\")\n",
```
wakamex, line 402: exhaustive prints kept in cus they were usseful to manually inspect whether our setup conditions were identical (they were)

examples/notebooks/bots_on_solidity.ipynb
```diff
@@ -384,57 +421,82 @@
...
     "pool_state[0][\"block_number_\"] = genesis_block_number\n",
+    "pool_state[0][\"spot_price\"] = spot_price\n",
+    "pool_state[0][\"apr\"] = apr\n",
     "logging.info(\"pool_state=%s\\n\", pool_state)\n",
+    "print(f\"{pool_state=}\")\n",
     "\n",
     "sim_to_block_time = {}\n",
     "trade_receipts = []\n",
-    "for trade in sim_trades:\n",
+    "for trade in sim_trades[1:]:\n",
```
wakamex, line 470: avoiding passing the first python trade to solidity, since it's our pool initialize transaction, which was already executed in the solidity `initialize` call

wakamex (commented):
going to move price and apr into get_pool_state then we should be good

wakamex (commented):
lgtm

wakamex (commented):
lgtm

dpaiton (changes requested):

elfpy/utils/apeworx_integrations.py
```diff
@@ -33,15 +33,58 @@ def get_transfer_single_event(tx_receipt: Receipt) -> ContractLog:
...
-    for tx_event in tx_receipt.events:
-        if tx_event.event_name == "TransferSingle":
-            single_events.append(tx_event)
-    if len(single_events) != 1:
-        raise ValueError(f'The transaction receipt should have one "TransferSingle" event, not {len(single_events)}.')
-    return single_events[0]
+    single_events = [tx_event for tx_event in tx_receipt.events if tx_event.event_name == "TransferSingle"]
+    if len(single_events) > 1:
+        single_events = [tx_event for tx_event in single_events if tx_event.id != 0]  # exclude token id 0
+    if len(single_events) > 1:
```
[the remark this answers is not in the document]
wakamex: yes, but after running single_events through an additional filter, which removes the edge case when we have more than 1 event (removing that one with id=0 leaves only 1 event)

tests/solidity/test_initialize_bots_on_solidity.py
```diff
@@ -0,0 +1,83 @@
+"""Market initialization tests that match those being executed in the solidity repo"""
```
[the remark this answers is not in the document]
wakamex: the point of this test is to match the exact values as solidity, where the results are copied from. if someone changed the inputs it won't match, but that defeats the point of the test.

elfpy/pricing_models/base.py
```diff
@@ -344,13 +345,13 @@ def check_input_assertions(
             "pricing_models.check_input_assertions: ERROR: "
             f"expected quantity.amount >= {WEI}, not {quantity.amount}!"
         )
-        assert market_state.share_reserves >= 0, (
```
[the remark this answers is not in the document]
wakamex: seems like the outcome of just testing more cases. the one we breached was the normalized time variable. I added it to `share_reserves` and `bond_reserves` because it felt like a missing equivalence to what we were already doing elsewhere. but looking at it now, it's redundant, so I'm just going to remove these two here, in favor of the check in `update_market`: `self.market_state.check_market_non_zero()`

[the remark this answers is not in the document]
wakamex: fixed here 81a5a0d402c3d721322c93a94d08ce6d89b8acbd

elfpy/utils/apeworx_integrations.py
```diff
@@ -55,7 +98,7 @@ def ape_open_position(
     Arguments
     ---------
     trade_prefix: hyperdrive_assets.AssetIdPrefix
-        IntEnum specifying whether the trade is a long (0) or a short (1).
+        IntEnum specifying whether the trade is a long (0), short (1), or LP (3).
```
dpaiton, line 101: This is not correct anymore

elfpy/utils/apeworx_integrations.py
```diff
@@ -74,38 +117,36 @@ def ape_open_position(
...
+            tx_receipt: Receipt = hyperdrive_contract.addLiquidity(
+                trade_amount,  # bonds
+                0,  # min apr
+                int(100 * 1e18),  # max apr
+                agent_address,
+                as_underlying,
+            )
         else:
-            raise ValueError(f"{trade_prefix=} must be 0 (long) or 1 (short).")
+            raise ValueError(f"{trade_prefix=} must be 0 (long), 1 (short), or 3 (LP)")
```
dpaiton, line 147: This is not correct anymore

elfpy/utils/apeworx_integrations.py
```diff
@@ -114,14 +155,14 @@ def ape_close_position(
...
-    maturity_time: int,
+    maturity_time: Optional[int] = None,
 ) -> tuple[dict[str, Any], Receipt]:
     r"""Close a long or short position on the Hyperdrive Solidity contract using apeworx.
 
     Arguments
     ---------
     trade_prefix: hyperdrive_assets.AssetIdPrefix
-        IntEnum specifying whether the trade is a long (0) or a short (1).
+        IntEnum specifying whether the trade is a long (0), short (1), or LP (3).
```
dpaiton, line 165: This is not correct anymore

pyproject.toml
```diff
@@ -42,3 +42,6 @@ line-length = "120"
 extend-exclude = "\\.ipynb"
 
 [tool.pyright]
+exclude = [
+    "hyperdrive_solidity",
+]
```
dpaiton, line 47: add newline

dpaiton (approved):
Couple of small comments to  fix then it is gtg

PR #322, opened 2023-03-31
add ape and codecov to CI, update to python 3.9

wakamex:
* add ape to CI
* simplify requirements.txt to remove pinned versions
* eliminate requirements-3.8.txt
* set python version to 3.9 in pyproject.toml 
* set python version to 3.9 in CI
* remove unnecessary lines from coverage, like typechecking statements
* more accurate codecov. uses pytest to check coverage, instead of its built-intesting
* remove unnecessary ape plugin (notebook is only a shortcut for running jupyter servers)
* remove separate requirements list for Colab, not needed on new Colab default python 3.9

[hyperdrive.ipynb on this branch](https://colab.research.google.com/github/element-fi/elf-simulations/blob/wakamex/clean_up_codecov/examples/notebooks/hyperdrive.ipynb#scrollTo=ww-iIxQ3fk45) works, since we've been keeping it up to date. now that [Colab is on 3.9](https://medium.com/google-colab/colab-updated-to-python-3-9-2593f8b1eb79) we don't need to pin any dependencies, it actually runs as default. I suggest we point our live demo to a recent pin after we merge this, maybe a v0.0.3 release. hyperdrive.ipynb only uses 3 dependencies: numpy, matplotlib, and pandas, whose updated versions on Colab suffice (matplotlib version is 3 years more recent!):

<table>
	<th><center>dependency
	<th><center>current<br>version
	<th><center>current<br>release date
	<th><center>previous<br>version
	<th><center>previous<br>release date
	<th><center>improvement
<tr>
	<td>matplotlib
	<td><center>3.7.1
	<td><center><a href=https://github.com/matplotlib/matplotlib/releases/tag/v3.7.1>Mar 4, 2023</a>
	<td><center>3.2
	<td><center><a href=https://github.com/matplotlib/matplotlib/releases/tag/v3.2.0>Mar 3, 2020</a>
	<td><center>3 years!
<tr>
	<td>numpy
	<td><center>1.22.4
	<td><center><a href=https://github.com/numpy/numpy/releases/tag/v1.22.4>May 20, 2022</a>
	<td><center>1.21.6
	<td><center><a href=https://github.com/numpy/numpy/releases/tag/v1.21.6>Apr 12, 2022</a>
	<td><center>1 month
<tr>
	<td>pandas
	<td><center>1.4.4
	<td><center><a href=https://github.com/pandas-dev/pandas/releases/tag/v1.4.4>Aug 31, 2022</a>
	<td><center>1.3.5
	<td><center><a href=https://github.com/pandas-dev/pandas/releases/tag/v1.3.5>Dec 12, 2021</a>
	<td><center>9 months
</table>


wakamex (commented):
rfr, can merge without `everlasting_bots.py`, i'm going to use it as a base for other stuff anyway

examples/apeworx_everlasting_bots.py
```diff
@@ -410,15 +411,31 @@ def do_trade(trade):
     """Execute agent trades on hyperdrive solidity contract"""
     agent_key = f"agent_{trade.wallet.address}"
     trade_amount = to_fixed_point(trade.trade_amount)
-    # if trade.action_type.name in ["ADD_LIQUIDITY", "REMOVE_LIQUIDITY"]:
-    #    continue  # todo
-    if trade.action_type.name == "OPEN_SHORT":
+    if trade.action_type.name == "ADD_LIQUIDITY":
```
wakamex, line 414: this is the only functional change, everything else is stylistic (sourcery), type-checks (pyright) or pylint

examples/apeworx_everlasting_bots.py
```diff
@@ -22,7 +23,7 @@
 import elfpy.types as types
 import elfpy.simulators as simulators
 import elfpy.agents.agent as agentlib
-import elfpy.markets.hyperdrive.assets as hyperdrive_assets
+import elfpy.markets.hyperdrive.assets as assets
```
sentilesdal, line 26: how it was before is the convention we've been using, should stick to it.

examples/apeworx_everlasting_bots.py
```diff
@@ -336,7 +338,7 @@ def get_agents(config):
     sol_agents = {"governance": governance}
     for agent_address, sim_agent in enumerate(sim_agents):
         sol_agent = ape.accounts.test_accounts.generate_test_account()  # make a fake agent with its own wallet
-        sol_agent.balance = to_fixed_point(sim_agent.budget)
+        sol_agent._set_balance_(to_fixed_point(sim_agent.budget))  # pylint: disable=protected-access
```
sentilesdal, line 341: might as well make a method that is public

sentilesdal (commented):
nice CI improvements!

examples/apeworx_everlasting_bots.py
wakamex: ya cus without a setter, we get this complaint: `Cannot assign member "balance" for type "TestAccountAPI"
  "int" is incompatible with "property"`

wakamex: switched to the public `set_balance(address, amount)` function in the `provider` we can access directly through `sol_agent.provider`

wakamex (commented):
lgtm

sentilesdal (approved):
LGTM, please update the repo's README now that we only have the one requirements.txt

.github/workflows/lint.yml
```diff
@@ -10,24 +10,26 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python ${{ matrix.python-version }}
+
+      - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+          python-version: 3.10.11
```
dpaiton, line 19: I would recommend pinning this to 3.10 instead of a specific subversion. Let the setup pick the latest.

examples/apeworx_everlasting_bots.py
```diff
@@ -262,91 +267,92 @@ def get_argparser() -> argparse.ArgumentParser:
 def get_config() -> simulators.Config:
     """Set config values for the experiment"""
     args = get_argparser().parse_args()
-    config = simulators.Config()
-    config.log_level = output_utils.text_to_log_level(args.log_level)
-    config.log_filename = "everlasting_bots"
-    config.title = "everlasting bot demo"
+    _config = simulators.Config()
```
dpaiton, line 270: why underscore?

.github/workflows/lint.yml
```diff
@@ -10,24 +10,26 @@ jobs:
...
+          python -m pip install .
+
       - name: Ensure clear Jupyter Notebooks
         uses: ResearchSoftwareActions/EnsureCleanNotebooksAction@1.1
+
+      - name: Run black
+        run: python -m black --check --diff .
+
+      - name: Run pylint
+        run: python -m pylint $(git ls-files '*.py')
```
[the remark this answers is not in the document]
wakamex: no, but I thought that was a goal to test ape

examples/apeworx_everlasting_bots.py
wakamex: lots of variable renamed to _variable to avoid, for example: `Redefining name 'position_duration_seconds' from outer scope (line 517)`

.github/workflows/lint.yml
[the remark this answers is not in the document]
wakamex: oh unless we want to just exclude them from lint/pyright (maybe that's what you meant). but if we go in that direction, why not exclude the entire `examples` folder?

.github/workflows/static.yml
```diff
@@ -10,19 +10,20 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python ${{ matrix.python-version }}
+
+      - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+          python-version: 3.10.11
```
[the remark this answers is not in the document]
wakamex: 3.10 worked, I just had to put it in quotes! it defaults to latest (3.10.11 right now)
implemented in commit 322a43d382d0676c4d86f7ed536a2ef48c2321eb

.github/workflows/lint.yml
wakamex: 3.10 worked, I just had to put it in quotes! it defaults to latest (3.10.11 right now)
implemented in commit 322a43d382d0676c4d86f7ed536a2ef48c2321eb

examples/apeworx_everlasting_bots.py
wakamex: so jsut a pylint thing, which would be avoided if we excluding it from linting... is that consistent with our other examples? maybe we should just exclude the whole `examples` folder

.github/workflows/lint.yml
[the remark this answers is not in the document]
dpaiton: there will be `apeworx` files that exist outside of `examples`. Presently, there is one:`utils/apeworx_integrations.py`. If you're able to get this to pass pylint & pyright without filling it with `ignore` comments, then lets include it in the CI. Otherwise, we should exclude it.

.github/workflows/static.yml
```diff
@@ -10,19 +10,20 @@ jobs:
...
-          python -m pip install --upgrade pip
-          python -m pip install -r requirements-${{ matrix.python-version }}.txt
-          python -m pip install . # Install the elfpy module
+          python -m pip install stochastic --ignore-requires-python
+          python -m pip install -r requirements.txt
+          ape plugins install .
+          python -m pip install .
+
       - name: Analysing code with pyright
         run: python -m pyright $(git ls-files '*.py' ':!:*apeworx*')
```
dpaiton, line 29: still ignoring apeworx here, but not in pylint

.github/workflows/test.yml
```diff
@@ -10,21 +10,51 @@ jobs:
...
+        run: |
+          coverage run -m pytest
+
+      - name: Generate coverage report
+        run: |
+          coverage xml -i
+          coverage html -i
+
+      - name: Upload coverage report to Codecov
+        # uses: codecov/codecov-action@v2.1.0
```
dpaiton, line 56: dead comment?

README.md
```diff
@@ -21,35 +21,29 @@ Documentation can be found [here](https://elfpy.element.fi).
...
+Set up your favorite python virutal environment with python 3.9 or 3.10 (we recommend [pyenv](https://github.com/pyenv/pyenv#how-it-works) and [virtualenv](https://github.com/pyenv/pyenv-virtualenv)). For example:
 
 ```bash
-pyenv install 3.8.16
-pyenv local 3.8.16
-python -m venv .venv
-source .venv/bin/activate
+pyenv install 3.10
+pyenv local 3.10
+pyenv virtualenv elf-env
```
dpaiton, line 29: we are calling it `venv` on purpose, keep that syntax

README.md
```diff
@@ -21,35 +21,29 @@ Documentation can be found [here](https://elfpy.element.fi).
...
-python -m pip install -e .
+pip install -r requirements.txt
+pip install -e .
 ```
 
 If you intend to improve the documentation, then you must also install the packages in `requirements-dev.txt`.
 
 ### Docker
 
+Using Docker is mostly untested, as the core team doesn't use it. But feel free to try out the below.
```
dpaiton, line 45: change "But feel free to try out the below." to "However, the following steps should get you started."

examples/apeworx_everlasting_bots.py
dpaiton: Ahh ok, right. This script defines some variables in the top-most scope (`__name__ == "__main__"`). Any variable defined there should be unique from variables that exist inside of functions. This is bc any variable defined in the global scope should still technically be accessible in the function. i.e. if we didn't overwrite `config` in get_agents, then we would be able to query it. Adding an underscore is typically used for "private" variables/functions, so I see where you're going with that now. However, I think it is more clear to rename the variable in the global context to be more specific to an instantiated config. e.g. `sim_config`, meaning "the config for this particular sim" and leaving the variable name as `config` throughout the rest of the functions. I think this makes the distinction between the config used in execution vs an internal variable more clear.

dpaiton: After writing that I see you also ran into the problem with `sim_agents` and `sol_agents`. With that in mind, and also keeping in mind that this script itself is probably not going to survive long in the repo (it's janky, will be integrated soon, was never intended to have staying power, etc), I'll revise the recommendation to just leave it alone. `_` is fine for now.

examples/apeworx_everlasting_bots.py
```diff
@@ -470,7 +494,7 @@ def do_trade(trade):
     # Instantiate the config using the command line arguments as overrides.
     config = get_config()
     # Set up ape
-    provider = ape.networks.parse_network_choice("ethereum:local:foundry").__enter__()
+    provider = ape.networks.parse_network_choice("ethereum:local:foundry", provider_settings={"port": 8546})
```
dpaiton, line 497: I don't think it's correct to remove `__enter__` as I mentioned in discord. Can you put it back in?

Also are you sure you want to hard-code the port like that? This means that the user has to be careful about their port selection when running anvil, right? If hardcoded port is necessary, can you update the readme to modify the anvil command to specify this port?

pyproject.toml
```diff
@@ -13,7 +13,7 @@ authors = [
 ]
 description = "Experiment management and market simulators by Element Finance"
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.9, <=3.10"
```
dpaiton, line 16: If we're going to continue to support colab, then we should pin the project to 3.9 and not 3.10. I'd prefer to sunset colab and bump up to 3.10, but that's outside of the scope of this PR.

README.md
wakamex: oh I noticed we're recommending pyenv-virtualenv then not using it (venv is a different virtual environment manager). was I the one who switched to recommend pyenv-virtualenv?

pyproject.toml
wakamex: I've been using 3.10 locally

examples/apeworx_everlasting_bots.py
wakamex: it's hardcoded anyway, the default is 8545 and that conflicts with my local node on my desktop. will add a note in the README.

wakamex: oh I see you're thinking of people who run anvil locally, most likely on the default 8545

wakamex: this is only in an example, so not the main codebase. do we really need to mention it in the README?

.github/workflows/static.yml
wakamex: passes in 192cb98cdca500be3f69865ab4ad4926962f78ac

examples/apeworx_everlasting_bots.py
wakamex: If someone is running anvil locally, they probably know how to manage port conflicts. Also it's not standard behavior, since we're not instructing them to do that. So I don't think we need to worry about non-standard behavior not working.

wakamex: I'll remove it since I don't need it on my laptop, and can just manually change it on my desktop

.github/workflows/lint.yml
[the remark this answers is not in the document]
wakamex: leave all the apes alone! excluding them for now (until ape gets better typing)

README.md
wakamex: switched text to point to venv, and clarify

pyproject.toml
wakamex: switched to 3.9 for Colab compatibility

wakamex (commented):
lgtm, looked at diffs again

PR #323, opened 2023-04-03
generalize precision threshold check

wakamex:
generalize precision threshold check to work for market state and wallet

### DIFF
--- a/elfpy/__init__.py
+++ b/elfpy/__init__.py
@@ -170,3 +170,30 @@ rc_params.update({"savefig.facecolor": GREY})
 
 # Set the params
 mpl.rcParams.update(rc_params)
+
+
+def check_non_zero(data):
+    r"""Checks that all market variables are non-zero within a precision threshold"""
+    # TODO: issue #146
+    # this is an imperfect solution to rounding errors, but it works for now
+    try:
+        if not isinstance(data, Dict):
+            data = data.__dict__
+    except AttributeError as exception:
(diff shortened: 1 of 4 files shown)


wakamex (commented):
lgtm, read through the diffs

elfpy/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 import shutil
 import logging
+from typing import Dict
```
dpaiton, line 5: If think if you add `from __future__ import annotations` you might be able to use `dict` instead of `Dict`. The latter is being deprecated in favor of the former in later versions of python.

elfpy/__init__.py
```diff
@@ -169,3 +170,40 @@
...
+                    key,
+                    value,
+                    PRECISION_THRESHOLD,
+                )
+                setattr(data, key, 0)
+            else:
+                assert (
+                    value > -PRECISION_THRESHOLD
+                ), f"values must be > {-PRECISION_THRESHOLD}. Error on {key} = {value}"
+        elif isinstance(value, (list, tuple, Dict)):
```
dpaiton, line 208: Wouldn't `list` and `tuple` types throw your error on line 193?

wakamex: only if it fails to convert it to a dict, which it should be able to

dpaiton (approved):
lgtm

PR #330, opened 2023-04-12
fix decimal invalid operation

wakamex:


wakamex:
the issue is unrelated

PR #342, opened 2023-04-20
allow URLs to pass line limit

wakamex:
default pylint setting allows URLs to pass the line limit, but only if they're the only thing in the comment.
we want a more permissive regex which captures lines in our docstrings where the URL is at the end of the line.

old and busted
image

new hotness
image


wakamex:
pointed to the wrong branch

PR #327, opened 2023-04-11
testnet bots

wakamex:
test random bots on testnet

PR #346, opened 2023-04-20
avoid pytest collection error by naming fp tests uniquely

wakamex:
was receiving this error locally:
```bash
_____________ERROR collecting tests_fp/pricing_models/calc_test_dataclasses.py_____________
import file mismatch:
imported module 'calc_test_dataclasses' has this __file__ attribute:
  /code/elfpy/tests/pricing_models/calc_test_dataclasses.py
which is not the same as the test file we want to collect:
  /code/elfpy/tests_fp/pricing_models/calc_test_dataclasses.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```


wakamex:
incorporated in #327

PR #349, opened 2023-04-20
FP pricing_model_utils and add_liquidity tests

dpaiton:


sentilesdal (approved):

tests_fp/solidity/test_add_liquidity_fp.py
```diff
@@ -84,85 +94,85 @@ def test_add_liquidity_with_long_immediately(self):
...
         # Ensure that the contribution was transferred to Hyperdrive.
-        self.assertEqual(market_deltas.d_base_asset, -wallet_deltas.balance.amount)
+        self.assertEqual(float(market_deltas.d_base_asset), float(-wallet_deltas.balance.amount))
 
         # Ensure the pool APR is still approximately equal to the target APR.
         pool_apr = self.hyperdrive.pricing_model.calc_apr_from_reserves(
             self.hyperdrive.market_state, self.hyperdrive.position_duration
         )
-        self.assertAlmostEqual(pool_apr, self.target_apr, 1)
+        self.assertAlmostEqual(float(pool_apr), float(self.target_apr), places=1)
```
sentilesdal, line 157: 1 place... it was like this already but let's try like 3 places.

sentilesdal (approved):
lgtm, nice fix!

tests_fp/solidity/test_add_liquidity_fp.py
```diff
@@ -84,85 +94,85 @@ def test_add_liquidity_with_long_immediately(self):
         market_deltas, wallet_deltas = self.hyperdrive.add_liquidity(self.bob.wallet, self.contribution)
 
         # Ensure that the contribution was transferred to Hyperdrive.
-        self.assertEqual(market_deltas.d_base_asset, -wallet_deltas.balance.amount)
+        self.assertEqual(float(market_deltas.d_base_asset), float(-wallet_deltas.balance.amount))
 
         # Ensure that the new LP receives the same amount of LP shares as the initializer.
-        self.assertAlmostEqual(market_deltas.d_lp_total_supply, lp_supply_before, 6)
-        self.assertEqual(self.hyperdrive.market_state.lp_total_supply, lp_supply_before * 2)
+        self.assertAlmostEqual(float(market_deltas.d_lp_total_supply), float(lp_supply_before), places=6)
```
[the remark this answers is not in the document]
dpaiton: you mean using `int` & `delta` instead of `float` and `places`? I guess they are interchangeable. `FixedPoint(48)` and `FixedPoint(49)` have delta of 1, and places would be 18. If it is larger, e.g. `FixedPoint(948)` and `FixedPoint(949)`, it also has delta of 1 and places of 18. Right? I don't care either way. I think `AlmostEqual` is using `round` under the hood so it's losing any float noise anyways? I think that's what you're saying when you say they get converted to the same float.

tests_fp/solidity/test_add_liquidity_fp.py
dpaiton: maxed out precision wherever I could. Still not great

PR #353, opened 2023-04-21
adds infinite and nan support for FixedPoint

dpaiton:


elfpy/utils/math/fixed_point.py
```diff
@@ -48,24 +55,26 @@ def _coerce_other(self, other):
...
-
     def __add__(self, other: int | FixedPoint) -> FixedPoint:
         """Enables '+' syntax"""
         other = self._coerce_other(other)
         if other is NotImplemented:
             return NotImplemented
+        if self.is_nan() or other.is_nan():
+            return FixedPoint("nan")
+        if self.is_inf():
+            if other.is_inf() and self.sign() != other.sign():
```
wakamex, line 73: inf * -inf = nan? TIL

elfpy/utils/math/fixed_point.py
```diff
@@ -48,24 +55,26 @@ def _coerce_other(self, other):
...
+                return FixedPoint("nan")
+            return self
+        if other.is_inf():
```
wakamex, line 76: why not combine put this in compound statement above? `if self.is_inf() or other.is_inf()` assuming inf's are fungible

elfpy/utils/math/fixed_point.py
```diff
@@ -77,20 +86,44 @@ def __sub__(self, other: int | FixedPoint) -> FixedPoint:
         other = self._coerce_other(other)
         if other is NotImplemented:
             return NotImplemented
+        if self.is_nan() or other.is_nan():
```
wakamex, line 89: they're fungible here!

elfpy/utils/math/fixed_point.py
```diff
@@ -77,20 +86,44 @@ def __sub__(self, other: int | FixedPoint) -> FixedPoint:
...
+            return FixedPoint("nan")
+        if self.is_inf():
+            if other.is_inf() and self.sign() == other.sign():
```
wakamex, line 92: here `inf * inf = nan` ?

elfpy/utils/math/fixed_point.py
```diff
@@ -77,20 +86,44 @@ def __sub__(self, other: int | FixedPoint) -> FixedPoint:
...
     def __mul__(self, other: int | FixedPoint) -> FixedPoint:
         """Enables '*' syntax"""
         other = self._coerce_other(other)
         if other is NotImplemented:
             return NotImplemented
+        if self.is_nan() or other.is_nan():
+            return FixedPoint("nan")
+        if self.is_zero() or other.is_zero():
+            if self.is_inf() or other.is_inf():
+                return FixedPoint("nan")  # zero * inf is nan
```
wakamex, line 123: I like this comment

elfpy/utils/math/fixed_point.py
```diff
@@ -116,16 +149,27 @@ def __floordiv__(self, other: int | FixedPoint) -> FixedPoint:
         other = self._coerce_other(other)
         if other is NotImplemented:
             return NotImplemented
-        if other <= FixedPoint("0.0"):
+        if other == FixedPoint("0.0"):
             raise errors.DivisionByZero
+        if self.is_nan() or other.is_nan():
+            return FixedPoint("nan")
+        if self.is_inf():
+            if other.is_inf():  # inf / inf is zero
```
wakamex, line 157: should read is nan right?

elfpy/utils/math/fixed_point.py
```diff
@@ -116,16 +149,27 @@ def __floordiv__(self, other: int | FixedPoint) -> FixedPoint:
         other = self._coerce_other(other)
         if other is NotImplemented:
             return NotImplemented
-        if other <= FixedPoint("0.0"):
```
wakamex, line 119: was this an error before? should we have tests to catch it?

wakamex (approved):
looks good, the tests seem very thorough!

elfpy/utils/math/fixed_point.py
dpaiton: Not quite! It's easiest to see the rules if you look at the test. In the `test_fixed_point_nonfinite.py` file you can see: `assert self.NEG_INF * self.INF == self.NEG_INF`

elfpy/utils/math/fixed_point.py
dpaiton: I'll push up some more comments to make this clearer.

elfpy/utils/math/fixed_point.py
dpaiton: yes good catch!

elfpy/utils/math/fixed_point.py
dpaiton: I think it was checking <0 before because we copied the solidity contracts which generally use unsigned ints. But we're living in a signed int world so we don't care if it is negative.

PR #357, opened 2023-04-24
Convert Policies to Fixed Point

sentilesdal:
- copy agents.  only signature/names changed
- convert to fixed point

### DIFF
--- a/elfpy/agents/policies/lp_and_withdraw.py
+++ b/elfpy/agents/policies/lp_and_withdraw.py
@@ -1,8 +1,11 @@
 """User strategy that adds liquidity and then removes it when enough time has passed"""
+from typing import List
+
 import elfpy.agents.agent as agent
 import elfpy.markets.hyperdrive.hyperdrive_actions as hyperdrive_actions
 import elfpy.markets.hyperdrive.hyperdrive_market as hyperdrive_market
 import elfpy.types as types
+from elfpy.utils.math import FixedPoint
 
 # pylint: disable=duplicate-code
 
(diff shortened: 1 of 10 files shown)


elfpy/agents/policies/lp_and_withdraw.py
```diff
@@ -54,3 +57,52 @@ def action(self, market: hyperdrive_market.Market) -> "list[types.Trade]":
...
         return action_list
+
+
+class SingleLPAgent(agent.AgentFP):
+    """
+    simple LP
+    only has one LP open at a time
+    """
+
+    def __init__(self, wallet_address, budget=1000):
```
dpaiton, line 68: add FixedPoint type & cast

elfpy/agents/policies/random_agent.py
```diff
@@ -156,7 +157,191 @@ def close_random_long(self) -> list[types.Trade]:
             )
         ]
 
-    def action(self, market: hyperdrive_market.Market) -> list[types.Trade]:
+    def action(self, market: hyperdrive_market.MarketFP) -> list[types.Trade]:
```
dpaiton, line 160: wrong action func

elfpy/agents/policies/single_lp.py
```diff
@@ -38,3 +39,33 @@ def action(self, _market: hyperdrive_market.Market) -> "list[types.Trade]":
...
+
+class SingleLpAgent(agent.AgentFP):
+    """simple LP that only has one LP open at a time"""
+
+    def __init__(self, wallet_address, budget=FixedPoint("1000.0")):
+        """call basic policy init then add custom stuff"""
+        self.amount_to_lp = FixedPoint("100.0")
+        super().__init__(wallet_address, budget)
+
+    def action(self, _market: hyperdrive_market.Market) -> "list[types.Trade]":
```
dpaiton, line 52: if you add `from __future__ import annotations` then you can remove the quotes around the return type.

dpaiton (approved):

elfpy/agents/policies/lp_and_withdraw.py
```diff
@@ -54,3 +57,52 @@ def action(self, market: hyperdrive_market.Market) -> "list[types.Trade]":
...
+    only has one LP open at a time
+    """
+
+    def __init__(self, wallet_address, budget: FixedPoint = FixedPoint("1000.0")):
+        """call basic policy init then add custom stuff"""
+        self.time_to_withdraw = FixedPoint("1.0")
+        self.amount_to_lp = FixedPoint("100.0")
+        super().__init__(wallet_address, budget)
+
+    def action(self, market: hyperdrive_market.MarketFP) -> List[types.Trade]:
```
dpaiton, line 74: `List` with a capital L is being phased out (as are similar types like `Dict`, `Tuple`, etc). The correct behavior is to add `from __future__ import annotations` and then use lower-case `list`. Then, when we move to 3.11+, we can remove the import and everything will work.

dpaiton (approved):

PR #361, opened 2023-05-02
fixes bug with open short calculation

dpaiton:


elfpy/markets/hyperdrive/hyperdrive_actions.py
```diff
@@ -1413,10 +1413,14 @@ def calc_open_short_fp(
         Returns the deltas to update the market and the agent's wallet after opening a short.
     """
     # get the checkpointed time remaining
-    mint_time = latest_checkpoint_time
-    days_remaining = position_duration.days - (block_time - latest_checkpoint_time) * FixedPoint("365.0")
+    annualized_position_duration = position_duration.days / FixedPoint("365.0")
```
sentilesdal, line 1416: Ok I'm confused why this is broken.  Hopefully we can avoid going from days to years back to days.  I rewrote this as:
```python
mint_time = latest_checkpoint_time
days_elapsed = (block_time - mint_time) * FixedPoint("365.0")
days_remaining = position_duration.days - days_elapsed
```
which looks right to me...
Going to take a look.

dpaiton: I agree that converting back and forth sucks, but I figured we just have to throw it on the list of stuff to fix when we fix time. `get_years_remaining` does includes some checks and a call to `maximum` that is missing in your alternative. Also I think the order of operations in the original formula was actually just wrong. `days_remaining` is `(position_duration_years - (market_time - mint_time)) * 365`

PR #364, opened 2023-05-02
Convert Simulator to FixedPoint

sentilesdal:
Converts the simulator class and dependencies over to FixedPoint arithmetic.  This is the minimum viable conversion to try to get the hyperdrive notebook to run.  I created a fixed point version of the notebook and also converted it to a python script.  It still fails because we getting slightly negative values for reserves and trying to raise them to a power.  After considerable bug hunting I decided to cut this off here and then try to pursue the fixes to the rounding problems that have been plaguing us in a separate PR.  The commits definitely tell a story, please use them as there are several commits that just copy things that you can ignore:

- copy things
- minimum conversion
- copy things
- convert utils
- copy file
- fix bug
- convert to script file
- copy things
- convert post_processing
- fix slew of bugs
- better error message
- fix bug
- update example to fixed point


elfpy/utils/math/fixed_point.py
```diff
@@ -481,7 +481,8 @@ def ln(x: int) -> int:
         Reverts if value is negative or 0.
         """
         if x <= 0:
-            raise ValueError("FixedPointMath_NegativeOrZeroInput")
+            print(f"{x=}")
```
dpaiton, line 484: No prints! I also fixed all of the fixed point error messages in one of my prs, I can add the value as well in that one as well. So can you revert these changes in yours?

examples/notebooks/hyperdrive_fp.ipynb
```diff
@@ -0,0 +1,546 @@
+{
```
dpaiton, line 1: are you intending for us to maintain the `ipynb` and `py` file? Or should the `ipynb` be deleted?

dpaiton (approved):
minor comments, otherwise looks great! Thanks a ton, huge effort

PR #369, opened 2023-05-03
Fixed Point Tests - Close Long and Close Short

sentilesdal:
Converts test_close_long and test_close_short to use fixed point math


tests_fp/solidity/test_close_long.py
```diff
@@ -217,22 +219,23 @@ def test_close_long_immediately_with_regular_amount(self):
...
             agent_base_paid=base_amount,
             agent_base_proceeds=agent_deltas_close.balance.amount,
             bond_amount=agent_deltas_open.longs[0].balance,
-            maturity_time=self.hyperdrive.position_duration.days / 365,
+            maturity_time=self.hyperdrive.position_duration.days / FixedPoint("365.0"),
         )
 
     def test_close_long_immediately_with_small_amount(self):
         """Open a small position, close it, and then verify that the close long updates were correct"""
         base_amount = 0.01
```
dpaiton: delete old line

tests_fp/solidity/test_close_short.py
```diff
@@ -141,61 +143,61 @@ def verify_close_short(
...
         self.assertEqual(  # short average maturity time
             self.hyperdrive.market_state.short_average_maturity_time,
-            0,
+            FixedPoint(0),
             msg="short_average_maturity_time is wrong",
         )
         self.assertAlmostEqual(  # short base volume
-            self.hyperdrive.market_state.short_base_volume,
+            int(self.hyperdrive.market_state.short_base_volume),
             0,
```
dpaiton: is `0` the argument for places? not sure what this is doing.

tests_fp/solidity/test_close_short.py
```diff
@@ -76,32 +77,33 @@ def verify_close_short(
...
+            market_state_before.bond_reserves,
+            market_state_before.lp_total_supply,
+            bond_amount * time_remaining,
+            FixedPoint("1.0") - stretch_time.stretched_time,
+            market_state_before.share_price,
+            market_state_before.init_share_price,
+        )
+
+        flat_shares = bond_amount * (FixedPoint("1.0") - time_remaining) / market_state_before.share_price
+        np.testing.assert_allclose(
```
dpaiton: I think we should remove all `np.testing` calls and replacing with `self.assertAlmostEqual` to stay consistent with the `unittest` stdlib. I've been doing this in my FixedPoint refactored tests.

dpaiton (approved):

tests_fp/solidity/test_close_short.py
sentilesdal: i think delta is better with ints because we know exactly how much it is off.  places can be dynamic with floats and doesn't work with int/fp

PR #373, opened 2023-05-04
adds fp freezable, logging, config, time tests

dpaiton:


tests_fp/test_freezable_fp.py
```diff
@@ -0,0 +1,164 @@
+"""Testing for ElfPy types"""
+from __future__ import annotations
+
+import unittest
+from dataclasses import dataclass
+
+from elfpy.types import freezable
```
wakamex, line 7: is this testing what you want? I see there's only 1 freezable class, but the only part that explicitly uses FixedPoint is the QuantityFP object, which I don't see being used here.

wakamex (approved):
lgtm, assume the tests are 1:1 copies of existing non-FP tests. is that right? mabye that's why there isn't much FP-specific stuff in them?

PR #379, opened 2023-05-05
add github token to CI flows

wakamex:
tell the `actions/checkout@v3` step to use the default assigned github token (auto-generated just for this CI run) which lets it avoid running into API rate limits from using unauthenticated API calls

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,3 +60,4 @@ jobs:
         with:
           flags: unittests
           fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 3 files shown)


sentilesdal (approved):

.github/workflows/test.yml
```diff
@@ -58,3 +60,4 @@ jobs:
         with:
           flags: unittests
           fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
```
wakamex, line 63: when in doubt, add more tokens

wakamex (commented):
lgtm, nice work ✅

PR #386, opened 2023-05-11
integrated bots

wakamex:


elfpy/utils/apeworx_integrations.py
```diff
@@ -198,37 +195,55 @@ def get_wallet_from_onchain_trade_info(
...
+                    value *= -1 if info.trades.loc[specific_trade, "from"] == address_ else 1
+                    sum_value += value
+                    sum_product_of_open_share_price_and_value += (
+                        value * info.share_price[info.trades.loc[specific_trade, "block_number"]]
+                    )
+                open_share_price = sum_product_of_open_share_price_and_value / sum_value
+                assert (
+                    abs(balance - sum_value) <= elfpy.MAXIMUM_BALANCE_MISMATCH_IN_WEI
+                ), "weighted average open share price calculation is wrong"
+                print(f"calculated weighted average open share price of {open_share_price}")
```
dpaiton, line 237: no prints

elfpy/utils/apeworx_integrations.py
```diff
@@ -198,37 +195,55 @@ def get_wallet_from_onchain_trade_info(
...
+                new_record = {mint_time: Short(balance=balance, open_share_price=open_share_price)}
+                wallet.shorts |= new_record
+                print(f"storing in wallet as {new_record=}")
+            elif asset_type == "LONG":
+                new_record = {mint_time: Long(balance=balance)}
+                wallet.longs |= new_record
```
dpaiton, line 243: `|=` performs different operations for different types. In this case, it's a dictionary update. But a new programmer looking over the code might not recall that `wallet.longs` is a dictionary. For this reason, I think it is better to use the more explicit `update` syntax.

elfpy/utils/apeworx_integrations.py
```diff
@@ -317,8 +334,7 @@ def get_transfer_single_event(tx_receipt: ReceiptAPI) -> ContractLog:
 
 
 def get_pool_state(tx_receipt: ReceiptAPI, hyperdrive_contract: ContractInstance):
-    """
-    Return everything returned by `getPoolInfo()` in the smart contracts.
+    """Return everything returned by `getPoolInfo()` in the smart contracts.
```
dpaiton, line 337: I _think_ you need the `r"""` syntax when using backticks to delineate code.

elfpy/utils/apeworx_integrations.py
```diff
@@ -707,43 +724,56 @@ def attempt_txn(
...
+        formatted_items = []
+        for k, v in kwargs.items():
+            value = fmt(v / 1e9) if "fee" in k else fmt(v)
+            formatted_items.append(f"{k}={value}")
+        log_and_show(f"txn attempt {attempt} of {mult} with {', '.join(formatted_items)}")
         serial_txn: TransactionAPI = contract_txn.serialize_transaction(*args, **kwargs)
         prepped_txn: TransactionAPI = agent.prepare_transaction(serial_txn)
         signed_txn: Optional[TransactionAPI] = agent.sign_transaction(prepped_txn)
-        log_and_show(f" => sending {signed_txn=}")
+        logging.debug(" => sending signed_txn %s", signed_txn)
```
dpaiton, line 763: What is the reasoning between when you want to "log and show" and when you want to use normal logging? I'm wondering if we should just set up a logging config parameter that supports the "log and show" method. The way you have it now, it is up to the programmer to be opinionated about when it is appropriate to do one or the other, which I don't think is the right model with logging. Let's discuss later and make an issue for it if we agree.

examples/testy_bots.py
```diff
@@ -238,53 +263,85 @@ def get_argparser() -> argparse.ArgumentParser:
         default=0.1,
         type=float,
     )
+
+    parser.add_argument("--devnet", help="Run on devnet", action="store_false")  # store_false because default is True
     return parser
 
 
+@dataclass
```
dpaiton, line 271: This is awesome!

examples/testy_bots.py
```diff
@@ -238,53 +263,85 @@ def get_argparser() -> argparse.ArgumentParser:
...
-    _config.scratch["frida_risk_mean"] = 0.02
-    _config.scratch["frida_risk_std"] = 0.01
-    _config.scratch["random_budget_mean"] = 5_000
-    _config.scratch["random_budget_std"] = 2_000
-    _config.scratch["random_budget_max"] = 10_000
-    _config.scratch["random_budget_min"] = 1_000
-    _config.scratch["bot_types"] = {"louie": LongLouie, "frida": FixedFrida, "random": random_agent.Policy}
+    _config.log_filename += "_devnet" if args.devnet else ""
+    trade_chance = _config.scratch["trade_chance"]
+    _config.scratch["louie"] = BotInfo(risk_threshold=0.0, policy=LongLouie, trade_chance=trade_chance)
```
dpaiton, line 331: much better

dpaiton (approved):
This looks good to me. I added a few comments for small changes and one thing we should discuss outside of this PR (logging vs log_and_show). I only gave testy_bots a cursory review because I'll need to look at it more closely anyway to convert it to use FixedPoint.

elfpy/utils/apeworx_integrations.py
wakamex: ya apparently its behavior on sets is quite different. it was an operator there before they added it to dicts.

elfpy/utils/apeworx_integrations.py
wakamex: oh right ya that sounds right. I think I was reading these as single quotes for a while.

elfpy/utils/apeworx_integrations.py
wakamex: it's just my ongoing obsession with seeing things on my console, and hating using log files.

don't want to create overhead for the coder. I tried to make this purely optional, by adding an extra handler to our existing logging setup. but committing it here feels like everyone reviewing this code is going to wonder when to use `log_and_show` and when to use regular logging. so I'm kinda imposing my preference on them, through that. feels like there should be a better way. maybe a... "print every error to stdout" as an additional toggle in our logging?

wakamex: I guess the current thinking is "do I want to see this on every run, or only when debugging?"

PR #416, opened 2023-05-13
add get_wallet_from_onchain_trade_info to ape_utils

wakamex:
I loop across all the positions as captured by the `unique_ids` in the `OnChainTradeInfo` input.
For each position:
- calculate the balance across all trades for that position, and compare it to a `balanceOf` query.
- if there's an outstanding balance, create a wallet entry for each trade that comprises it

fixes: #419

### DIFF
--- a/elfpy/utils/apeworx_integrations.py
+++ b/elfpy/utils/apeworx_integrations.py
@@ -17,11 +17,13 @@ from ape.contracts.base import ContractTransaction, ContractTransactionHandler
 import numpy as np
 import pandas as pd
 
+import elfpy
+from elfpy import types
 from elfpy.types import freezable
-
 from elfpy.markets.hyperdrive import hyperdrive_assets, hyperdrive_market
 from elfpy.utils.outputs import number_to_string as fmt
 from elfpy.utils.outputs import log_and_show
+from elfpy.agents.wallet import Long, Short, Wallet
 from elfpy.math import FixedPoint
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


elfpy/utils/apeworx_integrations.py
```diff
@@ -31,6 +33,28 @@
     from ethpm_types.abi import MethodABI
 
 
+class HyperdriveProject(ProjectManager):
```
dpaiton, line 36: wasn't this in another PR already?

elfpy/utils/apeworx_integrations.py
```diff
@@ -139,7 +161,81 @@ def get_on_chain_trade_info(hyperdrive: ContractInstance) -> OnChainTradeInfo:
...
+        Index of the wallet.
+    on_chain_trade_info: OnChainTradeInfo
+        On-chain trade info.
+
+    Returns
+    -------
+    Wallet
+        Wallet with Short, Long, and LP positions.
+    """
+
```
dpaiton, line 199: can you remove these extra newlines?

elfpy/utils/apeworx_integrations.py
```diff
@@ -139,7 +161,81 @@ def get_on_chain_trade_info(hyperdrive: ContractInstance) -> OnChainTradeInfo:
     for block_number_, price in share_price_.items():
         logging.debug(("block_number_={}, price={}", block_number_, price))
 
-    return OnChainTradeInfo(hyper_trades_, unique_maturities_, unique_ids_, unique_block_numbers_, share_price_)
+    return OnChainTradeInfo(trades, unique_maturities_, unique_ids_, unique_block_numbers_, share_price_)
+
+
+def log_subtotals(balance, query_balance, hyper_trades, idx, asset_type):
```
dpaiton, line 167: I don't think you really need this function, since it's only used once.

dpaiton (approved):

elfpy/utils/apeworx_integrations.py
wakamex: rebase accident

elfpy/utils/apeworx_integrations.py
wakamex: I only added it to get below the local variable limit per function. Let me see if I can get rid of it.

wakamex: I just got rid of it, since it was only used for debugging. To debug in the future, we can revive it from this PR. Or just re-do it, since it was just some fancy print statements.

Issue #417, opened 2023-05-13
add gas helpers in ape_utils


wakamex:
get basic information about gas in a block, like a list of gas fees per transaction, and their statistics.

PR #418, opened 2023-05-13
add get_gas_fees

wakamex:
facilitate looping across type2 transactions only and retrieve maximum and average values for both max_gas and priority_fee. this is used only for information purposes in `testy_bots`, but elsewhere I used it to ensure we're not overspending.

fixes: #417

### DIFF
--- a/elfpy/utils/apeworx_integrations.py
+++ b/elfpy/utils/apeworx_integrations.py
@@ -142,8 +142,36 @@ def get_on_chain_trade_info(hyperdrive: ContractInstance) -> OnChainTradeInfo:
     return OnChainTradeInfo(hyper_trades_, unique_maturities_, unique_ids_, unique_block_numbers_, share_price_)
 
 
-def get_gas_fees(block: BlockAPI) -> tuple[float, float, float, float]:
-    """Get the max and avg max and priority fees from a block.
+def get_gas_fees(block: BlockAPI) -> tuple[list[float], list[float]]:
+    """Get the max and priority fees from a block (type 2 transactions only).
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


elfpy/utils/apeworx_integrations.py
```diff
@@ -142,6 +142,30 @@ def get_on_chain_trade_info(hyperdrive: ContractInstance) -> OnChainTradeInfo:
...
+    ----------
+    block: `ape.eth2.BlockAPI <https://docs.apeworx.io/ape/stable/methoddocs/api.html#ape.api.providers.BlockAPI>`_
+        Block to get gas fees from.
+
+    Returns
+    -------
+    tuple[float, float, float, float]
+        Tuple containing the max and avg max and priority fees.
+    """
+    if type2 := [txn for txn in block.transactions if txn.type == 2]:  # noqa: PLR2004
```
dpaiton, line 158: Can you add a comment explaining what this line is doing?

dpaiton: Actually comments explaining the rest of these lines as well would be great

elfpy/utils/apeworx_integrations.py
```diff
@@ -142,6 +142,30 @@ def get_on_chain_trade_info(hyperdrive: ContractInstance) -> OnChainTradeInfo:
...
+        max_fees, priority_fees = zip(*((txn.max_fee, txn.max_priority_fee) for txn in type2))
+        max_fees = [f / 1e9 for f in max_fees if f is not None]
```
dpaiton, line 160: can you not use single character names? `fee` I think is what you mean?

dpaiton (approved):
this looks like a useful function, good idea. Add some comments explaining what is going on & improve the variable names then it's good to go

PR #422, opened 2023-05-13
fix codecov badge in README.md

wakamex:
point to delvtech


dpaiton (approved):
Thanks for doing this! It looks like the README icon is still not working in your branch? maybe it'll work once it's merged. Check once it's merged.

PR #436, opened 2023-05-16
bugfix in freezable.astype, import rename, remove example_main

dpaiton:


sentilesdal (approved):
approved with fix to astypes function in types.py

PR #438, opened 2023-05-16
remove `check_action` and improves `check_non_zero`

dpaiton:


wakamex:
can we use the package-level `elfpy.check_non_zero()`? I don't see why we'd want duplicates of this function

PR #439, opened 2023-05-17
fix bug with failed trade modifying market state

dpaiton:


dpaiton:
> can we use the package-level `elfpy.check_non_zero()`? I don't see why we'd want duplicates of this function

There was a bug with that one due to minor differences between the wallet and market state, which was fixed in an earlier PR. The fixed moved us over to a different version of `check_non_zero` in wallet and market. However, now that I'm looking, it seems like those two versions have already diverged, which is a good argument for keeping it in one place. I'll make the switch.

dpaiton:
> can we use the package-level `elfpy.check_non_zero()`? I don't see why we'd want duplicates of this function

There was a bug with the original `elfpy.check_non_zero()` due to minor differences between the wallet and market state, which was fixed in an earlier PR by making two separate functions. However, now that I'm looking, it seems like those two versions have already diverged, which is a good argument for keeping at least part of the check in one place. The recursive nature of it makes this hard. If it lived in `elfpy.__init__.py` then we would need to import a lot of types (e.g. right now `Checkpoint` from markets and `Quantity` from types). This is a bad pattern; we do not want to be importing a bunch of packages into elfpy core. I think for now we should keep them separate and create an issue for unifying the checks. I'll push up an example.

wakamex:
sounds like that's in the direction of the "more localized checks" mentioned here https://github.com/delvtech/elf-simulations/blob/738f4f53ab0782387ba1232344bc615ed1b3fd7b/elfpy/__init__.py#L201
```py
    # TODO: issue #146
    # this is an imperfect solution to rounding errors
    # and should be removed in favor of localized checks
```

when I created that check in elfpy, I think we had two identical functions. now we're going forward with more specific function-level checks. sounds good.

elfpy/markets/hyperdrive/hyperdrive_actions.py
```diff
@@ -2197,3 +2153,19 @@ def calc_free_margin_fp(
         withdraw_pool_deltas.withdraw_interest = interest
     # Finally return the amount used by this action and the caller can update reserves.
     return withdraw_pool_deltas
+
+
+def check_output_assertions(trade_result: trades.TradeResultFP):
```
wakamex, line 2158: this is basically `check_output_assertions_FP` right? why not put it in base pricing model next to the existing `check_output_assertions`? once we switch over, you'll have to track it down anyway.

wakamex (approved):
lgtm

elfpy/markets/hyperdrive/hyperdrive_actions.py
dpaiton: We are no longer going to call it from the pricing model. Instead it'll be a local function available from hyperdrive_actions (i.e. doesn't belong to any class). So the version in the pricing model will be deleted once I purge float from the repo.

Issue #426, opened 2023-05-14
Add components to FixedPoint type so that it conforms to `numbers.Real`


dpaiton:
FixedPoint type should include the necessary dunders to make it equivalent to a numbers.Real type. This will allow us to use e.g. assertAlmostEqual with FixedPoint arguments.

dpaiton:
Resolved by #446

PR #446, opened 2023-05-18
FixPoint conforms to python `numbers.Real` specification

dpaiton:
addresses #426

### DIFF
--- a/elfpy/math/fixed_point.py
+++ b/elfpy/math/fixed_point.py
@@ -154,18 +154,10 @@ class FixedPoint:
         return self * other
 
     def __truediv__(self, other):
-        """Enables '/' syntax, which mirrors `//` syntax.
+        """Enables '/' syntax.
 
         We mirror floordiv because most solidity contract equations use divdown
         """
-        return self.__floordiv__(other)
-
-    def __rtruediv__(self, other):
-        """Enables reciprocal division to support other / FixedPoint"""
-        return self.__truediv__(other)
-
-    def __floordiv__(self, other: int | FixedPoint) -> FixedPoint:
-        """Enables '//' syntax"""
         other = self._coerce_other(other)
         if other is NotImplemented:
             return NotImplemented
@@ -182,6 +174,18 @@ class FixedPoint:
         return FixedPoint(
             FixedPointIntegerMath.div_down(self.int_value, other.int_value), self.decimal_places, self.signed
         )
+
+    def __rtruediv__(self, other):
+        """Enables reciprocal division to support other / FixedPoint"""
+        return self.__truediv__(other)
+
+    def __floordiv__(self, other: int | FixedPoint) -> FixedPoint:
+        """Enables '//' syntax
+
+        While FixedPoint numbers are represented as integers, they act like floats.
+        So floordiv should return only whole numbers.
+        """
+        return (self.__truediv__(other)).__floor__()
 
     def __pow__(self, other: int | FixedPoint) -> FixedPoint:
         """Enables '**' syntax"""
@@ -241,10 +245,14 @@ class FixedPoint:
         return FixedPoint("-1.0") * self
 
     def __abs__(self) -> FixedPoint:
-        """Enables 'abs()' function"""
+        r"""Enables 'abs()' function"""
         if self.is_nan():
             return self
         return FixedPoint(abs(self.int_value), self.decimal_places, self.signed)
+
+    def __divmod__(self, other: FixedPoint) -> tuple[FixedPoint, FixedPoint]:
+        r"""Enables `divmod()` function"""
+        return (self // other, self % other)
 
     # comparison methods
     def __eq__(self, other: FixedPoint) -> bool:
--- a/elfpy/math/fixed_point.py
+++ b/elfpy/math/fixed_point.py
@@ -193,7 +193,7 @@ class FixedPoint:
     def __rpow__(self, other: int | FixedPoint) -> FixedPoint:
         r"""Enables reciprocal pow to support other ** FixedPoint"""
         other = self._coerce_other(other)
-        return other ** self
+        return other**self
 
     def __mod__(self, other: FixedPoint) -> FixedPoint:
         r"""Enables `%` syntax
--- a/elfpy/math/fixed_point.py
+++ b/elfpy/math/fixed_point.py
@@ -9,7 +9,7 @@ from .fixed_point_integer_math import FixedPointIntegerMath
 
 
 class FixedPoint:
-    """Fixed-point number datatype
+    r"""Fixed-point number datatype
 
     Values are stored internally as intergers, however they are generally treated like floats.
     The first (right-most) `decimal_places` digits represent what would be to the right of
@@ -32,7 +32,7 @@ class FixedPoint:
     int_value: int  # integer representation of self
 
     def __init__(self, value: Union[FixedPoint, float, int, str] = 0, decimal_places: int = 18, signed: bool = True):
-        """Store fixed-point properties"""
+        r"""Store fixed-point properties"""
         # TODO: support unsigned option
         if not signed:
             raise NotImplementedError("only signed FixedPoint ints are supported.")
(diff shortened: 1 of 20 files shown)


dpaiton:
I have figured out the static check issue, it has to do with the allowable type signatures in `https://github.com/python/typeshed/blob/main/stdlib/unittest/case.pyi#L192`. It's a little confusing still -- I don't know why FixedPoint is not passing the `SupportsRound`, `SupportsAbs`, or `SupportsRSub` checks. I'll file a ticket to resolve that later. Whenever the check does not raise an error, it is because the type is not known (e.g. `expected_result.without_fee_or_slippage` in `test_calc_out_given_in_fp.py` has type `Any`, which is honestly also a bug). The solution is to use `delta` instead of `places`. Then a different type signature is used, which has different requirements. This makes sense anyways, since `places` refers to decimal places, which `unittest` might assume is always passing for FixedPoint. (if it doesn't then it still doesn't make sense).

TL;DR: changing `places` to `delta` will fix the static errors. I'll do this before merging.

tests/fixed_point/test_fixed_point.py
```diff
@@ -322,10 +322,29 @@ def test_modulo_fail(self):
...
+        assert math.floor(FixedPoint("-2.1")) == FixedPoint("-3.0")
+        assert math.floor(FixedPoint("3.6")) == FixedPoint("3.0")
+        assert math.floor(FixedPoint("0.5")) == FixedPoint(0)
+        assert math.floor(FixedPoint(3)) == FixedPoint(0)
+        assert math.floor(FixedPoint(-6)) == FixedPoint("-1.0")
+        assert math.floor(FixedPoint(-6.0)) == FixedPoint(-6.0)
+        assert math.floor(FixedPoint(-6.8)) == FixedPoint(-7.0)
+
+    def test_ceil(self):
+        r"""Test floor operator"""
```
sentilesdal: Test ceil operator

dpaiton: Fixed in a later commit

elfpy/math/fixed_point.py
```diff
@@ -333,14 +333,24 @@ def __ge__(self, other: FixedPoint) -> bool:
...
+
     def __floor__(self) -> FixedPoint:
         """Returns an integer rounded following Python `math.floor` behavior
 
         Given a real number x, return as output the greatest integer less than or equal to x.
         """
         if not self.is_finite():
             return self
-        lhs, rhs = str(self).split(".")
+        lhs, rhs = str(self).split(".")  # lhs = integer part, rhs = fractional part
```
sentilesdal, line 334: could just call the variables themselves `integer, remainder`

elfpy/math/fixed_point.py
```diff
@@ -353,7 +363,7 @@ def __ceil__(self) -> FixedPoint:
...
```
sentilesdal, line 347: could just call the variables themselves integer, remainder

elfpy/math/fixed_point.py
```diff
@@ -372,6 +372,45 @@ def __ceil__(self) -> FixedPoint:
...
 
+    def __round__(self, n: int = 0) -> FixedPoint:
+        """Returns a number rounded following Python `round` behavior.
+
+        Given a real number x and an optional integer n, return as output the number rounded to the closest multiple of 10 to the power -n.
+        If n is omitted, it defaults to 0 (round to nearest integer). Uses Python's "round half to even" strategy.
+        """
+        if not self.is_finite():
+            return self
+        lhs, rhs = str(self).split(".")  # lhs = integer part, rhs = fractional part
```
sentilesdal, line 366: could just call the variables themselves `integer, remainder`

elfpy/math/fixed_point.py
```diff
@@ -372,6 +372,45 @@ def __ceil__(self) -> FixedPoint:
...
+        digit = int(rhs[n])
+        if n == 0 or len(rhs) < n:
+            left_digit = int(lhs[-1])
+        else:
+            left_digit = int(rhs[n - 1])
+        # If these conditions are met, we should round down
+        if digit < 5 or (  # digit less than 5 OR
+            digit == 5  # digit is exactly 5 AND
+            and all(d == "0" for d in rhs[n + 1 :])  # all of the following digits are zero AND
+            and (left_digit % 2 == 0)  # the digit to the left is even
```
sentilesdal, line 380: why does this matter?  it seems crazy to me that you'd round 1.75 to 1.8 and 1.45 to 1.4.  Is that really the standard?

sentilesdal (approved):
amazing!

elfpy/math/fixed_point.py
dpaiton: Yeah that's how python does rounding (round to even) https://realpython.com/python-rounding/#pythons-built-in-round-function

dpaiton: we can add args later to implement other methods: https://realpython.com/python-rounding/#a-menagerie-of-methods

elfpy/math/fixed_point.py
```diff
@@ -183,6 +175,18 @@ def __floordiv__(self, other: int | FixedPoint) -> FixedPoint:
             FixedPointIntegerMath.div_down(self.int_value, other.int_value), self.decimal_places, self.signed
         )
 
+    def __rtruediv__(self, other):
+        """Enables reciprocal division to support other / FixedPoint"""
+        return self.__truediv__(other)
+
+    def __floordiv__(self, other: int | FixedPoint) -> FixedPoint:
```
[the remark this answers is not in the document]
dpaiton: Yeah, for truediv we will call divdown (so whenever we see divdown in solidity we would use `/` in python). And floordiv truncates to "integer", i.e. does divdown (rounds down with 18 decimals) and then we use `floor` to truncate the 18 decimals to just the whole part.

elfpy/math/fixed_point.py
```diff
@@ -241,11 +245,15 @@ def __neg__(self) -> FixedPoint:
         return FixedPoint("-1.0") * self
 
     def __abs__(self) -> FixedPoint:
-        """Enables 'abs()' function"""
+        r"""Enables 'abs()' function"""
```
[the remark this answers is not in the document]
dpaiton: I might be wrong here, but I think it will make a difference for the backticks to render as an inline code block. If the docstring has only text you don't need `r`, but if it has any sort of sphinx or markdown or whatever then we should use `r`. And since it doesn't really hurt to use it I think it's fine to just use it everywhere.

elfpy/math/fixed_point.py
```diff
@@ -150,7 +153,7 @@ def __rmul__(self, other: int | FixedPoint) -> FixedPoint:
     def __truediv__(self, other):
         """Enables '/' syntax.
 
-        We mirror floordiv because most solidity contract equations use divdown
+        We use div_down to match the majority of Hyperdrive solidity contract equations
```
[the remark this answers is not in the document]
dpaiton: We want to use the syntax sugar whenever possible. That being said, if a situation comes up where it really makes sense to call it `div_down` then I think it's ok to add it to `FixedPoint`. Notice we already have an example because we have an explicit `div_up`.

elfpy/math/fixed_point.py
```diff
@@ -387,44 +353,44 @@ def __ceil__(self) -> FixedPoint:
                 return FixedPoint(str(int(lhs) + 1) + ".0")  # increase integer component by one
         return FixedPoint(lhs + ".0")  # the number has no remainder
 
-    def __round__(self, n: int = 0) -> FixedPoint:  # pylint: disable=invalid-name
+    def __round__(self, ndigits: int = 0) -> FixedPoint:
```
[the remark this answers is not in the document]
dpaiton: When I first wrote it I followed python `float` type, which uses `n` in the `__round__` function. However, after looking around more, I found that a lot of other custom types use `ndigits`, which is what is used by the python [round](https://docs.python.org/3/library/functions.html#round) built-in function.

PR #447, opened 2023-05-22
build python-base docker image

wakamex:


Dockerfile-python-base
```diff
@@ -0,0 +1,18 @@
...
+COPY requirements* ./
+COPY ape-config.yaml.example ./ape-config.yaml
+
+# install base dependencies
+RUN python -m pip install --upgrade pip
+RUN python -m pip install -r requirements.txt
+RUN ape plugins install .
+
+# install dev dependencies
+RUN python -m pip install -r requirements-dev.txt
```
dpaiton, line 18: should this also install elfpy? `pip install -e .`?

requirements-dev.txt
```diff
@@ -10,4 +10,7 @@ autodocsumm>=0.2.10
 nbsphinx>=0.8.12
 sphinx-rtd-theme>=1.1.1
 # urllib3 v2 requires OpenSSL v1.1.1+, but Vercel uses v1.0.2
-urllib3<2.0
\ No newline at end of file
+urllib3<2.0
+jinja2
+ipywidgets
+tqdm
```
dpaiton, line 16: why do we need these new ones? I don't see where they're being used

requirements.txt
```diff
@@ -11,5 +11,8 @@ pyright
 pytest
 scipy
 tomli
+python-dotenv
 eth-ape
-python-dotenv
\ No newline at end of file
+ape-solidity
```
dpaiton, line 16: please double check that these do not create dependency issues when pip installing. Thanks!

requirements-dev.txt
dpaiton: As a rule of thumb let's try to only add requirements if it's absolutely necessary. So e.g. if `tqdm` is not actually needed by us then don't include it.

dpaiton (approved):
Left some comments for you to check out. As long as you're sure the requirements are actually required and that they don't create dependency errors in COLAB & locally (APE has done this in the past) then you're gtg. Thanks for this! Excited to get this running :)

Dockerfile-python-base
wakamex: I want it to be a clean python-only image, then we install elfpy on top of it in a separate workflow.

Initially I was testing with both steps in the same flow. But I realized the "install python" step is doing the same thing every time, and taking like 3 minutes to do it. So this avoids that, by only building it when necessary (changes to requirements).

Then we only build the things that actually change (elfpy code) on top of this.

requirements.txt
wakamex: It seems to completely replace the `ape plugins install .` step. Once we get there, ape just says "already installed" and skips past. Makes me wonder why both approaches exist, if they do the same thing.

wakamex: Putting them in `requirements.txt` gives you more flexibility, to pin to specific commits, or specifc branches (which I had just used before they upstreamed the recent bugfix). Which the `ape-config` parser doesn't allow. It does allow specific numbered versions.

wakamex: I'll set it back to using `ape-config` for now, but keepign in mind that we may want to use requirements.txt for exceptional situations

PR #450, opened 2023-05-22
Updating install documentation

slundqui:


INSTALL.md
```diff
@@ -7,22 +7,32 @@ Set up your favorite python virtual environment. We use:
...
 
+After installation, we can use pyenv to install Python from within the repo.
 ```bash
+cd <repo_location>
 pyenv install 3.9
 pyenv local 3.9
 python -m venv .venv
 source .venv/bin/activate
 ```
+`pyenv install 3.9` You should now see the correct version when you run `pyenv versions`.
```
dpaiton, line 22: Can you bullet these so they're clearly different from the code block above?

dpaiton (approved):
awesome, thanks! small request, then good.

PR #456, opened 2023-05-23
build elf-sims docker image

wakamex:
build docker image containing latest version of `elf-simulations` code, built on top of the existing `python-base` docker image.


Dockerfile-elf-sims
```diff
@@ -0,0 +1,35 @@
...
+# copy hyperdrive contracts from migrations image
+COPY --from=migrations /src/ ./hyperdrive_solidity/
+# copy foundry over from migrations image
+COPY --from=migrations /usr/local/bin/ /usr/local/bin
+
+# compile contracts
+RUN ape compile
+
+# install elf-simulations
+RUN python -m pip install -e .
```
dpaiton, line 28: why is it that you do not need to run pip install -r requirements.txt first?

dpaiton: ohh I see it happens in the other file

dpaiton (approved):
very cool! Can't wait to spin these up

Dockerfile-elf-sims
wakamex: pulled in freshly built and dockerized from `python-base`! measure twice, cut once!

PR #457, opened 2023-05-23
reorganize fixedpoint tests

dpaiton:


tests/fixed_point/test_fp_arithmetic_sugar.py
```diff
@@ -157,6 +66,35 @@ def test_add_fail(self):
         with self.assertRaises(TypeError):
             _ = int_value + fixed_point_value  # type: ignore
 
+    def test_add_nonfinite(self):
+        """Test rules for non-finite addition"""
+        # nan + anything is nan
+        assert (self.NAN + self.ZERO).is_nan() is True
```
slundqui, line 72: You don't need to explicitly test `is True`, the statement evaluates itself as true or false, which assert then passes or fails depending on the output. See rest of file for this as well.

tests/fixed_point/test_fp_arithmetic_sugar.py
```diff
@@ -157,6 +66,35 @@ def test_add_fail(self):
...
+        assert (self.ONE + self.NAN).is_nan() is True
+        assert (self.EVEN_FINITE + self.NAN).is_nan() is True
+        assert (self.NAN + self.NAN).is_nan() is True
+        assert (self.INF + self.NAN).is_nan() is True
+        assert (self.NEG_INF + self.NAN).is_nan() is True
+        # inf + -inf is nan
+        assert (self.INF + self.NEG_INF).is_nan() is True
+        assert (self.NEG_INF + self.INF).is_nan() is True
+        # inf + finite is inf
+        assert self.INF + self.EVEN_FINITE == self.INF
```
slundqui, line 89: There might be an order of operations issue here, one possible way this can be interpreted is

(self.EVEN_FINITE == self.INF) + self.INF is [truthy](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/)

See rest of file for this as well.

tests/fixed_point/test_fp_arithmetic_sugar.py
```diff
@@ -188,6 +126,34 @@ def test_sub_fail(self):
         with self.assertRaises(TypeError):
             _ = int_value - fixed_point_value  # type: ignore
 
+    def test_sub_nonfinite(self):
```
slundqui, line 129: I know this isn't part of the review, but on line 125 where you test for type errors, I would argue it's reasonable to internally convert integers to fixed points since there's no loss of precision (very similar to how native python supports float \op int). I see a point with not supporting float + fixed point due.

tests/fixed_point/test_fp_class.py
```diff
@@ -0,0 +1,215 @@
...
+        assert FixedPoint("50_000.0") == FixedPoint("50000.0")
+        assert FixedPoint("50_000.0") == FixedPoint("50000.0")
+        assert FixedPoint("5_340_070.0") == FixedPoint("5340070.0")
+        assert FixedPoint("-5") == FixedPoint(-5.0)
+        assert FixedPoint("5.1") == FixedPoint(5.1)
+        assert FixedPoint("5.01") == FixedPoint(5.01)
+        assert FixedPoint("5.000001") == FixedPoint(5.000001)
+        assert FixedPoint("5.000001000") == FixedPoint(5.000001)
+        # int == float
+        assert FixedPoint(5) == FixedPoint(5e-18)
```
slundqui, line 29: Why would this pass?

tests/fixed_point/test_fp_class.py
```diff
@@ -0,0 +1,215 @@
...
+        assert round(FixedPoint("3.54"), ndigits=1) == FixedPoint("3.5")
+        assert round(FixedPoint("3.545"), ndigits=2) == FixedPoint("3.54")
+        assert round(FixedPoint("3.545"), ndigits=4) == FixedPoint("3.545")
+        assert round(FixedPoint("3.545"), ndigits=5) == FixedPoint("3.545")
+        assert round(FixedPoint("-3.5459857"), ndigits=5) == FixedPoint("-3.54599")
+        assert round(FixedPoint("-3.5459850"), ndigits=5) == FixedPoint("-3.54598")
+
+    def test_hash(self):
+        """Test hash method"""
+        assert hash(FixedPoint(-1)) == -2
```
slundqui, line 202: Would a better test for this be trying to use fixed points as a dictionary key?

slundqui (approved):

tests/fixed_point/test_fp_arithmetic_sugar.py
dpaiton: If you inspect the `__bool__` [function](https://github.com/delvtech/elf-simulations/blob/4226789f0da084d4fc46dfca4f1b8f68a4669887/elfpy/math/fixed_point.py#L450) you can see how we evaluate to True or False. Between that and the `__eq__` [function](https://github.com/delvtech/elf-simulations/blob/4226789f0da084d4fc46dfca4f1b8f68a4669887/elfpy/math/fixed_point.py#L284)
I am able to convince myself that we are not getting any unexpected truthy values. 

However, just in case, I decided to add some more tests. Here is what I just tried out:
```
>>> EVEN_FINITE == INF
False
>>> (EVEN_FINITE == INF) + INF
FixedPoint(inf)
>>> bool((EVEN_FINITE == INF) + INF)
False
>>> INF + EVEN_FINITE
FixedPoint(inf)
>>> (INF + EVEN_FINITE) == INF
True
```

And here are the tests I attempted:
```
# verify that we aren't accidentally seeing a truthy value
assert not bool((self.EVEN_FINITE == self.INF) + self.INF)
assert not self.EVEN_FINITE == self.INF
```

However, answering your comment revealed that I missed something else, which is how `bool` variables should interact with `FixedPoint` variables. For example, with floats:
```
>>> my_val = 5.0
>>> False + my_val
5.0
>>> True + my_val
6.0
>>> float(False)
0.0
>>> float(True)
1.0
```

This behavior is the same with `int`. So, that brings up a familiar question: Do we replicate it, and if so which type do we replicate? i.e. does `FixedPoint(True) == FixedPoint(int(1))` or does `FixedPoint(True) == FixedPoint(float(1.0))`. This ambiguity is why we decided not to support `int + FixedPoint` or any variant (explained in another reply below). As such, I would recommend that the solution be to reject these types. The code was already almost there, but there is a bug. Here is the output:

```
>>> from elfpy.math import FixedPoint
>>> my_val = FixedPoint("2.0")
>>> True + my_val
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dylan/code/element/elf-simulations/elfpy/math/fixed_point.py", line 144, in __radd__
    other = self._coerce_other(other)
  File "/Users/dylan/code/element/elf-simulations/elfpy/math/fixed_point.py", line 124, in _coerce_other
    raise TypeError(f"unsupported operand type(s): {type(other)}")
TypeError: unsupported operand type(s): <class 'bool'>
>>> False + my_val
FixedPoint(2000000000000000000)
```

So `True + FixedPoint` throws an error, but `False + FixedPoint` does not. This is because of the checks in `FixedPoint._coerce_other(...)`, which never explicitly checks for `bool`. Instead it checks for `FixedPoint` as well as `(int, float)`. The latter of which will eval to True when it's actually a `bool`:
```
>>> isinstance(float, bool)
False
>>> isinstance(False, float)
True
```
So I added a check for this explicitly. Now the earlier test of `(self.EVEN_FINITE == self.INF) + self.INF` does not make sense as it throws an error for the invalid `True + FixedPoint`.

tests/fixed_point/test_fp_arithmetic_sugar.py
dpaiton: I know, I thought it was more clear to be explicit here for someone reading through the file. i.e. I liked this better than `assert not FALSE_STATEMENT` and `assert TRUE_STATEMENT`. I know it's atypical, I just thought it was easier to parse in such a test.

tests/fixed_point/test_fp_class.py
dpaiton: This is answered in the docs for FixedPoint (which I now realize are not rendering in the docs page 🤦). It's a long story, so I'll leave it for when we meet and explain FixedPoint.

tests/fixed_point/test_fp_class.py
dpaiton: Not necessarily a better test, but yes we can check that. This is the reason I implemented `__hash__` in the first place, we use FixedPoint as dictionary keys throughout the codebase. But an explicit test makes sense. I'll add one.

PR #462, opened 2023-05-25
Modify apeworx demo & utils to use FixedPoint lib

dpaiton:


examples/evm_bots.py
```diff
@@ -512,78 +542,84 @@ def log_and_show_block_info():
...
     project = ape_utils.HyperdriveProject(Path.cwd())
-    base: ContractInstance
-    if config.scratch["devnet"]:
-        assert isinstance(deployer, TestAccountAPI)
-        base: ContractInstance = deployer.deploy(project.get_contract("ERC20Mintable"))
-        fixed_math: ContractInstance = deployer.deploy(project.get_contract("MockFixedPointMath"))
+    base_instance: ContractInstance = None
+    if experiment_config.scratch["devnet"]:
+        assert isinstance(account_deployer, TestAccountAPI)
+        project_container: ContractContainer = project.get_contract("ERC20Mintable")
```
wakamex, line 607: I'd put this back into a one-liner like `fixed_math` below it. `project_container` also sounds like it refers to the hyperdrive project.

examples/evm_bots.py
```diff
@@ -512,78 +542,84 @@ def log_and_show_block_info():
...
+        fixed_math: ContractInstance = account_deployer.deploy(project.get_contract("MockFixedPointMath"))
+        base_instance = account_deployer.deploy(project_container)
     else:
-        base: ContractInstance = ape_utils.get_instance(DAI_ADDRESS, provider=provider)  # sDai
-    if config.scratch["devnet"]:
-        hyperdrive: ContractInstance = deploy_hyperdrive()
+        base_instance = ape_utils.get_instance(DAI_ADDRESS, provider=provider)  # sDai
+    hyperdrive_instance: ContractInstance = None
```
wakamex, line 612: no need for this line, since `hyperdrive_instance` is defined on both conditional paths. unless you were just looking to add typehints (which you could do on each of the 2 conditional paths... or once here).

examples/evm_bots.py
```diff
@@ -512,78 +542,84 @@ def log_and_show_block_info():
...
     with ape.accounts.use_sender(deployer):
-        base.approve(hyperdrive_, initial_supply)
-        hyperdrive_.initialize(initial_supply, initial_apr, deployer, True)
-    return hyperdrive_
+        base_contract.approve(hyperdrive, initial_supply.int_value)
+        hyperdrive.initialize(initial_supply.int_value, initial_apr.int_value, deployer, True)
+    return hyperdrive
 
 
 if __name__ == "__main__":
```
wakamex, line 586: should we move this into a `main()` function since you already added passing variables between functions?

wakamex (approved):
lgtm

examples/evm_bots.py
dpaiton: Yeah I did that on purpose, it's the same I just like this flow better.

examples/evm_bots.py
dpaiton: Next pass-through, yes. I didn't carefully remove the global variable use, just as I noticed it. So we will need to go over the file again when we move stuff into elfpy.

PR #484, opened 2023-05-30
generalized bot settings working with FixedPoint

wakamex:
1 fix to cast ape output to FixedPoint, rest is creating main() and generalizing the settings

- bugfix: cast to `FixedPoint` in `apeworx_integrations.py` when updating user wallets
- add `--alchemy` command line parameter, default is `false`
- add `--fork_url` command line parameter, to override the default. I use `--fork_url="http://localhost:8547"` on my desktop to point to my local Goerli node.
- add `--fork_port` command line parameter, to override the default. I use `-fork_port=8549` on my desktop to give Anvil its own port that doesn't clash with my local nodes.
- import ape's logging instance and set logging level `ERROR` (requires [fork](https://github.com/wakamex/ape/pull/2) for full impact)
- ~save all experiment variables to `experiment_config` (avoids having to pass them around, kind of like... `global` 😬)~
- `set_up_experiment()` saves experiment variables into discrete variables
- use an `address` file to load and save deployed contract addresses ( I use this locally even outside of docker )
- `set_up_devnet()` reads deployed contract addresses or deploys new contracts
- `set_up_ape()` creates deployer, provider, load existing addresses (goerli or devnet) or deploy otherwise. also saves hyperdrive market config from deployed instance afterward.
- `do_policy()` executes each agent's policy
- `create_elfpy_market()` creates elfpy object from deployed smart contract
- ~`main()` runs very simple loop now:~
- add automine mode (goes heckin fast)

fixes: #491


examples/evm_bots.py
```diff
@@ -443,220 +490,300 @@ def create_agent(
...
 
-    Returns
-    -------
-    _sim_agents : dict[str, Agent]
-        Dictionary of agents.
-    dev_accounts : list[KeyfileAccount]
-        List of dev accounts.
+    Parameters
+    ----------
+    experiment_config : simulators.Config
```
dpaiton, line 498: As a rule of thumb we should be weary about functions that modify things in-place. It can cause confusion, since variables are modified "behind the scenes". Certain operations make sense to do so in-place, but they're typically member functions of classes. e.g. `my_array.sort()` might not return something, and instead just sort the array in-place.

examples/evm_bots.py
```diff
@@ -443,220 +490,300 @@ def create_agent(
...
+                base_=experiment_config.scratch["base_instance"],
                 on_chain_trade_info=on_chain_trade_info,
-                hyperdrive_contract=hyperdrive_contract,
-                config=experiment_config,
+                hyperdrive_contract=experiment_config.scratch["hyperdrive_instance"],
+                experiment_config=experiment_config,
             )
             _sim_agents[f"agent_{agent.wallet.address}"] = agent
-    return _sim_agents
+    experiment_config.scratch["sim_agents"] = _sim_agents
```
dpaiton, line 539: We should restrict `config` variables to be just that, configurations for the experiment. It should be easily summarized or logged. By putting more complicated objects, such as the simulation agents, into the scratch, you end up with a pattern that we do not want in practice. Specifically, the `config.scratch` dict shouldn't end up a catch-all holding anything we want to pass from function-to-function.

dpaiton (changes requested):
Overall I think this is a big improvement!

My primary suggestion is for you to go back to using individual function arguments instead of passing around the config variable and pulling out what you need. We do this some times (e.g. with `MarketState`), but in general it is not a good pattern to dynamically grow the dictionary with generated objects and then pull them out as needed in later functions.

Now that you've set up a `main` function, I think you can rename all of the variables their normal (consistent) names, and remove the trailing underscores. The problem before was because you had variables of the same name in the global `__main__` scope, and now you don't. This should also clean up the arguments, and remove some of the need for passing around the `experiment_config`.

dpaiton (approved):
The evm_bots script is great! It really came together 🎉

PR #486, opened 2023-05-31
Fixing fixed point bug for float mod 1

slundqui:


slundqui:
This PR is currently breaking checkpointing tests

dpaiton (approved):
Great catch! We don't want the `int` mod behavior anymore.

PR #489, opened 2023-05-31
don't lint .venv in hyperdrive

sentilesdal:
This was bogging down VSCode.  Make sure that .venv inside the hyperdrive repo doesn't get linted etc.


wakamex (approved):
lgtm

PR #490, opened 2023-05-31
Building notebooks in docs

slundqui:
This PR allows for building in notebooks in `examples/*_notebooks.py` into html files that Sphinx docs link to:

Screenshot 2023-05-31 at 12 36 48 PM

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,7 @@
+Elfpy documentation and notebook examples can be built locally using the provided Makefile. To build local documentations that include notebook examples in `examples/*_notebooks.py`, run from this directory
+
+```
+make notebook;
+make html
+```
+This executes and converts all the `jupytext` notebooks in `examples/*_notebooks.py` to html files and creates the documentation with links to the notebooks using Sphinx. Note that documentation built in CI does not include notebook examples.
--- a/.gitignore
+++ b/.gitignore
@@ -57,8 +57,8 @@ elf-env
 docs/source/_autosummary
 docs/source/generated
 docs/source/autoapi
+docs/source/examples
 docs/build
-docs/notebook_build
 docs/make_outputs.log
 
 # mypy
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,6 +14,9 @@ help:
 
 .PHONY: help Makefile
 
+notebook:
+	@../scripts/notebook_docs.sh
+
 # Catch-all target: route all unknown targets to Sphinx using the new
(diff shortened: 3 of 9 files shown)


wakamex:
so we put notebooks we want to turn into static docs in `examples/_static` going forward? is the Vercel deployment empty because we have nothing there right now? that is, this page: <https://elf-simulations-fefuyc34c-delvtech.vercel.app/examples/index.html>

wakamex:
actually this uses jupytext to convert from `examples` to `docs/source/examples/_static/`.
```
outdir=$parent_path/../docs/source/examples/_static/
indir=$parent_path/../examples/
```

slundqui:
Not quite, notebooks that you want to turn into docs go into `examples/` with the `_notebook.py` suffix. `make notebook` automatically turns the `jupytext` notebooks into `html`, which gets put into `examples/_static`.

We decided that we didn't want to build notebooks for CI, so Sphinx is still able to build docs without these notebook files, which is why you're not seeing them in Vercel. Hence, we only support manual building of these notebook docs for now.

wakamex:
so if we want an `examples/*_notebook.py` file to show up, we need to trigger that Sphinx build manually, then push the results? manual execution gives us a bit more control over the results, and we don't have to deal with CI breaking. we could create a workflow and have it triggered manually only with
```
on:
  workflow_dispatch:
```

docs/README.md
```diff
@@ -0,0 +1,7 @@
+Elfpy documentation and notebook examples can be built locally using the provided Makefile. To build local documentations that include notebook examples in `examples/*_notebooks.py`, run from this directory
+
+```
+make notebook;
+make html
```
wakamex, line 5: what does `make html` run? I only see `notebook` in `docs/Makefile`

slundqui: `make html` gets passed into Sphinx here:
https://github.com/delvtech/elf-simulations/blob/7f46eb985dfc16f60a8c69d258635929966129d5/docs/Makefile#L20

This makefile is autogenerated from Sphinx, I just added a `notebook` check into this file

slundqui:
> so if we want an `examples/*_notebook.py` file to show up, we need to trigger that Sphinx build manually, then push the results? manual execution gives us a bit more control over the results, and we don't have to deal with CI breaking. we could create a workflow and have it triggered manually only with
> 
> ```
> on:
>   workflow_dispatch:
> ```

Unsure what you mean by "push the results", but I'm also unclear on the steps to actually upload these docs to wherever it's being hosted. I just left this for whoever's doing Releases to figure out.

wakamex (approved):
lgtm

PR #492, opened 2023-05-31
Remove use of defaultdict

sentilesdal:
Solidity will always a default value of zero instead of None in python.  We were trying to emulate this with defaultdict.  This had the unintended consequence of creating new entries if an unknown key was looked up.  We could either create a custom DefaultDict or just use the .get() method on dict to ensure that we always get a default zero value.  We chose the latter.

- defaultdict -> dict
- update callers
- update pyproject
- fix lint

### DIFF
--- a/elfpy/markets/base.py
+++ b/elfpy/markets/base.py
@@ -1,7 +1,6 @@
 """Market simulators store state information when interfacing AMM pricing models with users."""
 from __future__ import annotations
 
-from collections import defaultdict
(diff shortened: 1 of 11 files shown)


dpaiton (approved):
looks great

wakamex:
defaultdict just implements quantum waveform collapse, when you observe an index, it coalesces into a point form that existed only as a cloud of probability prior

PR #500, opened 2023-06-01
lint docstrings

wakamex:
lets you know when you missed a variable which seems valuable when working with lots of inputs and outputs:
`"base_instance" missing in parameter documentation`

unfortunately it doesn't check whether you used a totally different description for the same variable name in another docstring.

fixes: #501

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -77,14 +77,15 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python module names) to load,
 # usually to register additional checkers.
-load-plugins=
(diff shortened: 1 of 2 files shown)


sentilesdal (commented):
love this!  is it a pain to land with our codebase that has missing docstrings?

wakamex:
I think the biggest issue is it's not as complete as pylint. here's a list of missing features: https://github.com/astral-sh/ruff/issues/970

for instance (pun intended) I just checked and it doesn't handle `too-many-instance-attributes` which I was looking for a second opinion on just now

sentilesdal:
Closing as this was addressed in #694

PR #502, opened 2023-06-02
add sqrt to FixedPoint

dpaiton:


elfpy/math/fixed_point_math.py
```diff
@@ -54,3 +54,8 @@ def exp(x: NUMERIC) -> NUMERIC:
                 return x
             return FixedPoint(scaled_value=FixedPointIntegerMath.exp(x.scaled_value))
         return type(x)(math.exp(x))
+
+    @staticmethod
+    def sqrt(x: NUMERIC) -> NUMERIC:
+        """Performs sqrt(x)"""
+        return type(x)(math.sqrt(x))
```
slundqui, line 61: What happens here if x is a fixed point? Does math.sqrt(x) work with fixed points?

slundqui (approved):

elfpy/math/fixed_point_math.py
dpaiton: Yup! It calls the sqrt dunder

PR #505, opened 2023-06-04
Add Fixtures for Cross Platform Tests

sentilesdal:
Adds test fixtures to be used for the the cross-platform tests.  The fixtures are not limited to the tests however, this is actually a pattern we could extend to the rest of the tests in the code base if we find use for it.  For now though we are creating a `hyperdrive_fixture` that has does the things necessary to spin up both a simulated hyperdrive market and locally deployed hyperdrive contract.  agents are supplied for both.  Examples tests initializing the markets are provided, both in the class format and standalone def format.  There is currently a difference between how the two markets initialize their bond reserves so that will need to be resolved for the test to actually work.  The fixed point arithmetic looks exactly correct though which is exciting.


tests/cross_platform/utils.py
```diff
@@ -0,0 +1,87 @@
...
+        <https://docs.apeworx.io/ape/stable/methoddocs/contracts.html#ape.contracts.base.ContractInstance>`_
+        wrapped around the initialized MockHyperdriveTestnet smart contract.
+    agent_address: ape.api.accounts.AccountAPI
+        Ape address container, or `AccountAPI
+        <https://docs.apeworx.io/ape/stable/methoddocs/api.html#ape.api.accounts.AccountAPI>`_
+        representing the agent which is executing the action.
+    """
+    # pylint: disable=too-many-arguments
+    pool_config = hyperdrive_data_contract.getPoolConfig()
+    print(f"{pool_config=}")
```
dpaiton, line 51: no prints

dpaiton (approved):
remove prints then good to go!

PR #514, opened 2023-06-07
convert snakes to camels

wakamex:
add helper to avoid repeated code


wakamex:
incorporated in https://github.com/delvtech/elf-simulations/pull/496

PR #526, opened 2023-06-08
Time stretch analysis

jrhea:
This PR contains the analysis used to calculate the new time stretch approximation and the new formula for calculate initial bond reserves.


jrhea:
This belongs in the analysis repo. closing

PR #528, opened 2023-06-08
Start evm_bots refactors, add `HALT_ON_ERROR` option for running evm_bots

sentilesdal:
This PR does two things
  Start the process of splitting up evm_bots.
  Adds a halt_on_error option so that a failed transaction will not stop the bots from running.

Cleanups were made along the way while moving functions out of evm bots.

### DIFF
--- a/elfpy/bots/get_env_args.py
+++ b/elfpy/bots/get_env_args.py
@@ -1,18 +1,17 @@
 """Get environment arguments for bots"""
 import os
 from dataclasses import dataclass
-from enum import Enum, auto
-from typing import Literal
+from enum import Enum
 
 from elfpy import DEFAULT_LOG_MAXBYTES
 
 
 class LogLevel(Enum):
-    DEBUG = auto()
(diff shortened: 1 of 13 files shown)


elfpy/bots/get_env_args.py
```diff
@@ -0,0 +1,81 @@
+"""Get environment arguments for bots"""
+import os
+from dataclasses import dataclass
+from enum import Enum
+
+from elfpy import DEFAULT_LOG_MAXBYTES
+
+
+class LogLevel(Enum):
```
dpaiton, line 9: Can you unify this with the existing solution? https://github.com/delvtech/elf-simulations/blob/main/elfpy/utils/outputs.py#L463

dpaiton (approved):
looks great!

elfpy/bots/get_env_args.py
sentilesdal: ya i'll make a logging.py soon enuff

PR #544, opened 2023-06-10
Reorganizes contract acquisition files and fixes up logging

dpaiton:
evm_bots and the data acquisition scripts all log to stdout and to file now using stdlib
data was moved into elfpy and the helper functions were moved to their own file.


elfpy/data/contract_interface.py
```diff
@@ -1,84 +1,55 @@
...
+    # pylint: disable=too-few-public-methods
 
-    def default(self, o):
-        if isinstance(o, HexBytes):
-            return o.hex()
-        if isinstance(o, (AttributeDict, MutableAttributeDict)):
-            return dict(o)
-        return super().default(o)
+    hyperdrive: str = attr.ib()
+    base_token: str = attr.ib()
```
wakamex, line 29: believe that's a hyperdrivemath address deployed in the latest

slundqui (approved):

elfpy/agents/agent.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, wallet_address: int, policy: BasePolicy = NoActionPolicy()):
     def __str__(self) -> str:
         # cls arg tells json how to handle numpy objects and nested dataclasses
         dict_to_encode = {str(key): str(value) for key, value in self.__dict__.items()}
```
wakamex, line 46: if everything is a string, do we need a custom encoder? :thinking:

wakamex (approved):
lgtm

elfpy/data/contract_interface.py
dpaiton: I'm not sure what you're saying. The full diff is super confusing because it tried to write contract_interface as a diff from acquire_transactions. This is the class:
```
@attr.s
class HyperdriveAddressesJson:
    """Addresses for deployed Hyperdrive contracts."""

    # pylint: disable=too-few-public-methods

    hyperdrive: str = attr.ib()
    base_token: str = attr.ib()
```

elfpy/agents/agent.py
dpaiton: Actually this is a legacy function from the wallet and would not work. Agent is not a dataclass, so it does not have a `__dict__` function. I removed it completely. We can add a pretty print later if we need it.

PR #545, opened 2023-06-11
fix args in logs

sentilesdal:


sentilesdal:
closing in favor of #548

PR #548, opened 2023-06-12
Fix args in transaction event logs

sentilesdal:
Not sure what the issue was but I went in and used proper types that Web3 was passing down and now the args are decoded properly.


dpaiton (approved):
change prints to logs, then lgtm

PR #555, opened 2023-06-12
Add Flask server to create/list/kill python scripts.

sentilesdal:
This PR adds a simple flask server that lets us spin up/down evm_bot processes (or any other python script).  The process now looks like this in development:
in infra: `dc up -d`
in elf_simulations: 
```bash
❯❯ python elfpy/bots/bot_server.py examples/evm_bots.py
```
then in a different terminal:
```bash
❯❯ curl -X POST -H "Content-Type: application/json" -d @my_config.json http://127.0.0.1:5000/run_script
❯❯ curl -X GET -H "Content-Type: application/json" http://127.0.0.1:5000/list_processes
❯❯ curl -X POST -H "Content-Type: application/json" -d '{"id": 1}' http://127.0.0.1:5000/kill_script
```

Note that this does not prevent anyone from still using evm_bots.py directly.  If we want, we can build more architecture to make things more streamlined locally.  This mostly helps us with our dockerized bot container deployments.

### DIFF
--- a/elfpy/bots/bot_server.py
+++ b/elfpy/bots/bot_server.py
@@ -27,10 +27,9 @@ def run_script():
     # Execute the python script with the provided JSON as an argument
     script_id = len(running_processes) + 1
     script_path = app.config["SCRIPT_PATH"]
-    process = subprocess.Popen(["python", script_path, temp_file_path])
-
-    # Store the process in the dictionary
-    running_processes[script_id] = process
+    with subprocess.Popen(["python", script_path, temp_file_path]) as process:
+        # Store the process in the dictionary
+        running_processes[script_id] = process
 
     return jsonify({"id": script_id}), 200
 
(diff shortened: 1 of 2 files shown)


elfpy/bots/bot_server.py
```diff
@@ -0,0 +1,72 @@
...
+@app.route("/list_processes", methods=["GET"])
+def list_processes():
+    """List all running python scripts."""
+    process_list = [{"id": script_id, "status": process.poll()} for script_id, process in running_processes.items()]
+    return jsonify({"processes": process_list}), 200
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("Usage: python app.py <script_path>")
```
dpaiton, line 66: `app.py` should be `bot_server.py`?

dpaiton (approved):
awesome!

wakamex (approved):
looks great to me
can't wait to rub arbitrary code on your machine

elfpy/bots/bot_server.py
sentilesdal: ya, will fix later

PR #559, opened 2023-06-12
fix logs

wakamex:
- delete second call to `setup_logging`
- keep `log_level` in human-readable text instead of int (consistent with comments and how we used to do it)
- add missing `delete_previous_logs` logging parameter (default to False)
- add new `log_formatter` parameter (default to current)
- add comment to tell the user what default values are in `bot_config.py`
- use explicit `var is True` logical checks to be more readable

### DIFF
--- a/elfpy/__init__.py
+++ b/elfpy/__init__.py
@@ -30,7 +30,7 @@ MAX_RESERVES_DIFFERENCE = FixedPoint(2e10)
 PRECISION_THRESHOLD: FixedPoint = FixedPoint(1 * 10**10)  # 1e-8 * 1e18 = 1e10
 
 # Logging defaults
-DEFAULT_LOG_LEVEL = logging.INFO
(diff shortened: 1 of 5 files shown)


elfpy/__init__.py
```diff
@@ -30,7 +30,7 @@
 PRECISION_THRESHOLD: FixedPoint = FixedPoint(1 * 10**10)  # 1e-8 * 1e18 = 1e10
 
 # Logging defaults
-DEFAULT_LOG_LEVEL = logging.INFO
+DEFAULT_LOG_LEVEL = "INFO"
```
sentilesdal, line 33: why not use logging.INFO here?

wakamex: cus that's a number and puts `"log_level": 20,` in `my_config.json` leaving me puzzled (I had to look it up)

slundqui (approved):
LGTM

PR #565, opened 2023-06-14
Acquire data bug fixes

slundqui:
Fixing an error where looking up pool info with block number sometimes fails. Also adding explicit limit for block lookbacks.


dpaiton (approved):
nice simple fix! 🤩

PR #572, opened 2023-06-17
use automine command and unpin ape

wakamex:
- use automine command available in ape-foundry 0.6.9 (nice)
- unpin ape since 0.6.11 includes 'quiet cached keys warnings'


slundqui (approved):
LGTM

PR #576, opened 2023-06-22
trade history applied marginally and simplified

wakamex:
trade history simplified to 1 dataframe (trade_history)
trade_history updates only marginally (missing blocks) which is faster


elfpy/utils/apeworx_integrations.py
```diff
@@ -84,6 +90,119 @@ class DefaultHyperdriveConfig:
...
+            relevant_trades = ((from_agent) | (to_agent)) & (trade_history["id"] == position_id)
+            positive_balance = int(trade_history.loc[relevant_trades & (to_agent), "value"].sum())
+            negative_balance = int(trade_history.loc[relevant_trades & (from_agent), "value"].sum())
+            balance = positive_balance - negative_balance
+            on_chain_balance = hyperdrive_instance.balanceOf(position_id, address)
+            first_relevant_row = relevant_trades.index[relevant_trades][0]
+            info = trade_history.loc[first_relevant_row, :]
+            if abs(balance - on_chain_balance) > tolerance:
+                log_str += f"{address[:8]} {info.trade_type:16} balance = ({balance/1e18:0.5f},{on_chain_balance/1e18:0.5f})"
+                log_str += " mismatch ❌"
```
dpaiton, line 154: pls no emoji

elfpy/utils/apeworx_integrations.py
```diff
@@ -332,56 +437,56 @@ def get_wallet_from_onchain_trade_info(
...
-    for position_id in info.unique_ids:  # loop across all unique positions
-        trades_in_position = ((info.trades["from"] == address) | (info.trades["to"] == address)) & (
-            info.trades["id"] == position_id
-        )
-        logging.info("found %s trades for %s in position %s", sum(trades_in_position), address[:8], position_id)
-        positive_balance = int(info.trades.loc[(trades_in_position) & (info.trades["to"] == address), "value"].sum())
-        negative_balance = int(info.trades.loc[(trades_in_position) & (info.trades["from"] == address), "value"].sum())
+    for position_id in trade_history["id"].unique():  # loop across all unique positions
+        from_agent = trade_history["from"] == address
+        to_agent = trade_history["to"] == address
```
dpaiton, line 442: What trades are we looking at here? I don't understand how there is a `from_agent` and `to_agent`? That framing makes me think it is peer-to-peer... wouldn't one of those two always be the hyperdrive pool (where which one depends on whether it is open or close)?

examples/evm_bots.py
```diff
@@ -240,28 +204,37 @@ def set_up_agents(
...
                 base_instance=base_instance,
-                on_chain_trade_info=on_chain_trade_info,
+                trade_history=trade_history,
                 hyperdrive_contract=hyperdrive_instance,
                 bot_config=bot_config,
                 rng=rng,
             )
+            log_str = "Created %s agent with wallet:"
+            for long in agent.wallet.longs:
+                print(f"{long=}")
```
dpaiton, line 228: reminder to remove prints

elfpy/utils/apeworx_integrations.py
```diff
@@ -84,6 +90,117 @@ class DefaultHyperdriveConfig:
     target_liquidity = FixedPoint(1 * 10**6)
 
 
+def inspect_dump(trade_history, agent_addresses, hyperdrive_instance, base_instance, tolerance=1e16):
```
dpaiton, line 93: Can you move both of these into evm_bots and refrain from adding functions to apeworx_integrations? It'll help with my tracking the changes that need to  be made.

examples/evm_bots.py
```diff
@@ -466,6 +436,12 @@ def set_up_ape(
         "devnet" if bot_config.devnet else network_choice,
         provider.get_block("latest").number,
     )
+    agent_addresses, trade_history = None, None
+    if config.load_state_id is not None:  # load state from specified id
+        logging.info("Loading state from id: %s", config.load_state_id)
+        load_state_block_number, addresses, agent_addresses, trade_history = load_state(bot_config, rng)
+        # print(f"Loaded state up to block number {load_state_block_number}")
```
dpaiton, line 443: ded prints

elfpy/utils/apeworx_integrations.py
wakamex: done in bda60758a4feb8e5df6547ebb6cf820f52a5ec1f

elfpy/utils/apeworx_integrations.py
```diff
@@ -84,6 +90,119 @@ class DefaultHyperdriveConfig:
     target_liquidity = FixedPoint(1 * 10**6)
 
 
+def inspect_dump(trade_history, agent_addresses, hyperdrive_instance, base_instance, tolerance = 1e16):
```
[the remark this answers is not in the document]
wakamex: done in 1de6bfa6d40cd7cf76a3850df6f46820dc42ff98

elfpy/utils/apeworx_integrations.py
wakamex: yup exactly, it's basically a distinction between

- opening a trade transfers a minted position token TO the agent, INCREASING their balance
- closing a trade transfers a burned position token FROM the agent, DECREASING their balance

so it just explicitly separates those two types of transfers, which are then added separately into `positive_balance` and `negative_balance` just for ease of inspection/debugging. they could otherwise just be added in one line without being assigned to those intermediate variables. that's how it used to be. it just makes reasoning and debugging harder.

elfpy/bots/bot_config.py
```diff
@@ -48,6 +48,10 @@ class BotConfig(types.FrozenClass):
     # In general the bot will be more risk averse as it grows to infinity.
     # A value of 0 will usually disable it.
     risk_threshold: float = 0.0
+    # whether or not to dump state at every block
```
dpaiton, line 51: this is not at every block now, right? only when reverts happen?

dpaiton (approved):
It looks good to me.

Most of the bots & ape stuff is not tested and this is a pretty big new feature, so please take some time to thoroughly test it yourself by running in different scenarios and viewing the outputs. e.g. force a crash, reload from the crash, monitor logs & crash reports with runs that fail occasionally, etc.

elfpy/bots/bot_config.py
wakamex: changed to only in try/except

dpaiton: Right, so can you fix the comment?

dpaiton:
Can you move the two tests into their own subfolder?

wakamex:
> Can you move the two tests into their own subfolder?

they'd be the only file in its own subfolder, are you sure? 
- `test_solidity` has 10 files
- `pricing_models` has 6 files
- `cross_platform` has 11 files
- base `tests` folder has 10 files included `test_dumps.py`

PR #578, opened 2023-06-22
use custom github token

wakamex:
followed this recommendation: https://github.com/actions/setup-python/blob/v4/docs/advanced-usage.md#avoiding-rate-limit-issues

checked that the rate limit is actually being used using the linked example: https://github.com/actions/setup-python/pull/443#issuecomment-1206776401


slundqui (approved):
LGTM

PR #585, opened 2023-06-27
Save poolInfo data from hyperdrive to postgres

sentilesdal:
- add poolinfo schema and methods for postgres
- acquire the data for poolinfo
- add some tests

### DIFF
--- a/elfpy/data/pool_info.py
+++ b/elfpy/data/pool_info.py
@@ -1,13 +1,17 @@
 """Pool info struct retured from hyperdrive contract"""
-from dataclasses import dataclass
+from __future__ import annotations
 
-# solidity returns things in camelCase.  Keeping the formatting to indicate the source.
-# pylint: disable=invalid-name
+from dataclasses import dataclass
 
 
 @dataclass
 class PoolInfo:
     """Pool info struct returned from hyperdrive contract"""
+
+    # solidity returns things in camelCase.  Keeping the formatting to indicate the source.
+    # pylint: disable=invalid-name
+    # this is what is returned from the contract, no choice here
+    # pylint: disable=too-many-instance-attributes
 
     shareReserves: int | None = None
     bondReserves: int | None = None
--- a/elfpy/data/pool_info.py
+++ b/elfpy/data/pool_info.py
@@ -1,0 +1,24 @@
+"""Pool info struct retured from hyperdrive contract"""
+from dataclasses import dataclass
+
+# solidity returns things in camelCase.  Keeping the formatting to indicate the source.
+# pylint: disable=invalid-name
+
+
+@dataclass
+class PoolInfo:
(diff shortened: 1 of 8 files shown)


elfpy/data/postgres.py
```diff
@@ -0,0 +1,106 @@
...
+    """Transactions Schema"""
+
+    __tablename__ = "transactions"
+
+    id = Column(String, primary_key=True)
+    user_id = Column(String, ForeignKey("users.id"))
+    amount = Column(Integer)
+
+
+class PoolInfoTable(Base):
```
slundqui, line 34: I wonder if there's a way to automatically get the schema between the data class and this, so that the schema lives in one place (we'll have to do checks to ensure the current data class adheres to the existing data schema on postgres)

elfpy/data/postgres.py
```diff
@@ -0,0 +1,106 @@
...
+
+    return session
+
+
+def close_session(session):
+    """Close the session"""
+    session.close()
+
+
+def add_pool_infos(pool_infos: list[PoolInfo], session):
```
slundqui, line 82: The above suggestion will simplify this to be `PoolInfoTable(**pool_info)`

pyproject.toml
```diff
@@ -38,6 +38,7 @@ with-dependencies = [
     "numpy",
     "pandas",
     "pylint",
+    "psycopg2",
```
slundqui, line 41: Should sqlalchemy be part of this list?

slundqui (approved):

elfpy/data/postgres.py
sentilesdal: I love this and yes I think we can accomplish this with some codegen tools!

elfpy/data/postgres.py
sentilesdal: yeah that would be nice.  i'd love for us to codegen our way to guarantee the 1-1 relationship.  I think we do this buy scanning the abi and generating both the dataclass and the table schema.

pyproject.toml
sentilesdal: yes thanks.

elfpy/data/pool_info.py
```diff
@@ -0,0 +1,24 @@
...
+    bondReserves: int = 0
+    lpTotalSupply: int = 0
+    sharePrice: int = 0
+    longsOutstanding: int = 0
+    longAverageMaturityTime: int = 0
+    shortsOutstanding: int = 0
+    shortAverageMaturityTime: int = 0
+    shortBaseVolume: int = 0
+    withdrawalSharesReadyToWithdraw: int = 0
+    withdrawalSharesProceeds: int = 0
```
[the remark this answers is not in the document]
sentilesdal: great thought.  I'll initialize to None.

PR #601, opened 2023-06-28
Replacing fixedpoint to use external fixedpointmath library from monorepo

slundqui:


wakamex (approved):

elfpy/markets/hyperdrive/hyperdrive_market_deltas.py
```diff
@@ -1,10 +1,11 @@
 """Class for for storing the delta values for MarketState"""
 from __future__ import annotations
+
 from dataclasses import dataclass, field
 
-import elfpy.types as types
+from fixedpointmath import FixedPoint
 
-from elfpy.math import FixedPoint
+import elfpy.types as types
```
dpaiton, line 8: Since we're here, can you change this to `from elfpy import types`? We want to migrate to the `from` syntax wherever possible.

dpaiton: Actually never mind -- this shows up all over the place so it might as well be its own PR. I don't want to put that on you at this time.

dpaiton (approved):
here we gooo

PR #611, opened 2023-07-01
Data pipeline to postgres

slundqui:
Reworks acquire_data.py to read/write to/from postgres instead of from file.


elfpy/data/acquire_data.py
```diff
@@ -40,52 +40,64 @@ def main(
     transactions_hyperdrive_contract = contract_interface.get_hyperdrive_contract(
         transactions_abi_file_path, contracts_url, web3
     )
+
     # get pool config from hyperdrive contract
+    # TODO should this be added to postgres to put everything in one place?
```
dpaiton, line 45: Yes, ideally it would go into it's own DB that can be related to the others. The config is defines how HD was initialized, so it's needed to reproduce any bugs. You _could_ just replicate the data across rows in the pool info DB but I think the "proper" way to do it is to have it live on its own with some column that relates it to the others.

elfpy/data/acquire_data.py
```diff
@@ -40,52 +40,64 @@ def main(
...
-    latest_block_number = web3.eth.get_block_number()
-    lookback_block_limit = BlockNumber(lookback_block_limit)
-    if (latest_block_number - block_number) > lookback_block_limit:
-        block_number = BlockNumber(latest_block_number - lookback_block_limit)
-        logging.warning("Starting block is past lookback block limit, starting at block %s", block_number)
 
-    pool_info: list[PoolInfo] = []
-    pool_info_file = os.path.join(save_dir, "hyperdrive_pool_info.json")
+    # Make sure to not grab current block, as the current block is subject to change
+    # Current block is still currently being built
```
dpaiton, line 62: -"currently"

elfpy/data/acquire_data.py
```diff
@@ -103,19 +115,18 @@ def main(
                         continue
 
                 if block_pool_info:
-                    postgres.add_pool_infos([block_pool_info], session)
                     pool_info.append(block_pool_info)
+
```
dpaiton, line 119: pls try to avoid extraneous newlines, per [pep8](https://peps.python.org/pep-0008/#:~:text=Extra%20blank%20lines,indicate%20logical%20sections) they should be used sparingly to differentiate logical sections. We've been using commented lines to serve this purpose instead, since they explain the different sections and also serve as a visual break.

elfpy/data/contract_interface.py
```diff
@@ -109,24 +112,104 @@ def fetch_and_decode_logs(web3: Web3, contract: Contract, tx_receipt: TxReceipt)
...
 
-def fetch_transactions_for_block(
-    web3: Web3, contract: Contract, block_number: BlockNumber
-) -> list[dict[str, Any]] | None:
+def convert_fixedpoint(input_val: int | None) -> float | None:
+    """Given a scaled value int, converts it to an unscaled value in float, while dealing with Nones"""
+
+    # We cast to FixedPoint, then to floats to keep noise to a minimum
+    # This is assuming there's no loss of precision going from Fixedpoint to float
+    # Once this gets fed into postgres, postgres has fixed precision Numeric type
```
dpaiton, line 120: Would it be better to pass the `fixedpoint.scaled_value` in directly, then? Casting to float can introduce noice (it is done via `float = fixedpoint.scaled_value / 1e18`, which will introduce noise that scales with the magnitude of the number).

I think this doesn't matter much in the post-processing phase, at least while our goal is only to make plots of macro trends. So feel free to leave it alone, just a thought.

elfpy/data/contract_interface.py
```diff
@@ -109,24 +112,104 @@ def fetch_and_decode_logs(web3: Web3, contract: Contract, tx_receipt: TxReceipt)
...
+def fetch_transactions_for_block(web3: Web3, contract: Contract, block_number: BlockNumber) -> list[Transaction]:
     """Fetch transactions related to the contract"""
     block: BlockData = web3.eth.get_block(block_number, full_transactions=True)
     transactions = block.get("transactions")
     if not transactions:
         logging.info("no transactions in block %s", block.get("number"))
-        return None
-    decoded_block_transactions = []
+        return []
+    # decoded_block_transactions = []
```
dpaiton, line 203: dead comment

elfpy/data/db_schema.py
```diff
@@ -0,0 +1,154 @@
...
+    # Method: initialize
+    input_params_contribution: Mapped[float | None] = mapped_column(Numeric, default=None)
+    input_params_apr: Mapped[float | None] = mapped_column(Numeric, default=None)
+    input_params_destination: Mapped[str | None] = mapped_column(String, default=None)
+    input_params_asUnderlying: Mapped[bool | None] = mapped_column(Boolean, default=None)
+
+    # Method: openLong
+    input_params_baseAmount: Mapped[float | None] = mapped_column(Numeric, default=None)
+    input_params_minOutput: Mapped[float | None] = mapped_column(Numeric, default=None)
+    # input_params_destination
```
dpaiton, line 83: What's the intent with all of these commented variables?

elfpy/data/db_schema.py
```diff
@@ -0,0 +1,154 @@
...
+    # args_spender
+    # args_id
+    event_value: Mapped[float | None] = mapped_column(Numeric, default=None)
+    event_operator: Mapped[str | None] = mapped_column(String, default=None)
+    event_id: Mapped[int | None] = mapped_column(Numeric, default=None)
+    # Fields calculated from base
+    event_prefix: Mapped[int | None] = mapped_column(Integer, default=None)
+    event_maturity_time: Mapped[int | None] = mapped_column(Numeric, default=None)
+
+    # Fields not used by postprocessing
```
dpaiton, line 133: Could move these into the docstring. But tbh I feel like we should just get everything and decide later (in postprocessing) if we need it. What's the computational expense for adding these columns? Is the number of columns something we need to be concerned about in general going forward?

elfpy/data/postgres.py
```diff
@@ -1,68 +1,37 @@
...
-    user_id = Column(String, ForeignKey("users.id"))
-    amount = Column(Integer)
-
-
-class PoolInfoTable(Base):
-    """PoolInfo Schema"""
-
-    # names are in snake case to indicate values came from solidity
-    # pylint: disable=invalid-name
+# TODO figure out what this table is supposed to hold
```
dpaiton, line 20: I guess this is a question for Jacob? But honestly if it's not being used elsewhere then I'd rather delete it instead of keeping it here in comments.

dpaiton (approved):
Left some comments, but otherwise it looks great. Did you try syncing this up with the full demo pipeline (devnet -> bot trades -> postgres -> streamlit)? I think that would be a good test to do before merging.

elfpy/data/db_schema.py
slundqui: Was going through each function call to solidity, and all of the parameters for each call. Some of the parameters are duplicates, but still wanted to keep track of parameters per function call, hence the commented variables

elfpy/data/contract_interface.py
```diff
@@ -250,14 +330,32 @@ def get_block_pool_info(web3_container: Web3, hyperdrive_contract: Contract, blo
...
     current_block: BlockData = web3_container.eth.get_block(block_number)
     current_block_timestamp = current_block.get("timestamp")
     if current_block_timestamp is None:
         raise AssertionError("Current block has no timestamp")
+
     pool_info_data_dict.update({"timestamp": current_block_timestamp})
     pool_info_data_dict.update({"blockNumber": block_number})
+
+    pool_info_dict = {}
+    for key in PoolInfo.__annotations__.keys():
```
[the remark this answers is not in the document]
slundqui: I think __dict__ returns too much, as PoolInfo isn't exactly just a Dataclass anymore

elfpy/data/acquire_data.py
slundqui: TODO in next PR

slundqui:
> Left some comments, but otherwise it looks great. Did you try syncing this up with the full demo pipeline (devnet -> bot trades -> postgres -> streamlit)? I think that would be a good test to do before merging.

This was tested with the demo pipeline.

elfpy/data/db_schema.py
slundqui: I'd argue that if the docstrings turn into web docs, I'd keep it where it is now

PR #630, opened 2023-07-05
improve pnl calc

wakamex:
calculate marginal and full updates of open share price
remove empty positions from wallet


elfpy/utils/apeworx_integrations.py
```diff
@@ -400,47 +400,44 @@ def get_wallet_from_trade_history(
...
-                            )
-                        }
-                    )
-                    logging.debug(
-                        "storing in wallet as %s",
-                        {
-                            mint_time: Short(
-                                balance=new_balance, open_share_price=FixedPoint(scaled_value=open_share_price)
+                else:  # we either do a marginal update or full weighted average for open share price
+                    if add_to_existing_wallet:  # marginal update
```
dpaiton, line 409: Can you summarize the logic in here? I'm confused as to why we are ok with setting the `previous_share_price` to 0.

dpaiton (commented):
Can you add a top level comment explaining the `get_wallet_from_trade_history` function logic in more detail? I'm having trouble following what is going on from the in-line comments & code. This will also make it easier to maintain functionality when we switch to web3.

elfpy/utils/apeworx_integrations.py
wakamex: that denotes the case where the previous balance is also 0, so the new weighted average = `delta_price * delta_amount + old_balance * old_price` where the value of old_price doesn't matter. I'll see if I can draw this out more clearly.

elfpy/utils/apeworx_integrations.py
```diff
@@ -397,50 +407,38 @@ def get_wallet_from_trade_history(
...
+                        )
+                        delta_open_share_price = FixedPoint(scaled_value=int(trade_history.share_price.iloc[0]))
+                        # weighted average update: new_y = ( old_x * old_y + new_x * new_y ) / (old_x + new_x)
+                        updated_open_share_price = (
+                            previous_balance * previous_share_price + delta_balance * delta_open_share_price
+                        ) / new_balance
+                        updated_open_share_price = FixedPoint(scaled_value=int(updated_open_share_price))
+                    else:  # weighted average across a bunch of trades, assuming trade_history contains EVERY trade
+                        value = trades_df["value"] * np.where(trades_df["from"] == address, -1, 1)
+                        updated_open_share_price = FixedPoint(int(np.average(trades_df["share_price"], weights=value)))
```
dpaiton, line 430: are you sure you want to cast this to `int`? The weighted average is going to be a float, so casting to `int` will perform a truncation of the decimal portion. The better way to convert to int is to [round half to even](https://en.wikipedia.org/wiki/Rounding#:~:text=Rounding%20half%20to%20even%5Bedit%5D). Or, assuming you want to keep the decimal information, why not cast it to float and pass it directly to FixedPoint? e.g. `FixedPoint(np.average(...).item())` should work.

dpaiton (approved):
thanks for getting this done!

elfpy/utils/apeworx_integrations.py
wakamex: changed to do so in a helper function with error-checking

PR #632, opened 2023-07-05
make postgres a required install

dpaiton:


INSTALL.md
```diff
@@ -39,13 +39,6 @@ If you intend to improve the documentation, then you must also install the packa
 python -m pip install --upgrade -e ".[with-dependencies,docs]"
 ```
 
-If you intend to work postgres then you'll also need to install the postgres and the required packages:
```
sentilesdal, line 42: can we add a note that postgresql is required for working with files in elfpy/data?

sentilesdal (approved):
lgtm

PR #634, opened 2023-07-06
Postgres wallet table

slundqui:


dpaiton (commented):
I don't think the pattern for `get_funding_contract` is what we want. My understanding is that the `addresses.base_token` address and the `ERC20Mintable` ABI arguments are tied to a "funding account". These are tied together inside Hyperdrive, which takes a "base token" that must follow the ERC20 standard. I might have this wrong, but if I'm right then I think we want to hardcode these things together on our end as well. @sentilesdal let us know if this sounds correct to you.

my proposal: I think the process of setting the constant `FUNDING_ABI_FILE_PATH` and passing that as an argument to `get_funding_contract` is an unnecessary abstraction given above. Assuming I'm right, an alternative route is to load all of the ABIs at the outset and then look for the correct ABI for a given token address. This might also remove the need to specify any of the `*_ABI_FILE_PATH` variables, replacing them with a single `ABI_FOLDER` argument. So the new workflow is to run `contract_interface.load_all_abis(ABI_FOLDER)` and then tie together the base address with the ERC20Mintable contract:

```
# Do these outside of the function since they may take time & should only be done once
addresses = fetch_addresses_from_url(contracts_url)
abis = load_all_abis(abi_folder)

def get_funding_contract(web3: Web3, abis: dict, addresses: HyperdriveAddressesJson) -> Contract:
    """Get the base contract for a given abi"""
    if "ERC20Mintable" not in abis:
        raise AssetionError("ERC20 ABI for minting base tokens was not provided")
    # choose the ABI from the list
    state_abi =abis["ERC20Mintable"]
    # get contract instance of hyperdrive
    # NOTE: not a "hyperdrive_contract" because we use that var name elsewhere to refer to a Contract from the "IHyperdrive" ABI
    funding_contract: Contract = web3.eth.contract(
        address=address.to_checksum_address(addresses.base_token), abi=state_abi
    ) 
    return funding_contract
```

elfpy/data/contract_interface.py
```diff
@@ -420,3 +435,87 @@ def get_hyperdrive_config(hyperdrive_contract: Contract) -> PoolConfig:
...
     return PoolConfig(**out_config)
+
+
+def get_wallet_info(
+    hyperdrive_contract: Contract,
+    base_contract: Contract,
+    block_number: BlockNumber,
+    transactions: list[Transaction],
+) -> list[WalletInfo]:
+    """Retrieves wallet information at a given block given a transaction
```
sentilesdal, line 446: We seem to be drifting from adding sphinx documentation from our functions.  lets stop the bleeding and add them to these functions.  We were pretty good about this for awhile, lets get back to it.  Look around the repo for examples.

elfpy/data/contract_interface.py
```diff
@@ -420,3 +435,87 @@ def get_hyperdrive_config(hyperdrive_contract: Contract) -> PoolConfig:
...
+            for _ in range(RETRY_COUNT):
+                try:
+                    num_custom_token_fp = hyperdrive_contract.functions.balanceOf(int(token_id), wallet_addr).call(
+                        block_identifier=block_number
+                    )
+                except ValueError:
+                    logging.warning("Error in getting custom token balance, retrying")
+                    time.sleep(1)
+                    continue
+            num_custom_token = convert_fixedpoint(num_custom_token_fp)
```
sentilesdal, line 507: what does `convert_fixedpoint` do?  is `num_custom_token_fp` a FixedPoint or floating point?  can't the FixedPoint class convert to float already?

elfpy/data/contract_interface.py
```diff
@@ -420,3 +435,87 @@ def get_hyperdrive_config(hyperdrive_contract: Contract) -> PoolConfig:
...
+                token_type = base_token_type + "-" + str(token_maturity_time)
+                maturity_time = token_maturity_time
+            else:
+                token_type = base_token_type
+                maturity_time = None
+
+            num_custom_token_fp = None
+            for _ in range(RETRY_COUNT):
+                try:
+                    num_custom_token_fp = hyperdrive_contract.functions.balanceOf(int(token_id), wallet_addr).call(
```
sentilesdal, line 500: balanceOf doesn't return FixedPoint does it?

elfpy/data/db_schema.py
```diff
@@ -22,6 +22,28 @@ class Base(MappedAsDataclass, DeclarativeBase):
...
+    # Note that we use postgres in production and sqlite in testing, but sqlite has issues with
+    # autoincrement with BigIntegers. Hence, we use the Integer variant when using sqlite in tests
+    id: Mapped[int] = mapped_column(
+        BigInteger().with_variant(Integer, "sqlite"), primary_key=True, init=False, autoincrement=True
+    )
+
+    blockNumber: Mapped[int] = mapped_column(BigInteger, ForeignKey("poolinfo.blockNumber"), index=True)
+    walletAddress: Mapped[Union[str, None]] = mapped_column(String, index=True, default=None)
+    baseTokenType: Mapped[Union[str, None]] = mapped_column(String, index=True, default=None)
+    tokenType: Mapped[Union[str, None]] = mapped_column(String, default=None)
```
sentilesdal, line 42: what are tokenTypes?  the names of the tokens?

elfpy/data/postgres.py
```diff
@@ -161,6 +163,68 @@ def get_transactions(session: Session, start_block: int | None = None, end_block
     return pd.read_sql(query.statement, con=session.connection())
 
 
+def get_all_wallet_info(session: Session, start_block: int | None = None, end_block: int | None = None) -> pd.DataFrame:
```
sentilesdal, line 166: won't comment everywhere, but let's sphinx all public methods.

elfpy/data/postgres.py
```diff
@@ -52,6 +40,20 @@ def close_session(session: Session):
     session.close()
 
 
+def add_wallet_infos(wallet_infos: list[WalletInfo], session: Session):
+    # TODO is there a case where a single wallet can make multiple trades with the same
+    # blockNumber, walletAddress, and tokenType?
+
```
wakamex, line 46: remove if no longer necessary

elfpy/data/contract_interface.py
dpaiton: agreed! Thanks for this Matt, we all need to be better about it.

slundqui: Sure, a lot of these functions were not written by me, but I'll add docstrings to the functions I've touched.

elfpy/data/contract_interface.py
slundqui: convert_fixedpoint is the way I'm doing scaled_values -> unscaled_value using the FixedPoint class. It also handles the case when the input is None.

elfpy/data/db_schema.py
slundqui: tokenType can either be "BASE", "LONG-<maturity_time>", "SHORT-<maturity_time>", or "LP"

sentilesdal (approved):
looks great!

PR #638, opened 2023-07-06
Using bots python script to generate default config

slundqui:


dpaiton (approved):
lgtm
consider putting it in `elfpy/bots` and renaming to `bot_config.py.default`

PR #640, opened 2023-07-06
Revert "Reverting back to using GITHUB_TOKEN (#602)"

wakamex:
track github rate limit using custom token on every workflow. shouldn't get random rate limit errors anymore.
- two steps added to each workflow .yml that prints to the action log the rate limit at the start and end
- a separate ratelimit.yml action triggers at the completion of the lint workflow (the slowest)
- this flow is parameterized to comment on the associated PR issue with the remaining rate limit
- this can be turned off by removing the `post_comment: true` parameter, at which point the `sha` parameter is no longer needed (it's only used to look up the associated PR based on the SHA of the triggerring workflow)
- each of these steps uses this github action stored in my own repo: https://github.com/wakamex/rate-limit-action/

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,14 +21,14 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          token: ${{ secrets.GITHUB_TOKEN}}
+          token: ${{ secrets.GH_GITHUB_COM_TOKEN}}
 
       - name: Log in to the Container registry
         uses: docker/login-action@v2.2.0
         with:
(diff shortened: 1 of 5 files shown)


slundqui (approved):
Great! What did you have to change to get this working again?

wakamex:
> Great! What did you have to change to get this working again?

nothing, I think it was just a random outage. I just added a bunch of ways to see what it's doing. but I wouldn't be surprised if we hit a rate limit again, apparently there's something called "secondary rate limit" which sounds unavoidable. but using a token with a 5000 rate limit instead of the default 60 should do a lot of good.

PR #650, opened 2023-07-12
fixes logging

dpaiton:
The logging config is a string that specifies the log format. However the `logging.Formatter` function takes in both the log format _and_ the date format. So we need to take a str as the `setup_logging` function argument and pass it as well as the `elfpy` default for the date format to `logging.Formatter()`.

I also added some comments and TODOS to improve testing & documentation in the `setup_logging` function.


sentilesdal (approved):
oops thanks, please add pr description.

PR #651, opened 2023-07-12
Add a crash report issue template.

sentilesdal:
This is mostly for finding bugs with the hyperdrive protocol.  We can also use this if we have bugs with the simulation repo as well.


dpaiton (approved):
nice!

PR #655, opened 2023-07-12
web3 functions and classes refactor

dpaiton:


elfpy/eth/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Interface functions"""
+
+from . import abi
+from . import accounts
+from .rpc_interface import *
```
slundqui, line 5: I'm not a fan of `import *` here, makes it a bit hard to track down where something is getting imported from.

slundqui (approved):

PR #658, opened 2023-07-12
Fix tests broken from fixedpointmath version update.

sentilesdal:
A couple tests pass now that used to have a DivisionByZero error now that we can handle negative numbers.  We still want to have the tests make sure we are limiting terms to less than a year so I added that assertion.

### DIFF
--- a/tests/pricing_models/test_calc_in_given_out.py
+++ b/tests/pricing_models/test_calc_in_given_out.py
@@ -360,7 +360,7 @@ class TestCalcInGivenOut(unittest.TestCase):
                 time_remaining=time.StretchedTime(
                     days=FixedPoint("500.0"), time_stretch=FixedPoint("1.1"), normalizing_constant=FixedPoint("365.0")
                 ),
-                exception_type=(AssertionError),
+                exception_type=AssertionError,
             ),
             CalcInGivenOutFailureTestCase(  # test 11
                 # amount very high, can't make trade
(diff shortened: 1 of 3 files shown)


tests/pricing_models/test_calc_in_given_out.py
```diff
@@ -360,7 +360,7 @@ def test_calc_in_given_out_failure(self):
                 time_remaining=time.StretchedTime(
                     days=FixedPoint("500.0"), time_stretch=FixedPoint("1.1"), normalizing_constant=FixedPoint("365.0")
                 ),
-                exception_type=(AssertionError, fperrors.DivisionByZero),
+                exception_type=(AssertionError),
```
wakamex, line 363: does this need to be a tuple?

tests/pricing_models/test_calc_out_given_in.py
```diff
@@ -1621,7 +1621,7 @@ def test_calc_out_given_in_failure(self):
                 time_remaining=time.StretchedTime(
                     days=FixedPoint("500.0"), time_stretch=FixedPoint("1.0"), normalizing_constant=FixedPoint("365.0")
                 ),
-                exception_type=(AssertionError, fperrors.DivisionByZero),
+                exception_type=(AssertionError),
```
wakamex, line 1624: does this need to be a tuple?

sentilesdal: nope

tests/pricing_models/test_calc_in_given_out.py
sentilesdal: nope

wakamex (approved):

PR #667, opened 2023-07-13
Username to wallet address

slundqui:
When running evm bots, will read `username` from bot config and add mapping between bots and username to postgres

### DIFF
--- a/elfpy/data/postgres.py
+++ b/elfpy/data/postgres.py
@@ -113,6 +113,7 @@ def add_wallet_infos(wallet_infos: list[WalletInfo], session: Session):
     try:
         session.commit()
     except sqlalchemy.exc.DataError as err:  # type: ignore
+        session.rollback()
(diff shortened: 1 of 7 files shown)


elfpy/data/postgres.py
```diff
@@ -187,10 +194,51 @@ def add_transactions(transactions: list[Transaction], session: Session):
     try:
         session.commit()
     except sqlalchemy.exc.DataError as err:  # type: ignore
+        session.rollback()
         print(f"{transactions=}")
         raise err
 
 
+def add_user_map(username: str, addresses: list[str], session: Session):
```
dpaiton, line 202: can you add a `-> None` return type? This is the default, but better to be explicit imo.

dpaiton (approved):
Awesome, thanks a lot! 

Can you please add the line for changing the default username to `bots_default_config.py` as well?

PR #674, opened 2023-07-14
so many edited. please check and merge it

idiotcommerce:
@dpaiton 

i am very happy to contribute in this code. 

so many files edited. all are related to **fixedpointmath** related. 
due to versions update i think no **fixedpointmath** instad only it has **fixedpoint** . 

i run the acquir_data.py file. when i run the file all errors related fixedpoint only.  still not completed. so i just contribute this far.

please check

### DIFF
--- a/elfpy/__init__.py
+++ b/elfpy/__init__.py
@@ -6,14 +6,14 @@ from collections import defaultdict
 from typing import Any
 
 import matplotlib as mpl
-from fixedpointmath import FixedPoint
+from fixedpoint import FixedPoint
 
 # Setup barebones logging without a handler for users to adapt to their needs.
 logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 # This is the minimum allowed value to be passed into calculations to avoid
 # problems with sign flips that occur when the floating point range is exceeded.
-WEI = FixedPoint(scaled_value=1)  # smallest denomination of ether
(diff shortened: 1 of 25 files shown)


dpaiton:
@idiotcommerce the library that we used is `fixedpointmath` (located [here](https://pypi.org/project/fixedpointmath/)). You seem to have confused wit with an alternative library called `fixedpoint`, which is the source of the errors. Our code is expected to run with the correct packages installed, and the testing CI suite verifies that _most_ of it does.  I recommend starting over with the [install script](https://github.com/delvtech/elf-simulations/blob/main/INSTALL.md) to see if you can't get it all working without changing the required packages.

idiotcommerce:
sorry for that.  while i am executing the command 
(fastapienv) F:\gitcontrib\elf-simulations>pip install fixedpointmath

ERROR: Ignored the following versions that require a different python version: 0.1.0 Requires-Python **<=3.11,>=3.8; 0.1.1** Requires-Python **<=3.11,>=3.8; 0.1.2** Requires-Python **<=3.11,>=3.8**
ERROR: Could not find a version that satisfies the requirement fixedpointmath (from versions: none)
ERROR: No matching distribution found for fixedpointmath

sorry for my mistake, if it is working fine.  but if i run the above command i got the error like this


image

PR #676, opened 2023-07-18
test values not series

wakamex:


elfpy/data/postgres.py
```diff
@@ -326,16 +326,16 @@ def get_transactions(session: Session, start_block: int | None = None, end_block
...
-        end_block = _get_latest_block_number_transactions(session) + end_block + 1
+        end_block = get_latest_block_number_from_table(Transaction.__tablename__, session) + end_block + 1
 
     if start_block is not None:
         query = query.filter(Transaction.blockNumber >= start_block)
     if end_block is not None:
         query = query.filter(Transaction.blockNumber < end_block)
 
-    return pd.read_sql(query.statement, con=session.connection())
+    return pd.read_sql(query.statement, con=session.connection()).set_index("blockNumber")
```
slundqui, line 338: I think this is fine, but just a reminder that in the transactions, blockNumber isn't unique. Pandas is able to handle non-unique indices in most functions, but I do think there are limits to this.

slundqui (approved):

elfpy/data/postgres.py
wakamex: yes once/if we have more than 1 transaction per block, this would need to be removed, maybe to a multi-index of blockNumber and transactionIndex. we currently don't support that however.

PR #678, opened 2023-07-18
fix requires-python

wakamex:
according to [python.org](https://packaging.python.org/en/latest/specifications/core-metadata/#requires-python) have to match [version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/) most recently defined by [PEP 440](https://peps.python.org/pep-0440/)
[examples](https://peps.python.org/pep-0440/#examples):
>    ~=3.1: version 3.1 or later, but not version 4.0 or later.
    ~=3.1.2: version 3.1.2 or later, but not version 3.2.0 or later.
    ~=3.1a1: version 3.1a1 or later, but not version 4.0 or later.
    == 3.1: specifically version 3.1 (or 3.1.0), excludes all pre-releases, post releases, developmental releases and any 3.1.x maintenance releases.
    == 3.1.*: any version that starts with 3.1. Equivalent to the ~=3.1.0 compatible release clause.
    ~=3.1.0, != 3.1.3: version 3.1.0 or later, but not version 3.1.3 and not version 3.2.0 or later.

I was alerted to this by [ruff](https://github.com/astral-sh/ruff) which was correctly throwing an error:
```
error: TOML parse error at line 20, column 19
   |
20 | requires-python = "3.10"
   |                   ^^^^^^
Failed to parse version:
3.10
^^^^
```

PS: also noticed pip would throw a warning:
```
pip install -e .
Obtaining file:///code/elfpy
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
WARNING: Package 'elfpy' has an invalid Requires-Python: Invalid specifier: '3.10'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,6 @@ with-dependencies = [
     "pytest",
     "python-dotenv",
     "scipy",
-    "tomli",
     "typing_extensions==4.5.0",
     # bot dependencies
     "web3",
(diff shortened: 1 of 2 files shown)


sentilesdal (approved):
looks good to me

PR #679, opened 2023-07-18
improve pnl calc in demo

wakamex:
add get_agent_positions in postgres.py to prepare agent position data for pnl calculation

### DIFF
--- a/elfpy/data/acquire_data.py
+++ b/elfpy/data/acquire_data.py
@@ -172,6 +172,7 @@ def main(
     )
 
     # get pool config from hyperdrive contract
+    # TODO: figure out best way to represent this. options are PoolInfo object in db_schema, TypeDict, etc.
     pool_config_dict = hyperdrive_interface.get_hyperdrive_config(hyperdrive_contract)
     for key in db_schema.PoolConfig.__annotations__:
         if key not in pool_config_dict:
--- a/elfpy/data/acquire_data.py
+++ b/elfpy/data/acquire_data.py
@@ -172,7 +172,11 @@ def main(
     )
 
     # get pool config from hyperdrive contract
-    pool_config = db_schema.PoolConfig(**hyperdrive_interface.get_hyperdrive_config(hyperdrive_contract))
+    pool_config_dict = hyperdrive_interface.get_hyperdrive_config(hyperdrive_contract)
+    for key in db_schema.PoolConfig.__annotations__:
+        if key not in pool_config_dict:
+            pool_config_dict[key] = None
+    pool_config = db_schema.PoolConfig(**pool_config_dict)
     postgres.add_pool_config(pool_config, session)
 
     # Get last entry of pool info in db
@@ -257,19 +261,24 @@ def main(
                     postgres.add_pool_infos([block_pool_info], session)
 
                 # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
-                checkpoint_info = None
+                block_checkpoint_info = None
                 for _ in range(RETRY_COUNT):
                     try:
-                        checkpoint_info = hyperdrive_interface.get_hyperdrive_checkpoint_info(
+                        checkpoint_info_dict = hyperdrive_interface.get_hyperdrive_checkpoint_info(
                             web3, hyperdrive_contract, block_number
                         )
+                        # Set defaults
+                        for key in db_schema.CheckpointInfo.__annotations__:
+                            if key not in checkpoint_info_dict:
+                                checkpoint_info_dict[key] = None
+                        block_checkpoint_info = db_schema.CheckpointInfo(**checkpoint_info_dict)
                         break
                     except ValueError:
                         logging.warning("Error in get_hyperdrive_checkpoint_info, retrying")
                         time.sleep(1)
                         continue
-                if checkpoint_info:
-                    postgres.add_checkpoint_infos([checkpoint_info], session)
+                if block_checkpoint_info:
+                    postgres.add_checkpoint_infos([block_checkpoint_info], session)
 
                 # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
                 block_transactions = None
--- a/elfpy/data/acquire_data.py
+++ b/elfpy/data/acquire_data.py
@@ -243,43 +243,47 @@ def main(
                     continue
                 # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
                 block_pool_info = None
+                pool_info_dict = None
                 for _ in range(RETRY_COUNT):
                     try:
                         pool_info_dict = hyperdrive_interface.get_hyperdrive_pool_info(
                             web3, hyperdrive_contract, block_number
                         )
-                        # Set defaults
-                        for key in db_schema.PoolInfo.__annotations__:
-                            if key not in pool_info_dict:
-                                pool_info_dict[key] = None
-                        block_pool_info = db_schema.PoolInfo(**pool_info_dict)
                         break
                     except ValueError:
                         logging.warning("Error in get_hyperdrive_pool_info, retrying")
                         time.sleep(1)
                         continue
-                if block_pool_info:
-                    postgres.add_pool_infos([block_pool_info], session)
+                if pool_info_dict:  # Proceed only if we have data, otherwise do nothing
+                    # Set defaults
+                    assert pool_info_dict is not None
+                    for key in db_schema.PoolInfo.__annotations__:
+                        if key not in pool_info_dict:
+                            pool_info_dict[key] = None
+                    # Add to db
+                    block_pool_info = db_schema.PoolInfo(**pool_info_dict)
+                    if block_pool_info:
+                        postgres.add_pool_infos([block_pool_info], session)
 
                 # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
-                block_checkpoint_info = None
+                checkpoint_info_dict = None
                 for _ in range(RETRY_COUNT):
                     try:
                         checkpoint_info_dict = hyperdrive_interface.get_hyperdrive_checkpoint_info(
                             web3, hyperdrive_contract, block_number
                         )
-                        # Set defaults
-                        for key in db_schema.CheckpointInfo.__annotations__:
-                            if key not in checkpoint_info_dict:
-                                checkpoint_info_dict[key] = None
-                        block_checkpoint_info = db_schema.CheckpointInfo(**checkpoint_info_dict)
                         break
                     except ValueError:
                         logging.warning("Error in get_hyperdrive_checkpoint_info, retrying")
                         time.sleep(1)
                         continue
-                if block_checkpoint_info:
-                    postgres.add_checkpoint_infos([block_checkpoint_info], session)
+                if checkpoint_info_dict:  # Proceed only if we have data, otherwise do nothing
+                    # Set defaults
+                    for key in db_schema.CheckpointInfo.__annotations__:
+                        if key not in checkpoint_info_dict:
+                            checkpoint_info_dict[key] = None
+                    # Add to db
+                    postgres.add_checkpoint_infos([db_schema.CheckpointInfo(**checkpoint_info_dict)], session)
 
                 # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
                 block_transactions = None
@@ -293,7 +297,7 @@ def main(
                         logging.warning("Error in fetch_transactions_for_block, retrying")
                         time.sleep(1)
                         continue
-                if block_transactions:
+                if block_transactions:  # Proceed only if we have data, otherwise do nothing
                     postgres.add_transactions(block_transactions, session)
                 if block_transactions and block_pool_info:
                     wallet_info_for_transactions = get_wallet_info(
--- a/elfpy/data/acquire_data.py
+++ b/elfpy/data/acquire_data.py
@@ -257,6 +257,21 @@ def main(
                     postgres.add_pool_infos([block_pool_info], session)
 
                 # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
+                checkpoint_info = None
+                for _ in range(RETRY_COUNT):
+                    try:
+                        checkpoint_info = hyperdrive_interface.get_hyperdrive_checkpoint_info(
+                            web3, hyperdrive_contract, block_number
+                        )
+                        break
+                    except ValueError:
+                        logging.warning("Error in get_hyperdrive_checkpoint_info, retrying")
+                        time.sleep(1)
+                        continue
+                if checkpoint_info:
+                    postgres.add_checkpoint_infos([checkpoint_info], session)
+
+                # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
                 block_transactions = None
                 for _ in range(RETRY_COUNT):
                     try:
--- a/elfpy/data/postgres.py
+++ b/elfpy/data/postgres.py
@@ -243,7 +243,7 @@ def add_pool_infos(pool_infos: list[PoolInfo], session: Session) -> None:
 
 
 def add_checkpoint_infos(checkpoint_infos: list[CheckpointInfo], session: Session) -> None:
-    """Add checkpoint info to the checkpointinfo table
+    """Add checkpoint info to the checkpointinfo table.
 
     Arguments
     ---------
--- a/elfpy/data/postgres.py
+++ b/elfpy/data/postgres.py
@@ -1,4 +1,4 @@
-"""Initialize Postgres Server"""
+"""Initialize Postgres Server."""
 
 from __future__ import annotations
 
@@ -8,10 +8,10 @@ from typing import Type
 
 import pandas as pd
 import sqlalchemy
-from sqlalchemy import URL, create_engine, func, inspect, MetaData, Table, exc
+from sqlalchemy import URL, MetaData, Table, create_engine, exc, func, inspect
 from sqlalchemy.orm import Session, sessionmaker
 
-from elfpy.data.db_schema import Base, PoolConfig, PoolInfo, CheckpointInfo, Transaction, UserMap, WalletInfo
+from elfpy.data.db_schema import Base, CheckpointInfo, PoolConfig, PoolInfo, Transaction, UserMap, WalletInfo
 
 # classes for sqlalchemy that define table schemas have no methods.
 # pylint: disable=too-few-public-methods
@@ -88,7 +88,7 @@ def query_tables(session: Session) -> list[str]:
 
     Returns
     -------
-    table_names : list[str]
+    list[str]
         A list of table names in the database
     """
     inspector = inspect(session.bind)  # nice gadget
@@ -103,7 +103,7 @@ def drop_table(session: Session, table_name: str) -> None:
     ---------
     session : Session
         The initialized session object
-    table_names : str
+    table_name : str
         The name of the table to be dropped
     """
     metadata = MetaData()
@@ -122,7 +122,6 @@ def initialize_session() -> Session:
     session : Session
         The initialized session object
     """
-
     postgres_config = build_postgres_config()
 
     url_object = URL.create(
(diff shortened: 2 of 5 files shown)


examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
+            else:
+                maturity = None
+
             spot_price = calculate_spot_price(
-                row.share_reserves,
-                row.bond_reserves,
-                row.lp_total_supply,
-                row.maturity_time,
-                row.block_timestamp,
+                state.shareReserves,
```
sentilesdal, line 198: There's a new calculate_spot_price that doesn't use lpTotalSupply in hyperdrive_pricing_model_sol.py which I think we should use.  It's a copy of what's used in the solidity.

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
+
+
+def calculate_pnl(pool_config, pool_info, checkpoint_info):
+    """Calculate pnl for all agents."""
+    #  pylint: disable=too-many-locals
+    session = postgres.initialize_session()
+    all_agent_info = get_agent_info(session)
+    position_duration = pool_config.positionDuration.iloc[0]
+
+    for _, agent_info in all_agent_info.items():
```
sentilesdal, line 185: `for agent_info in all_agent_info.values()`?

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
+        agent_info[agent] = Info(
+            wallet=wallet,
+            positions=positions,
+            deltas=deltas,
+            open_share_price=share_price_on_increases,
+        )
+    return agent_info
+
+
+def calculate_pnl(pool_config, pool_info, checkpoint_info):
```
sentilesdal, line 178: this function definitely requires a nice long description in the docstring explaining what the methodology is here.

sentilesdal: also types plz

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
+                    pnl += position.loc[position_name] * spot_price
+                elif position_name.startswith("SHORT"):
+                    # SHORT value is calculated as the:
+                    # total amount paid for the position (position * 1)
+                    # minus the closing cost (position * spot_price)
+                    # this equals position * (1 - spot_price)
+                    pnl += position.loc[position_name] * (1 - spot_price)
+
+            wallet.loc[block, "pnl"] = pnl
+        agent_info = wallet, positions, deltas, open_share_price
```
sentilesdal, line 225: why are we overwriting agent_info here?

examples/hackweek_demo/plot_pnl.py
sentilesdal: this returns an `agent_info`?  what is that?  shouldn't it return a pnl value per agent?

elfpy/data/acquire_data.py
```diff
@@ -225,6 +240,27 @@ def main(
                         continue
                 if block_pool_info:
                     postgres.add_pool_infos([block_pool_info], session)
+
+                # keep trying until it returns
+                block_checkpoint_info = None
+                for _ in range(RETRY_COUNT):
```
dpaiton, line 246: just wondering -- why would this break in the first place? no checkpoints yet?

elfpy/hyperdrive_interface/hyperdrive_interface.py
```diff
@@ -119,6 +119,39 @@ def get_hyperdrive_pool_info(web3: Web3, hyperdrive_contract: Contract, block_nu
     return pool_info
 
 
+def get_hyperdrive_checkpoint_info(
```
dpaiton, line 122: this is function is really well scoped, documented, and typed. Thanks!

examples/hackweek_demo/run_demo.py
```diff
@@ -54,9 +52,7 @@
 
 def get_ticker(data):
     """Given transaction data, return a subset of the dataframe"""
-    # Return reverse of methods to put most recent transactions at the top
-    out = data[["blockNumber", "input_method"]].set_index("blockNumber").iloc[::-1]
-    return out
+    return data[["input_method"]].iloc[::-1]
```
sentilesdal, line 55: does this warrant having a function, its a one-liner?  admittedly I don't know what it does really, is it actually specific to transaction data?

elfpy/data/postgres.py
```diff
@@ -59,6 +59,24 @@ def build_postgres_config() -> PostgresConfig:
     return PostgresConfig(**arg_dict)
 
 
+def query_tables(session):
+    """Return a list of tables in the database."""
+    stmt = text(
```
sentilesdal, line 64: I saw this in another PR and left a comment about doing this in a sqlAlchemy kinda way.  Same here.

elfpy/data/postgres.py
```diff
@@ -302,6 +339,46 @@ def get_pool_info(session: Session, start_block: int | None = None, end_block: i
...
+
+    Returns
+    -------
+    DataFrame
+        A DataFrame that consists of the queried checkpoint info
+    """
+
+    query = session.query(CheckpointInfo)
+
+    # Support for negative indices
```
sentilesdal, line 365: yikes, we need this?  blocks should never be negative!

elfpy/hyperdrive_interface/hyperdrive_interface.py
```diff
@@ -119,6 +119,39 @@ def get_hyperdrive_pool_info(web3: Web3, hyperdrive_contract: Contract, block_nu
...
+    web3: Web3, hyperdrive_contract: Contract, block_number: BlockNumber
+) -> dict[str, Any]:
```
sentilesdal, line 124: better than no type! :P

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
-                add_to_existing_wallet=agent_wallets[agent],
+Info = NamedTuple(
+    "Info", wallet=pd.DataFrame, positions=pd.DataFrame, deltas=pd.DataFrame, open_share_price=pd.DataFrame
+)
+
+
+def get_agent_info(session=None) -> dict[str, Info]:
+    """Arrange agent wallet info for plotting, while calculating open share price."""
+    if session is None:
+        session = postgres.initialize_session()
```
sentilesdal, line 124: I'd rather this throw an error.  Not the job of this function to initialize the postgres session.

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
+        positions = positions.where(positions != 0, pd.NA)
+
+        # Create deltas
+        deltas = positions.diff()
+        deltas.iloc[0] = positions.iloc[0]
+
+        # Create NaNs of the same size as deltas
+        share_price_on_increases = pd.DataFrame(data=pd.NA, index=deltas.index, columns=deltas.columns)
+
+        # Replace positive deltas with share_price
```
sentilesdal, line 141: positive or negative?  looks like you're matching on negative.

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
-
-            marginal_trades = pd.DataFrame(row).T
-
-            # pass in one trade at a time, and get out the updated wallet
-            agent_wallets[agent] = get_wallet_from_onchain_trade_info(
-                address=agent,
-                trades=marginal_trades,
-                index=agents.index(agent),
-                add_to_existing_wallet=agent_wallets[agent],
+Info = NamedTuple(
```
sentilesdal, line 116: nice, but can we call it something specific like `AgentPlottingInfo`?  It looks like this is specific to plotting, not a generic `AgentInfo`?

examples/hackweek_demo/plot_pnl.py
```diff
@@ -112,93 +113,134 @@ def get_wallet_from_onchain_trade_info(
...
-            for _, short in agent_wallets[agent].shorts.items():
-                pnl += float(short.balance) * (1 - spot_price)
-
-            new_pnl[f"{agent_shortname}"] = pnl
-        pnl_data.append(new_pnl)
-        time_data.append(row["timestamp"])
+            pnl = 0
+            for position_name in positions.columns:
+                if position_name.startswith("LP"):
+                    # LP value (TODO: check why certain agents own more than 100% of the pool)
```
sentilesdal, line 209: whhaaat

sentilesdal (commented):
lotta small things.  I actually might need you to run me through the pnl calc, couldn't quite parse the code.  a docstring at the top of the method would be 💯

sentilesdal (commented):
lotta small things.  I actually might need you to run me through the pnl calc, couldn't quite parse the code.  a docstring at the top of the method would be 💯

elfpy/data/acquire_data.py
slundqui: I can't speak to getting checkpoints, but generally I found that anything querying the chain using web3 sometimes fails without reason, and retrying typically fixes it

elfpy/data/postgres.py
slundqui: Once again, I can't speak to checkpointing, but the idea from other functions allowing negative indices was to match python list notation for accessing lists. E.g., [0, 1, 2, 3] [-1] = 3. The immediate use case is that, e.g., I want only the last 1000 blocks from the db. I can do this by passing in -1000 for the start block.

examples/hackweek_demo/run_demo.py
slundqui: I think a function here is fine, this encapsulates the "ticker" panel that happens to be relatively simple in logic. Regardless, seeing the caller call "get_ticker" before displaying it seems better to me than the one liner.

tests/data/test_checkpoint.py
```diff
@@ -0,0 +1,129 @@
+"""CRUD tests for CheckpointInfo"""
```
slundqui, line 1: Rebase from main to remove these changes please

elfpy/data/postgres.py
```diff
@@ -302,6 +339,46 @@ def get_pool_info(session: Session, start_block: int | None = None, end_block: i
...
+
+    This includes
+    - `sharePrice` : The share price of the first transaction in the checkpoint.
+    - `longSharePrice` : The weighted average of the share prices that all longs in the checkpoint were opened at.
+    - `shortBaseVolume` : The aggregate amount of base committed by LPs to pay for bonds sold short in the checkpoint.
+
+    Arguments
+    ---------
+    session : Session
+        The initialized session object
```
slundqui, line 353: The docstring here doesn't seem to match the function signature

examples/hackweek_demo/plot_pnl.py
wakamex: interesting, don't we need that to measure virtual reserves? I'll try and have a look.

wakamex: oh I think their bond_reserves variable in solidity includes the virtual reserves baked in
so solidity's
`spot_price = (initial_share_price * share_reserves / bond_reserves) ** time_stretch`
is the same as python's
`spot_price = ((initial_share_price * share_reserves) / (bond_reserves + lp_total_supply)) ** time_remaining_stretched`

examples/hackweek_demo/plot_pnl.py
wakamex: kek forgot that exists

examples/hackweek_demo/plot_pnl.py
wakamex: they used to, the tripwire isn't getting triggerred anymore

examples/hackweek_demo/plot_pnl.py
wakamex: that and the kitchen sink

examples/hackweek_demo/run_demo.py
wakamex: also, we may want to have more complicated ticker behaviour in the future

examples/hackweek_demo/plot_pnl.py
wakamex: it does the opposite of what it says. it doesn't REPLACE `where`... it KEEPS `where`

so I guess you're supposed to read it as
> equals `share_price_on_increases` where `matches clause` otherwise `share_price`

I guess there's some logic to this, but I definitely read it backwards the first 50 times. it doesn't help that [np.where](https://numpy.org/doc/stable/reference/generated/numpy.where.html) although quite different is used to select and replace what matches the clause like: `share_price_on_increases.loc[np.where(deltas > 0)] = share_price`

examples/hackweek_demo/plot_pnl.py
wakamex: replacing the wallet info in the dict[agent, wallet_info] with dict[agent, lots_more_stuff_than_that]

examples/hackweek_demo/plot_pnl.py
wakamex: I just made this mandatory

examples/hackweek_demo/plot_pnl.py
wakamex: class is in with `ap = AgentPosition()`

tests/data/test_checkpoint.py
wakamex: which changes specifically? is there something messed up from the rebase? or all good now?

elfpy/data/postgres.py
wakamex: ruff concurs: `elfpy/data/postgres.py:99:5: D417 Missing argument description in the docstring: `table_name``

elfpy/data/acquire_data.py
```diff
@@ -256,6 +261,26 @@ def main(
...
 
+                # keep querying until it returns to avoid random crashes with ValueError on some intermediate block
+                block_checkpoint_info = None
+                for _ in range(RETRY_COUNT):
+                    try:
+                        checkpoint_info_dict = hyperdrive_interface.get_hyperdrive_checkpoint_info(
+                            web3, hyperdrive_contract, block_number
+                        )
+                        # Set defaults
+                        for key in db_schema.CheckpointInfo.__annotations__:
```
slundqui, line 272: Can we move this logic outside of the try catch? The try catch should really only be over the chain query. Eventually, we should move this logic to hyperdrive_interface

examples/hackweek_demo/run_demo.py
```diff
@@ -70,8 +67,15 @@ def get_ticker(data):
 
 
 while True:
+    pool_config_data = postgres.get_pool_config(session)
     txn_data = postgres.get_transactions(session, start_block=start_block)
     pool_info_data = postgres.get_pool_info(session, start_block=start_block)
+    checkpoint_info = postgres.get_checkpoint_info(session, start_block=start_block)
+
+    assert max(pool_info_data.index) == max(checkpoint_info.index), (
```
slundqui, line 75: I think we should handle this case a bit more gracefully. I'm assuming you're worried about a misalignment of pool info data and checkpoint info data since acquire_data.py writes these fields sequentially. I think that if this is the case, lets just throw out any information that doesn't exist in both tables. This way, the streamlit doesn't crash due to this assertion, and instead can continue on.

examples/hackweek_demo/plot_pnl.py
sentilesdal: good catch

examples/hackweek_demo/run_demo.py
wakamex: I was thinking of implementing a more complicated approach with re-querying, but after discussing with you, let's just go with the simple "throw out the data" approach, since we're running non-stop queries in Streamlit, increasing update speed is actually a nice benefit. let me see if I can do that quick.

elfpy/data/postgres.py
sentilesdal: gotcha, thanks for explanation

examples/hackweek_demo/plot_pnl.py
sentilesdal: cool thanks

examples/hackweek_demo/plot_pnl.py
sentilesdal: phew

examples/hackweek_demo/run_demo.py
wakamex: implemented in c5e85d1580c60fb20190b2f9861fac14fc946641

elfpy/data/acquire_data.py
sentilesdal: this makes me think we should add retries to our generic smart_contract_read_call instead of replicating it everywhere.

elfpy/data/acquire_data.py
wakamex: Now that I see these lines side-by-side, it seems the obvious choice would be to set defaults either in the dataclass itself or in the `add_item` classes inside postgres, where we have access to db_schema stuff no problem:

```py
for key in db_schema.CheckpointInfo.__annotations__:
    if key not in checkpoint_info_dict:
        checkpoint_info_dict[key] = None
if block_checkpoint_info := db_schema.CheckpointInfo(**checkpoint_info_dict):
    postgres.add_checkpoint_infos([block_checkpoint_info], session)
```

wakamex: I'll just move them outside of the try/except in this PR, and we can re-org them in a future PR.

slundqui (approved):
This looks good to me, but I'd double check with @sentilesdal for the pnl calculation stuff

examples/hackweek_demo/plot_pnl.py
```diff
@@ -1,204 +1,206 @@
...
-     -------
-     Wallet
-         Wallet with Short, Long, and LP positions.
+from sqlalchemy.orm import Session
+
+from elfpy.data import postgres as pg
+
+
+@dataclass
+class AgentPosition:
```
sentilesdal, line 15: this class is so well documented, thanks!

wakamex:
aaand it's merged

examples/hackweek_demo/plot_pnl.py
```diff
@@ -1,204 +1,206 @@
...
+    Arguments
+    ---------
+    pool_config : PoolConfig
+        Configuration with which the pool was initialized.
+    pool_info : pd.DataFrame
+        Reserves of the pool at each block.
+    checkpoint_info : pd.DataFrame
+    Checkpoint information at each block.
+    """
+    session = pg.initialize_session()
```
sentilesdal, line 144: this should be passed in.  even better, agent_positions should be passed in.  We should talk about this to clarify but  very few functions should know about postgres.  There should be specific functions that go get a thing from postgres.  If a function has 'calculate' in the name that's a signal that its just doing math things, it shouldn't fetch data.

sentilesdal (approved):
much nicer!  let's remove the postgres dependency in calculate_pnl and write some tests for it in a follow-up pr.

examples/hackweek_demo/plot_pnl.py
wakamex: moved it to postgres, cus it's really a data manipulation task, and matches this pattern in run_demo.py:
```py
    pool_config_data = postgres.get_pool_config(session)
    txn_data = postgres.get_transactions(session, start_block=start_block)
    pool_info_data = postgres.get_pool_info(session, start_block=start_block)
    checkpoint_info = postgres.get_checkpoint_info(session, start_block=start_block)
    agent_positions = postgres.get_agent_positions(session)
```

PR #684, opened 2023-07-19
update docker python to 3.10

wakamex:
also smaller image size at 1.2GB instead of 1.8GB by switching to python-slim


slundqui (approved):

wakamex:
also smaller image size at 1.2GB instead of 1.8GB by switching to python-slim

PR #686, opened 2023-07-19
Improve error handling for hyperdrive contract

sentilesdal:
Previously custom errors raised from the contracts looked like 'CustomError(0xDEADBEEF)'  which were not very useful.  I added a list of known custom errors from the hyperdrive ecosystem and added a helper function to return the name of the error.  logging was added when those exceptions are caught.  logging also added for ContractLogicErrors as well.


dpaiton (approved):
woo!!

PR #666, opened 2023-07-13
set up config and main script for replacing evm_bots

dpaiton:
Sets up the `main.py` script outlined in #598 .

This is an incomplete PR. The last major TODO remaining is getting the agent wallets for each trade. However, likely once that is done there will be a good amount of cleanup required.

One side-quest I went on was removing the dependency of `db_schema` from the `eth` folder.

This PR results in several new action items, marked as TODOS. Several reference `evm_bots`, which can be searched for, so I will leave them out. I'm the rest here for easy reference while writing the next PR. They are:

`elfpy/agents/agent.py`: 
- TODO: wallet_address is no longer unique with the eth_bots setup and therefore needs to be deprecated in favor of a real checksum address

`elfpy/data/acquire_data.py`:
- TODO: Rename this to something more accurate to what is happening, e.g. decode_hyperdrive_transactions
- TODO: fewer nested blocks!

`elfpy/data/db_schema.py`:
- TODO: Rename this to something more accurate to what is happening, e.g. HyperdriveTransactions

`elfpy/eth/web3_setup.py`:
- TODO: Check that the user is running on anvil, raise error if not

`elfpy/hyperdrive_interface/hyperdrive_interface.py`:
- TODO: Would it be safe to assume that earliest_block.timestamp always equals zero?
- TODO: get agent wallet from receipts, similar to acquire_data.get_wallet_info

`elfpy/time/time.py`:
- TODO: Rename these to e.g. SECONDS_PER_YEAR

`examples/eth_bots/execute_agent_trades.py`:
- TODO: Fix up this function to allow pylint checks
- TODO: update agent wallet
- TODO: allow for min_output

`examples/eth_bots/main.py`:
- TODO: Move much of this out of main
- TODO: if provider.auto_mine is set then run the `mine` function
- TODO: deliver crash report

`examples/eth_bots/setup_agents.py`:
- TODO: raise issue on failure if `rpc_response` and `tx_receipt` indicate that the `set_anvil_account_balance` and `smart_contract_transact` fail, respectively.

We also need to:
- write the examples/eth_bots/README.md file
- fix up db_schema; I moved functions to there that should live elsewhere.
- test `examples/eth_bots/custom_policies/example_custom_policy.py`

PR #690, opened 2023-07-19
fix: use EnvironmentConfig in bot profiles

wakamex:


elfpy/bots/create_default_profile.py
```diff
@@ -1,9 +1,9 @@
 """Example script for writing a custom bot config to json"""
-from elfpy.bots import DEFAULT_USERNAME, BotConfig
+from elfpy.bots import DEFAULT_USERNAME, EnvironmentConfig
```
slundqui, line 2: My assumption was that this file is now obsolete in favor of `config.py` in examples/eth_bots

slundqui: @dpaiton is this right?

wakamex: we replaced BotConfig with EnvironmentConfig in #666 and I haven't seen anything to indicate otherwise since

slundqui: This is just through looking at and running `main.py` in `eth_bots`. I guess this makes sense if we still want to support `evm_bots` through this patch

sentilesdal (approved):

PR #691, opened 2023-07-19
Add `decode_error_selector_for_contract`

sentilesdal:
Adds a helper to decode an error selector for a contract.  Just realized that this could be even more generalized for function selectors as well but I'll leave that for another PR.

### DIFF
--- a/elfpy/eth/errors/errors.py
+++ b/elfpy/eth/errors/errors.py
@@ -1,10 +1,10 @@
 """Error handling for the hyperdrive ecosystem"""
-from typing import Literal, Sequence, TypedDict
 
 from eth_utils.conversions import to_hex
 from eth_utils.crypto import keccak
 from web3.contract.contract import Contract
-from web3.types import ABIFunctionParams
+
+from .types import ABIError
 
 
 def decode_error_selector_for_contract(error_selector: str, contract: Contract) -> str:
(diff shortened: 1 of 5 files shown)


elfpy/eth/errors/errors.py
```diff
@@ -0,0 +1,61 @@
...
+        print(f"{decoded_error_selector=}")
+        if decoded_error_selector == error_selector:
+            error_name = error.get("name")
+            break
+
+    return error_name
+
+
+# TODO: add this to web3.py
+class ABIError(TypedDict, total=True):
```
dpaiton, line 56: Can you put `class` defs at the top of the file or in their own file?

elfpy/eth/errors/errors.py
```diff
@@ -0,0 +1,60 @@
...
+        A web3.py Contract interface, the abi is required for this function to work.
+
+    Returns
+    -------
+    str
+       The name of the error. If the error is not found, returns UnknownError.
+    """
+
+    abi = contract.abi
+    if not abi:
```
dpaiton, line 29: I think we want to explicitly say `if abi is None` to avoid unexpexted truthy issues

dpaiton (approved):
sick

slundqui (approved):
LGTM

wakamex (approved):

PR #695, opened 2023-07-19
Username handling in streamlit demo

slundqui:


elfpy/data/postgres.py
```diff
@@ -806,6 +806,27 @@ def __init__(self, wallet_history: pd.DataFrame):
...
+        The initialized session object
+    filter_addr : list[str] | None
+        Only return these addresses. Returns all if None
+
+    Returns
+    -------
+    dict[str, AgentPosition]
+        Returns a dictionary keyed by wallet address, value of an agent's position
+    """
+    if filter_addr is None:
```
dpaiton, line 824: nit: I think a cleaner pattern is
```
if condition:
    return thing
return else_thing
```

dpaiton (approved):

PR #697, opened 2023-07-20
fix budget

wakamex:


sentilesdal (approved):

elfpy/bots/budget.py
```diff
@@ -17,10 +17,26 @@ class Budget:
...
     def sample_budget(self, rng: NumpyGenerator) -> FixedPoint:
-        """Returns a sample from a clipped normal distribution"""
+        """Returns a sample from a clipped normal distribution.
+
+        Parameters
+        ----------
+        rng : NumpyGenerator
+            Random number generator
+
+        Note
```
dpaiton, line 27: note should go below Returns

Parameters should be Arguments

elfpy/bots/budget.py
```diff
@@ -17,10 +17,26 @@ class Budget:
...
+        Sample from normal distribution with mean of mean_wei and standard deviation of std_wei.
+        Then clip to between a minimum of min_wei and a maximum of max_wei.
+
+        Returns
+        -------
+        FixedPoint
+            The resulting sample
+        """
         return FixedPoint(
-            FixedPointMath.clip(
```
dpaiton, line 22: pls change to
```
        return FixedPoint(
            scaled_value=FixedPointMath.clip(
                int(np.round(rng.normal(loc=self.mean_wei, scale=self.std_wei))),
                self.min_wei,
                self.max_wei,
            )
        )
```

dpaiton (approved):
comments to fix then good 👍

elfpy/bots/budget.py
wakamex: I was looking at the [Numpy style guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) last night. It seems to say Parameters, though I just realized we've been using Arguments this entire time. Didn't we want to follow the Numpy style?

for notes, I was looking for something to include below the first sentence. looks like they allow a paragraph there without a title. I guess that's what they refer to as the [Extended Summary](https://numpydoc.readthedocs.io/en/latest/format.html#extended-summary) which is meant to "clarify functionality, not to discuss implementation detail or background theory, which should rather be explored in the [Notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) section below"

wakamex: changed it to Arguments and using Extended Summary instead of Notes in 2c19b16decdd183149727b1b604d0a0a77a96cee

PR #700, opened 2023-07-20
better allowance logging

wakamex:


examples/evm_bots.py
```diff
@@ -304,15 +304,30 @@ def do_trade(
...
-        txn_args = (
-            hyperdrive_instance.address,
-            FixedPoint("50_000.0", decimal_places=18).scaled_value,
-        )
-        ape_utils.attempt_txn(agent_contract, base_instance.approve, *txn_args)
-        logging.info("Trade had insufficient allowance, approving an additional 50k base.")
+    # If agent does not have enough base approved for this trade, then approve
+    # an additional amount equal to double of the shortfall, or 50k, whichever is greater.
+    # the ABI for allowance is: allowance(address_owner, address_spender) → uint256
+    initial_allowance = base_instance.allowance(agent_contract.address, hyperdrive_instance.address) / 1e18
```
sentilesdal, line 310: honestly we should set to max allowance, allowance is just noise for the sake of simulations.

sentilesdal: land this if you want.  if you change to max allowance in the setup then use the `eth_utils` package to get the MAX_INT value.  In fact, that whole package is worth a read through to make sure we aren't reinventing things that exist already.

dpaiton (approved):

examples/evm_bots.py
wakamex: added in b58454fd201fa4550b926f32be6641ed99d32c65

PR #702, opened 2023-07-20
Ease linting on test files

sentilesdal:
Some of these don't make sense for test files.  Test functions don't need sphinx docs, and often don't have any public methods etc.


wakamex (approved):
I assume this works

PR #703, opened 2023-07-20
streamline CI and remove ape

wakamex:
- main is now the tag for every docker image built on main
- testing is now the tag for every docker image built on any other branch (these builds are not automatically triggerred)
- be careful with the edge tag as it will just pick the latest of either of these two


elfpy/markets/hyperdrive/hyperdrive_pricing_model_sol.py
```diff
@@ -46,6 +44,8 @@ def calculate_spot_price(
     FixedPoint
         The spot price of bonds in terms of base as an 18 fixed-point value.
     """
+    if share_reserves <= FixedPoint(0):
```
sentilesdal, line 47: would rather let this go negative so we can debug

sentilesdal (approved):
let's remove that NaN coercion.  would rather see negative values and then debug.  calculate_spot_price is just a pure math function, the responsibility to ensure proper values lies upstream

wakamex:
replaced by #716 #717 and #718

PR #704, opened 2023-07-20
loosen linting in ruff

wakamex:


dpaiton (approved):

sentilesdal (changes requested):
I'm not sure about these ignores.  Should have a discussion with the team

sentilesdal (approved):
this is fine if just for ruff.  if this becomes our linter for ci then we'll have to discuss

PR #707, opened 2023-07-20
Register bot server

slundqui:
Creates a flask server that opens an endpoint to register wallet addresses to a username on the server side. Adds calling of this server in eth_bots to register usernames to wallet addresses.

Overall, this alleviates adding credentials to eth bots to postgres

### DIFF
--- a/elfpy/bots/environment_config.py
+++ b/elfpy/bots/environment_config.py
@@ -21,6 +21,8 @@ class EnvironmentConfig(types.FrozenClass):
 
     # Logical username for who is running bots
     username: str = DEFAULT_USERNAME
+    # Logical username for who is running bots
+    username_register_url: str = "http://localhost:5001"
     # whether to run on alchemy
     alchemy: bool = False
     # url for retrieving the contract artifacts
--- a/elfpy/data/register_username_server.py
+++ b/elfpy/data/register_username_server.py
@@ -9,7 +9,7 @@ app = Flask(__name__)
 
 @app.route("/register_bots", methods=["POST"])
 def register_bots():
-    """Run a python script and return the script id."""
+    """Registers a list of wallet addresses to a username via post request"""
     # TODO: validate the json
     data = request.json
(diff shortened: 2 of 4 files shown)


elfpy/bots/environment_config.py
```diff
@@ -21,6 +21,8 @@ class EnvironmentConfig(types.FrozenClass):
 
     # Logical username for who is running bots
     username: str = DEFAULT_USERNAME
+    # Logical username for who is running bots
```
sentilesdal, line 24: copy pasta on comment, update plz

dpaiton (approved):

elfpy/data/register_username_server.py
```diff
@@ -0,0 +1,37 @@
...
+@app.route("/register_bots", methods=["POST"])
+def register_bots():
+    """Registers a list of wallet addresses to a username via post request"""
+    # TODO: validate the json
+    data = request.json
+
+    # initialize the postgres session
+    session = postgres.initialize_session()
+    try:
+        # Typing doesn't work with request objects
```
sentilesdal, line 19: It does if you make a type ;)  you can do something like this:
```python
from flask import Flask, request
from flask_expects_json import expects_json

app = Flask(__name__)

# Define the JSON schema for the expected request payload
json_schema = {
    'type': 'object',
    'properties': {
        'wallet_addrs': {'type': 'string'},
        'username': {'type': 'string'},
    },
    'required': ['username', 'wallet_addrs']
}

@app.route('/register_bots', methods=['POST'])
@expects_json(json_schema)  # Decorator to validate the request against the JSON schema
def register_bots():
    # Access the validated JSON payload using request.json
    data = request.json

    # Perform further processing using the validated data
    username = data['username']
    wallet_addrs = data['wallet_addrs']

    ...

    return out

    ```

sentilesdal (commented):
already merged but see comments

PR #711, opened 2023-07-21
Add guard rails to max long/short calculations.

sentilesdal:
These basically needed some guard rails on them.  RandomAgent runs now.  This PR introduces and fixes a new bug, I've changed the agent to not return an open long of 0, instead returning an empty action list.  This causes main.py to get stuck when we are in automine because no new trades are submitted, so we never get a new block and thus never try to execute trades again.  I now check if we are in automine to determine if we should wait for a new block or not.

### DIFF
--- a/examples/eth_bots/main.py
+++ b/examples/eth_bots/main.py
@@ -94,30 +94,30 @@ def main():  # TODO: Move much of this out of main
         if latest_block_number is None or latest_block_timestamp is None:
             raise AssertionError("latest_block_number and latest_block_timestamp can not be None")
 
-        if latest_block_number > last_executed_block:
-            # log and show block info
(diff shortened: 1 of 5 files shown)


examples/eth_bots/main.py
```diff
@@ -114,6 +120,32 @@ def main():  # TODO: Move much of this out of main
                 # TODO: deliver crash report
 
 
+def get_wait_for_new_block(web3: Web3) -> bool:
```
dpaiton, line 123: Awesome! Can you clear the TODO on line 113? `# TODO: if provider.auto_mine is set then run the mine function`

dpaiton (approved):
siiick

PR #717, opened 2023-07-21
Lint in docker

wakamex:
- image is smaller (1.06GB from 1.16GB)
- workflows are faster:
- lint in 33s down from 1m 13s
- static in 52s down from 1m 40s
- test in 56s down from 2m 2s
- no change to docker tags


wakamex:
this passes with an image built in the new way
tested with images build that way, tagged with the "testing" tag in #703

wakamex:
will update after #727

PR #718, opened 2023-07-21
add testing docker tag

wakamex:
- this is only applied to images built on branches OTHER than main
- these are only rarely constructed for testing, and are triggerred manually
- edge tags will still be applied to every commit ON main


sentilesdal:
so everything not main gets the testing tag?

wakamex:
> so everything not main gets the testing tag?

yes, but that happens only if you intentionally go out of your way for it to happen
for example, see this commit which I immediately reverted after it built a "testing" image: [`80cf5b7` (#717)](https://github.com/delvtech/elf-simulations/pull/717/commits/80cf5b75bf228e2f5714e79639c869d916d31aec)

PR #719, opened 2023-07-21
specific error instead of div by zero

wakamex:


elfpy/markets/hyperdrive/hyperdrive_pricing_model_sol.py
```diff
@@ -46,8 +44,9 @@ def calculate_spot_price(
     FixedPoint
         The spot price of bonds in terms of base as an 18 fixed-point value.
     """
-    spot_price = (initial_share_price * share_reserves / bond_reserves) ** time_stretch
-    return spot_price
+    if bond_reserves <= FixedPoint(0):
```
sentilesdal, line 47: can't land this.  any file that ends in sol.py ought to mirror the solidity, per the function comment above.  move this up a level.

sentilesdal (changes requested):
move bond_reserves check out of sol.py file

wakamex:
addressed in #711

PR #722, opened 2023-07-22
Fixing streamlit for a case where LP doesn't exist

slundqui:


dpaiton (approved):

examples/hackweek_demo/run_demo.py
```diff
@@ -186,10 +186,11 @@ def username_to_address(lookup: pd.DataFrame, selected_list: list[str]) -> list[
     with lp_plot_placeholder.container():
         ax_lp_token.clear()
         # TODO extract this out
-        num_lp_tokens = [ap.positions.loc[:, "LP"] for ap in agent_positions.values()]
+        num_lp_tokens = [ap.positions.loc[:, "LP"] for ap in agent_positions.values() if "LP" in ap.positions.columns]
+        addrs = [addr for addr, ap in agent_positions.items() if "LP" in ap.positions.columns]
```
wakamex, line 190: doesn't every agent have the same columns because they're sliced from the same pivot?

slundqui: This is for the case where no wallets had LP tokens

PR #723, opened 2023-07-22
polish for readability

wakamex:


wakamex:
move to https://github.com/wakamex/elf-simulations/pull/9

PR #724, opened 2023-07-22
fixes config errors

dpaiton:


slundqui (approved):
Thanks!

elfpy/bots/environment_config.py
```diff
@@ -20,12 +20,12 @@ class EnvironmentConfig(types.FrozenClass):
 
     # Logical username for who is running bots
     username: str = DEFAULT_USERNAME
-    # hostname for container URLs
-    hostname: str = "http://localhost"
-    # ports for container URLs
-    artifacts_port: str = "80"
-    rpc_port: str = "8545"
-    user_registry_port: str = "5002"
+    # Logical username for who is running bots
```
sentilesdal, line 23: now that i look at this more, it feels weird to have these defaults in here.  leave em for now but might be good to leave the urls unopinionated in the definition and have like default_local_config, default_aws_config etc that we use in our examples.

dpaiton: yeah we need to fix up config in general. This should definitely be on the todo list.

PR #731, opened 2023-07-22
Added a checkpointing bot for Hyperdrive

jalextowle:
Hyperdrive is relatively low maintenance; however, it's important to mint every checkpoint for each Hyperdrive pool to ensure that matured positions are closed at the correct share price. This bot is a minimum crappy product that has been lightly tested (by production standards) and mints new checkpoints successfully when started up alongside the Hyperdrive compose app. 

Every time I've written a bot like this, they have hit unforeseen issues when interacting with production Ethereum networks. I hope that this will be reasonably reliable on devnet; however, this is going to be a work in progress until it becomes battle-hardened after a lot of use.


elfpy/bots/checkpoint_bot.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def does_checkpoint_exist(hyperdrive: Contract, checkpointTime: int) -> bool:
+    return smart_contract_read(hyperdrive, "getCheckpoint", checkpointTime)["sharePrice"] > 0
+
+
+def main() -> None:
+    # Get the configuration and initialize the web3 provider.
+    config = get_config()
+    provider = eth.web3_setup.initialize_web3_with_http_provider(config.rpc_url, reset_provider=False)
```
dpaiton, line 56: We've been calling this `web3` elsewhere. The web3 docs seem to oscillate randomly between `w3` and `web3`, but we like the latter because it conforms to pep8 linting standards. The "provider" is a communication component (how web3 talks to the chain). The thing returned is a `Web3` object, which is not given a name afaik, but I guess could be called a "container" or "wrapper". More here: https://web3py.readthedocs.io/en/latest/internals.html#web3-internals

elfpy/bots/checkpoint_bot.py
```diff
@@ -0,0 +1,125 @@
...
+    # every checkpoint after a waiting period. It will poll very infrequently
+    # to reduce the probability of needing to mint a checkpoint.
+    config = get_hyperdrive_config(hyperdrive)
+    checkpoint_duration = config["checkpointDuration"]
+    while True:
+        # Get the latest block time and check to see if a new checkpoint should
+        # be minted. This bot waits for a portion of the checkpoint to reduce
+        # the probability of needing a checkpoint. After the waiting period,
+        # the bot will attempt to mint a checkpoint.
+        timestamp = provider.eth.get_block("latest").timestamp  # type: ignore
```
dpaiton, line 94: if you want to avoid the `type: ignore` you need to add a default condition. e.g.:
```
latest_block = provider.eth.get_block("latest")
timestamp = latest_block.get("timestamp", None)
if timestamp is None:
    raise AssertionError(f"{latest_block=} has no timestamp")
```

elfpy/bots/checkpoint_bot.py
```diff
@@ -0,0 +1,125 @@
...
+        checkpoint_portion_elapsed = timestamp % checkpoint_duration
```
dpaiton, line 95: The `timestamp` variable is of type `Timestamp` which is a `NewType("Timestamp", int)`. Checkpoint duration is also an integer. [Integer modulo can behave differently from float](https://realpython.com/python-modulo-operator/#python-modulo-operator-basics). This has burned us in the past, so I just wanted to bring it to your attention. To be safe & consistent with elsewhere in the codebase, you might want to convert everything to `FixedPoint` before proceeding.

We also do intend to update the `hyperdrive_interface` functions to return FixedPoint for all values, instead of a mixture of floats and ints as it does now. The current behavior is from an oversight.

elfpy/hyperdrive_interface/hyperdrive_addresses.py
```diff
@@ -11,5 +11,7 @@ class HyperdriveAddresses:
     # pylint: disable=too-few-public-methods
 
     base_token: str = attr.ib()
+    factory: str = attr.ib()
```
dpaiton, line 14: Can you check that these changes don't break the fetch function here:
https://github.com/delvtech/elf-simulations/blob/main/elfpy/hyperdrive_interface/hyperdrive_interface.py#L28

We've fallen behind on tests in the last week.

dpaiton (approved):

elfpy/bots/checkpoint_bot.py
jalextowle: Ok, I'm fine with moving back to `web3`. I called it `provider` since we use it like a provider (this is how we make RPC requests and interact with the ethereum node), but I see your point.

PR #733, opened 2023-07-24
reverts to linting on all files & no docs linting

dpaiton:


.github/workflows/lint.yml
```diff
@@ -36,18 +36,18 @@ jobs:
       - name: Run black
         run: python -m black --check --diff .
 
-      - name: Get changed Python files
-        id: changed_files
+      - name: Get all Python files
+        id: list_files
         run: |
-          echo "files=$(git diff --name-only HEAD^ HEAD | grep '.py$' | xargs)" >> $GITHUB_OUTPUT
+          echo "::set-output name=files::$(git ls-files '*.py' '*.pyi')"
```
wakamex, line 42: set-output is deprecated in favor of `echo "{name}={value}" >> $GITHUB_OUTPUT` https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

dpaiton: hmm ok I tried that in the first commit and it failed. I'll try again.

.github/workflows/test.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           token: ${{ secrets.GH_GITHUB_COM_TOKEN || github.token}}
       - run: |
           python -m pip install --upgrade pip
-          python -m pip install .[with-dependencies,postgres,ape]
+          python -m pip install .[with-dependencies,postgres]
```
wakamex, line 50: we no longer have a postgres dependency section. should we just add it to `with-dependencies`?

.github/workflows/lint.yml
dpaiton: yeah it is failing now:
```
Run echo "files=$(git ls-files '*.py' '*.pyi')" >> $GITHUB_OUTPUT
Error: Unable to process file command 'output' successfully.
Error: Invalid format 'elfpy/__init__.py'
```

dpaiton: yeah it's a whole thing :( 
https://stackoverflow.com/questions/74137120/how-to-fix-or-avoid-error-unable-to-process-file-command-output-successfully

dpaiton: xargs worked

sentilesdal (approved):

PR #735, opened 2023-07-24
more pythonic logs with per-handler log levels

wakamex:
- set log level at the handler level, to allow separate levels
- clarify setup_logging should only be called once, at initialization
- for custom logging, use add_stdout_handller and add_file_handler
- sane defaults in eth_bots reduces spam from libraries:
	- urllib3 logging set to WARNING
    - web3 logging set to WARNING
    - filter web3 warnings about base_contract `MismatchedABI(The event signature did not match the provided ABI). It has been discarded.`

### DIFF
--- a/elfpy/utils/logs.py
+++ b/elfpy/utils/logs.py
@@ -5,7 +5,6 @@ import json
 import logging
 import os
 import sys
-import warnings
 from logging.handlers import RotatingFileHandler
 
 from web3.exceptions import InvalidTransaction
@@ -56,10 +55,6 @@ def initialize_basic_logging(
         - Test the various optional input combinations
     """
     # pylint: disable=too-many-arguments
-    # sane defaults to avoid spam from libraries
-    logging.getLogger("urllib3").setLevel(logging.WARNING)
-    logging.getLogger("web3").setLevel(logging.WARNING)
-    warnings.filterwarnings("ignore", category=UserWarning, module="web3.contract.base_contract")
     # remove all handlers if requested
     if not keep_previous_handlers:
         _remove_handlers(get_root_logger())
--- a/elfpy/utils/logs.py
+++ b/elfpy/utils/logs.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import sys
+import warnings
 from logging.handlers import RotatingFileHandler
 
 from web3.exceptions import InvalidTransaction
@@ -14,7 +15,7 @@ import elfpy.utils.format as format_utils
 from elfpy.data.db_schema import Base, PoolConfig, PoolInfo
 
 
-def setup_logging(
+def initialize_basic_logging(
     log_filename: str | None = None,
     max_bytes: int | None = None,
     log_level: int | None = None,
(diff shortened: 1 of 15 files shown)


elfpy/utils/logs.py
```diff
@@ -19,39 +19,39 @@ def setup_logging(
     max_bytes: int | None = None,
     log_level: int | None = None,
     delete_previous_logs: bool = False,
-    log_stdout: bool = True,
```
dpaiton, line 22: leave this true if possible to avoid accidental problems downstream

dpaiton: or at a minimum double check the elfpy/examples/eth_bots endpoint to make sure we aren't changing the behavior from default logging to stdout

dpaiton (approved):

elfpy/utils/logs.py
```diff
@@ -23,81 +23,79 @@ def setup_logging(
...
     if delete_logs:
-        for handler in logging.getLogger().handlers:
+        for handler in root_logger.handlers:
             if hasattr(handler, "baseFilename") and not isinstance(handler, logging.StreamHandler):
                 # access baseFilename in a type safe way
                 handler_file_name = getattr(handler, "baseFilename", None)
                 if handler_file_name is not None and os.path.exists(handler_file_name):
                     os.remove(handler_file_name)
             handler.close()
-    # logging.getLogger().handlers = []
```
wakamex, line 96: why did we have this commented out before? did we want to not remove handlers on a `close_logging` call? My assumption here is that after this call, we'll be re-initializing logging from scratch, so we don't want to save anything. That's how the testing works.

elfpy/utils/logs.py
wakamex: left it as True

elfpy/utils/logs.py
```diff
@@ -23,81 +24,83 @@ def setup_logging(
...
-    handlers = logging.getLogger().handlers if keep_previous_handlers else []
-    # pipe to stdout if requested
-    if log_stdout:
-        stream_handler = logging.StreamHandler(sys.stdout)
-        stream_handler.setFormatter(log_formatter)
-        handlers.append(stream_handler)
-    # log to file
+    # pylint: disable=too-many-arguments
+    # sane defaults to avoid spam from libraries
+    logging.getLogger("urllib3").setLevel(logging.WARNING)
```
dpaiton, line 60:
```    logging.getLogger("urllib3").setLevel(logging.WARNING)
    logging.getLogger("web3").setLevel(logging.WARNING)
    warnings.filterwarnings("ignore", category=UserWarning, module="web3.contract.base_contract")```
    These should not happen here -- they should be in whatever script is being run. We want to be as unopinionated as possible about logging. Changing package levels "under the hood" is bad.

Comment (wakamex) [reply elfpy/utils/logs.py:60]: right, i'll move it to main.py, that's where I had it initially!

File: elfpy/utils/logs.py [hunk 4]
```diff
@@ -116,18 +114,97 @@ def _prepare_log_path(log_filename: str):
...
+    # Delete the log file if requested
+    if delete_previous_logs and os.path.exists(os.path.join(log_dir, log_name)):
+        os.remove(os.path.join(log_dir, log_name))
+    file_handler = _create_file_handler(
+        log_dir, log_name, _create_formatter(log_format_string), max_bytes, _create_log_level(log_level)
+    )
+    logger.addHandler(file_handler)
+
+
+def _remove_handlers(logger: logging.Logger):
```
dpaiton, line 182: might as well keep all of these public until we have a good reason to make them private

elfpy/utils/logs.py
```diff
@@ -116,18 +114,97 @@ def _prepare_log_path(log_filename: str):
...
+    while logger.handlers:
+        logger.removeHandler(logger.handlers[-1])
+
+
+def _create_file_handler(log_dir: str, log_name: str, log_formatter: logging.Formatter, max_bytes: int, log_level: int):
```
dpaiton, line 187: can you add proper docstrings for the new functions?

dpaiton (approved):

elfpy/utils/logs.py
wakamex: double-checked that `close` does not remove a handler. but I believe removing it is what we want.

PR #738, opened 2023-07-24
adds the ability to fund bots from user accounts

dpaiton:
Adds two new endpoints:
- `examples/eth_bots/populate_env.py`
This is used to return the arguments required to construct a `.env` file.  For example, here is mine:
```
export USER_KEY="0xSOMEKEY"
export RPC_URL="http://localhost:8545"
export ARTIFACTS_URL="http://localhost:80"
export BASE_ABI_FILE="/Users/dylan/code/element/elf-simulations/hyperdrive_solidity/.build/ERC20Mintable.json"
export AGENT_KEYS='["0xSOMEKEY"]'
export AGENT_BASE_BUDGETS='[3396163194603698651136]'
export AGENT_ETH_BUDGETS='[1000000000000000000]'
```

The bottom three values are populated from settings defined in `examples/eth_bots/eth_bots_config.py` by running `python populate_env.py >> .env`

Note that multiple bots should be encoded in JSON format. e.g. for two:
```
export AGENT_KEYS='["0xSOMEKEY", "0xSOMEOTHERKEY"]'
```
The single & double quote positions matters. The length of `AGENT_BASE_BUDGET` and `AGENT_ETH_BUDGET` must equal the length of `AGENT_KEYS`.

- `examples/eth_bots/fund_bots.py`
This is used to fund the bots from the user wallet. As long as the `.env` variable is set correctly, it'll pull money from the user's account and transfer it to the bot wallet addresses.

The other major change is to `get_agent_accounts.py`, where we populate the agent budgets with the `.env` values. There is a conditional in here:
```
            if len(agent_base_budgets) >= agent_count:
                kwargs["budget"] = FixedPoint(scaled_value=agent_base_budgets[agent_count])
            else:
                kwargs["budget"] = agent_info.base_budget.sample_budget(rng)
                # TODO: This is where we would fund the bots if we wanted to mint money from nothing
```
That is left to remind us where we should change up the code to generate bots that don't need user funding. This is a scenario that we will want for testing later, but I am leaving out during the trading competition to prevent users from accidentally minting infinite money for their bots.


sentilesdal (approved):

sentilesdal:
looks fine.  could use some descriptions in the functions for a quick rundown.

PR #739, opened 2023-07-25
Bots now use async waits on tx reciept

slundqui:
This fixes the bug of constraining one trade per eth_bots script

### DIFF
--- a/elfpy/eth/transactions.py
+++ b/elfpy/eth/transactions.py
@@ -70,6 +70,22 @@ async def async_wait_for_transaction_receipt(
     """Async version of wait_for_transaction_receipt
     This function is copied from `web3.eth.wait_for_transaction_receipt`, but using a non-blocking wait
     instead of a blocking wait
+
+    Arguments
+    ---------
+    web3: Web3
+        web3 provider object
+    transaction_hash: HexBytes
+        The hash of the transaction
+    timeout: float
+        The amount of time in seconds to time out the connection
+    poll_latency: float
+        The amount of time in seconds to wait between polls
+
+    Returns
+    -------
+    TxReceipt
+        The transaction receipt
     """
     try:
         with Timeout(timeout) as _timeout:
@@ -83,11 +99,10 @@ async def async_wait_for_transaction_receipt(
                 await _timeout.async_sleep(poll_latency)
         return tx_receipt
 
-    except Timeout:
-        # pylint: disable=raise-missing-from
+    except Timeout as exc:
         raise TimeExhausted(
             f"Transaction {HexBytes(transaction_hash) !r} is not in the chain " f"after {timeout} seconds"
-        )
+        ) from exc
 
 
 async def async_smart_contract_transact(
--- a/elfpy/eth/transactions.py
+++ b/elfpy/eth/transactions.py
@@ -90,6 +90,7 @@ async def async_wait_for_transaction_receipt(
     try:
         with Timeout(timeout) as _timeout:
             while True:
+                await _timeout.async_sleep(poll_latency)
                 try:
                     tx_receipt = web3.eth.get_transaction_receipt(transaction_hash)
                 except TransactionNotFound:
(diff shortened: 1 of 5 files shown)


elfpy/eth/transactions.py
```diff
@@ -62,6 +64,89 @@ def smart_contract_read(contract: Contract, function_name: str, *fn_args, **fn_k
     return {f"value{idx}": value for idx, value in enumerate(return_values)}
 
 
+async def async_wait_for_transaction_receipt(
+    web3: Web3, transaction_hash: HexBytes, timeout: float = 120, poll_latency: float = 0.1
+) -> TxReceipt:
+    """Async version of wait_for_transaction_receipt
+    This function is copied from `web3.eth.wait_for_transaction_receipt`, but using a non-blocking wait
+    instead of a blocking wait
```
dpaiton, line 72: can you add Arguments & Returns docs pls?

elfpy/eth/transactions.py
```diff
@@ -62,6 +64,89 @@ def smart_contract_read(contract: Contract, function_name: str, *fn_args, **fn_k
...
+                    tx_receipt = web3.eth.get_transaction_receipt(transaction_hash)
+                except TransactionNotFound:
+                    tx_receipt = None
+                if tx_receipt is not None:
+                    break
+                await _timeout.async_sleep(poll_latency)
+        return tx_receipt
+
+    except Timeout:
+        # pylint: disable=raise-missing-from
```
dpaiton, line 87: adding in the `from` is actually pretty nice, instead of disabling. It's also easy. Here's a solution:
https://realpython.com/python-raise-exception/#:~:text=You'll%20typically%20use%20from,__cause__%20attribute.

examples/eth_bots/execute_agent_trades.py
```diff
@@ -69,14 +71,23 @@ def transact_and_parse_logs(
...
+    if status == 0:
+        raise UnknownBlockError(f"Receipt has no status or status is 0 \n {tx_receipt=}")
+
     hyperdrive_event_logs = eth.get_transaction_logs(
         web3,
         hyperdrive_contract,
         tx_receipt,
         event_names=[fn_name[0].capitalize() + fn_name[1:]],
     )
+    # if len(hyperdrive_event_logs)
```
dpaiton, line 90: dead comment

examples/eth_bots/execute_agent_trades.py
```diff
@@ -98,53 +109,86 @@ def transact_and_parse_logs(
...
-                "AGENT %s to perform %s for %g",
-                str(account.checksum_address),
-                trade_object.trade.action_type,
-                float(trade_object.trade.trade_amount),
-            )
-            wallet_deltas = match_contract_call_to_trade(
+    # This condition is unlikely;
+    # it would happen if they created an eth address but didn't associate it with an elfpy Agent
+    if account.agent is None:
+        return
```
dpaiton, line 135: I feel like we might as well raise an assertion error here, right? The code assumes this is true throughout.

examples/eth_bots/execute_agent_trades.py
```diff
@@ -98,53 +109,86 @@ def transact_and_parse_logs(
...
+    Arguments
+    ---------
+    web3 : Web3
+        web3 provider object
+    hyperdrive_contract : Contract
+        Any deployed web3 contract
+    agent_accounts : list[EthAccount]
+        A list of EthAccount that are conducting the trades
+    """
+    # get latest market
```
dpaiton, line 174: Add a `# TODO: This might _not_ be the latest market, due to async`

dpaiton (approved):

examples/eth_bots/execute_agent_trades.py
slundqui: Currently, the code here will return gracefully. This function isn't returning anything, and instead, updating state variables internally. Still want to raise an assertion error here?

slundqui: Leaving as is for now

PR #745, opened 2023-07-25
add eth_transfer function

dpaiton:


wakamex (approved):

elfpy/eth/transactions.py
```diff
@@ -220,6 +222,47 @@ def smart_contract_transact(
         raise err
 
 
+def eth_transact(
```
sentilesdal, line 225: I'd rename this `eth_transfer`, lets' save `eth_transact` for a generic raw eth call function

sentilesdal (approved):
nice

PR #751, opened 2023-07-25
Using stretch time from pool config in fixed rate calculation

slundqui:


examples/hackweek_demo/extract_data_logs.py
```diff
@@ -26,15 +26,17 @@ def read_json_to_pd(json_file):
...
     """Calculate spot price from reserves stored in a state variable."""
     return calculate_spot_price(
         state.shareReserves,
         state.bondReserves,
         state.lpTotalSupply,
         maturity_timestamp,
         block_timestamp,
-        position_duration,
+        position_duration=config_data["positionDuration"],
+        stretch_time=config_data["invTimeStretch"],
```
wakamex, line 38: are you sure we want the inverted version here?

slundqui: i think so. the other one was around 22, where the invTimeStretch was closer to the value

wakamex: oh I see what we store as `invTimeStretch` is their `hyperdrive_config["timeStretch"]` that's correct

wakamex (approved):

examples/hackweek_demo/extract_data_logs.py
wakamex: looking at https://github.com/delvtech/elf-simulations/pull/751/files#diff-3a0341d06c9bd4cf543c2823827294caf1cb0188fadd925744a47390051d05b7R180

examples/hackweek_demo/extract_data_logs.py
```diff
@@ -45,16 +47,20 @@ def calculate_spot_price(
...
     """Calculate the spot price given the pool info data."""
     # pylint: disable=too-many-arguments
 
     # Hard coding variables to calculate spot price
-    initial_share_price = 1
-    time_remaining_stretched = 0.045071688063194093
-    full_term_spot_price = (
-        (initial_share_price * share_reserves) / (bond_reserves + lp_total_supply)
-    ) ** time_remaining_stretched
+    if initial_share_price is None:
```
dpaiton, line 57: These should be required w/o defaults.

dpaiton (approved):

PR #757, opened 2023-07-26
remove hyperdrive dependency

dpaiton:


slundqui (approved):
best pr ever

.env.sample
```diff
@@ -2,6 +2,3 @@ export USER_KEY="0xUSER_PRIVATE_KEY"
 export AGENT_KEYS='["0xAGENT_PRIVATE_KEY"]'
 export AGENT_BASE_BUDGETS='[3396163194603698651136]'
 export AGENT_ETH_BUDGETS='[1000000000000000000]'
-export RPC_URL="http://localhost:8545"
```
slundqui, line 5: what were these getting used for originally?

dpaiton: it was redundant from the environment_config object, so we don't need it in env.

PR #760, opened 2023-07-27
Remove slippage defaults from policies

sentilesdal:
We had several defaults for slippage tolerance throughout the repo.  Now, the slippage_tolerance is only set in the AgentConfig.

### DIFF
--- a/examples/eth_bots/custom_policies/example_custom_policy.py
+++ b/examples/eth_bots/custom_policies/example_custom_policy.py
@@ -36,10 +36,7 @@ class ExampleCustomPolicy(BasePolicy):
         else:
             self.trade_amount: FixedPoint = trade_amount
 
-        if slippage_tolerance is None:
-            super().__init__(budget, rng)
-        else:
-            super().__init__(budget, rng, slippage_tolerance)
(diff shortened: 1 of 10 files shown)


wakamex:
so the 3 spots are:
1. agent policy init parameter (allow None here)
2. HyperdriveMarketAction (allow None here)
	- having None here allows us to skip simulating the transaction 👍
3. AgentConfig
	- don't allow None here, this is the earliest instance, which flows everywhere else

I would love to merge trade (or is it trade_envelope?) and actiontype just to simplify things (less code is best code!)

examples/eth_bots/custom_policies/example_custom_policy.py
```diff
@@ -36,10 +36,7 @@ def __init__(
         else:
             self.trade_amount: FixedPoint = trade_amount
 
-        if slippage_tolerance is None:
-            super().__init__(budget, rng)
-        else:
-            super().__init__(budget, rng, slippage_tolerance)
+        super().__init__(budget, rng, slippage_tolerance)
```
wakamex, line 39: 4x fewer lines! 😎

wakamex (approved):

PR #780, opened 2023-08-02
refactor elfpy to look like a monorepo

dpaiton:
Basically moves files into a new `src/` directory.  Well-formed packages will be located in the `lib/` directory and cannot depend on `src/`.  They should be considered standalone and possible candidates for their own repo.


sentilesdal (commented):
This is a nice transition.

sentilesdal (approved):
This is a nice transition.

sentilesdal (approved):
This is a nice transition.

sentilesdal (approved):
This is a nice transition.

PR #785, opened 2023-08-03
acquire data on latest block

wakamex:


dpaiton (approved):
👍

dpaiton (changes requested):
Requesting a review from Sheng, since I am now realizing that the `-1` was probably on purpose.

slundqui:
Requesting a review from @sentilesdal, since it was a conversation with him that suggested we shouldn't assume the most current block is fixed. Although, we should look into if the block is only queryable after it's mined.

wakamex:
discussion continued in #801 
merged into my testing branch at https://github.com/wakamex/elf-simulations

PR #787, opened 2023-08-03
Clean up data/ folder

sentilesdal:
Commits listed below.  I'm trying to clean up the files in the data/ directory.  A lot of this is to collocated hyperdrive specific functionality in a sub-directory for now so that pure database things can live in data/.  We should also separate out things that aren't specifically for the database which I'll do next.  This PR corresponds with 
https://github.com/delvtech/infra/pull/65

- move address fetcher to hyperdrive/addresses.py
- move helper function
- move server out of data
- move hyperdrive specific things
- move hyperdrive specific things
- move more things to data/hyperdrive/ files


src/eth_bots/core/setup_experiment.py
```diff
@@ -1,5 +1,5 @@
 """Setup helper function for running eth bot experiments."""
-from __future__ import annotations
+fro...hyperdrive __futuraddressesions
```
dpaiton: umm.. what's up with this?

dpaiton (approved):
lgtm but you should probably wait for Sheng to check it.

src/data/db_schema.py
```diff
@@ -22,111 +21,7 @@
...
-    __tablename__ = "walletdelta"
-
-    # Default table primary key
-    # Note that we use postgres in production and sqlite in testing, but sqlite has issues with
-    # autoincrement with BigIntegers. Hence, we use the Integer variant when using sqlite in tests
-    id: Mapped[int] = mapped_column(
-        BigInteger().with_variant(Integer, "sqlite"), primary_key=True, init=False, autoincrement=True
-    )
-    transactionHash: Mapped[str] = mapped_column(String, ForeignKey("transactions.transactionHash"), index=True)
     blockNumber: Mapped[int] = mapped_column(BigInteger, ForeignKey("poolinfo.blockNumber"), index=True)
```
slundqui, line 24: I think this is now saying every db schema table has a field called `blockNumber` that's a foreignKey to poolinfo.blockNumber, which doesn't seem correct.

src/data/db_schema.py
```diff
@@ -146,6 +41,9 @@ class Transaction(Base):
     transactionHash: Mapped[str] = mapped_column(String, index=True, unique=True)
 
     #### Fields from base transactions ####
+    ##############
+    # NOOOOOOOOO, transaction shouldn't have a foreign key to poolInfo
+    ##############
```
slundqui, line 46: Why not?

src/data/hyperdrive/convert_data.py
```diff
@@ -280,36 +278,16 @@ def convert_checkpoint_info(checkpoint_info_dict: dict[str, Any]) -> db_schema.C
             if isinstance(value, FixedPoint):
                 value = Decimal(str(value))
         args_dict[key] = value
-    block_checkpoint_info = db_schema.CheckpointInfo(**args_dict)
+    block_checkpoint_info = src.data.hyperdrive.db_schema.CheckpointInfo(**args_dict)
     return block_checkpoint_info
 
 
-def _convert_scaled_value(input_val: int | None) -> Decimal | None:
```
slundqui, line 287: We should consider moving this back. This function was meant to be a helper function for converting chain query to postgres specifically, and not a general use function

src/data/postgres.py
```diff
@@ -174,107 +176,6 @@ def close_session(session: Session) -> None:
...
-    for checkpoint_info in checkpoint_infos:
-        session.add(checkpoint_info)
-    try:
-        session.commit()
-    except exc.DataError as err:
-        session.rollback()
-        raise err
-
-
 def add_transactions(transactions: list[Transaction], session: Session) -> None:
```
slundqui, line 179: I'd argue that these transactions isn't at all generic, but instead specific to hyperdrive. This is due to decoding of logs. Fields of `Transaction` include specific columns that match the input parameters to methods (e.g., openLong)

src/data/db_schema.py
slundqui: For example, there are tables that don't have a blockNumber at all, e.g., `usermap`, `poolconfig`, etc

src/data/db_schema.py
sentilesdal: So hopefully we can make a basic Transaction table that has no dependency on hyperdrive.  We could attach hyperdrive specific transaction info from another table and join when necessary.

src/data/db_schema.py
sentilesdal: ya its not right.  i was going for an abstract table here, fixed it.  good catch.

src/data/postgres.py
sentilesdal: yes very true, I didn't scroll down enough.

src/data/hyperdrive/convert_data.py
sentilesdal: per IRL convo leaving as is since its referrenced in two files

PR #788, opened 2023-08-04
calculate pnl for closeout

wakamex:
image

### DIFF
--- a/src/dashboard/calc_pnl.py
+++ b/src/dashboard/calc_pnl.py
@@ -2,13 +2,80 @@
 from __future__ import annotations
 
 import logging
+from decimal import Decimal
 
+import numpy as np
 import pandas as pd
+from eth_typing import ChecksumAddress, HexAddress, HexStr
+from fixedpointmath import FixedPoint
+from web3 import Web3
 
+from src import eth
 from src.dashboard.extract_data_logs import calculate_spot_price
+from src.eth.transactions import smart_contract_preview_transaction
+from src.eth_bots.core.environment_config import EnvironmentConfig
+from src.hyperdrive import contract_interface
 
 
-def calc_total_returns(pool_config: pd.Series, pool_info: pd.DataFrame, wallet_deltas: pd.DataFrame) -> pd.Series:
+def calc_closeout_pnl(current_wallet: pd.DataFrame, pool_info: pd.DataFrame, env_config: EnvironmentConfig):
+    """Calculate closeout value of agent positions."""
+    web3: Web3 = eth.initialize_web3_with_http_provider(env_config.rpc_url, request_kwargs={"timeout": 60})
+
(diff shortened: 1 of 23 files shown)


wakamex:
@sentilesdal I changed your `smart_contract_preview_transaction` to take an address as an input instead of an EthAgent, so I can call it when I don't have an EthAgent, but I do have an address. Changed where it's called to extract the address from EthAgent prior to calling it.

wakamex:
@slundqui I used `coerce_float=False` in the postgres queries. Since that test is a success, should we just maket that the default?

slundqui:
> @slundqui I used `coerce_float=False` in the postgres queries. Since that test is a success, should we just maket that the default?

No, I think in general, anything consuming from the database for the purposes of dashboarding should be able to handle the faster but less accurate floats instead of the more accurate but slower decimal format

src/dashboard/calc_pnl.py
```diff
@@ -134,12 +199,12 @@ def calculate_spot_price_for_position(
...
     maturity_timestamp : pd.Series
         The maturity timestamp
-    block_timestamp : pd.Series
+    block_timestamp : Decimal
         The block timestamp
     """
     # pylint: disable=too-many-arguments
     full_term_spot_price = calculate_spot_price(share_reserves, bond_reserves, initial_share_price, time_stretch)
-    time_left_seconds = maturity_timestamp - block_timestamp
+    time_left_seconds = maturity_timestamp.apply(lambda x: x - block_timestamp)
```
slundqui, line 207: I'm pretty sure this can just be `maturity_timestamp - block_timestamp`, as this is a vector - scalar, which should just broadcast

src/dashboard/calc_pnl.py
```diff
@@ -2,13 +2,80 @@
...
+    web3: Web3 = eth.initialize_web3_with_http_provider(env_config.rpc_url, request_kwargs={"timeout": 60})
+
+    # send a request to the local server to fetch the deployed contract addresses and
+    # all Hyperdrive contract addresses from the server response
+    addresses = contract_interface.fetch_hyperdrive_address_from_url(env_config.artifacts_url)
+    abis = eth.abi.load_all_abis(env_config.abi_folder)
+    contract = contract_interface.get_hyperdrive_contract(web3, abis, addresses)
+
+    # Define a function to handle the calculation for each group
+    def calc_single_closeout(position: pd.DataFrame, min_output: int, as_underlying: bool):
```
slundqui, line 31: Lets move this function definition outside of this function. It's complicated enough to be upgraded from a local function definition.

slundqui (approved):
This looks great, thanks!

src/dashboard/calc_pnl.py
wakamex: ya it was, I just had a pyright error... maybe `#type: ignore` is better here?

PR #789, opened 2023-08-04
fix imports

wakamex:


wakamex:
not needed if you `pip uninstall elfpy` then `pip install -r requirements.txt` which installs elfpy from `lib/elfpy`

PR #790, opened 2023-08-04
Use standard  python project folder structure

sentilesdal:
This gets us back to the standard project folder structure used in python projects.  code is moved from `src/` to `elf-simulations/`.  The pyproject.toml was grabbed from a previous commit and needs to be cleaned up.

- rename src -> elf_simulations
- add old pyproject, copy previous version
- update name in pyproject.toml
- change 'from src' to 'from elf_simulations'
- fix more imports
- fix more imports


dpaiton:
Not implementing in favor of some other structure

PR #795, opened 2023-08-07
Adds developer mode to easily fund bots

dpaiton:


lib/agent0/agent0/hyperdrive/exec/setup_experiment.py
```diff
@@ -47,31 +46,54 @@ def setup_experiment() -> tuple[Web3, Contract, Contract, EnvironmentConfig, lis
...
+    tuple[Web3, Contract, Contract]
+        A tuple containing:
+            - The web3 container
+            - The base token contract
+            - The hyperdrive contract
+    """
     # Check for default name and exit if is default
     if environment_config.username == DEFAULT_USERNAME:
-        raise ValueError("Default username detected, please update 'username' in eth_bots_config.py")
+        raise ValueError(
```
slundqui, line 79: we should not do this check if `--develop` is true

lib/agent0/agent0/hyperdrive/exec/setup_experiment.py
```diff
@@ -47,31 +46,54 @@ def setup_experiment() -> tuple[Web3, Contract, Contract, EnvironmentConfig, lis
...
     )
     setup_hyperdrive_crash_report_logging()
+    web3, base_token_contract, hyperdrive_contract = get_web3_and_contracts(environment_config)
+    # load agent policies
+    # rng is shared by the agents and can be accessed via `agent_accounts[idx].policy.rng`
+    agent_accounts = get_agent_accounts(agent_config, web3, base_token_contract, hyperdrive_contract.address, rng)
+    # Set up postgres to write username to agent wallet addr
+    # initialize the postgres session
+    wallet_addrs = [str(agent.checksum_address) for agent in agent_accounts]
+    register_username(environment_config.username_register_url, wallet_addrs, environment_config.username)
```
slundqui, line 56: we should not register username if --develop is true

lib/agent0/agent0/hyperdrive/generate_env.py
```diff
@@ -0,0 +1,62 @@
...
+import argparse
+import json
+import os
+
+import numpy as np
+from agent0.base.make_key import make_private_key
+from agent0.hyperdrive.config import get_eth_bots_config
+
+
+def generate_env(user_key: str) -> str:
```
slundqui, line 11: No need for changes here, but just so you're aware, running data stuff locally reads the .env file (e.g., getting sql credentials)

slundqui (approved):

lib/agent0/agent0/hyperdrive/generate_env.py
dpaiton: set_env will not overwrite the existing file, only add env variables.

PR #796, opened 2023-08-07
Updating installation to allow for installing individual subpackages

slundqui:
This solves the issue of specifying lateral dependencies for solo installation. The installation now allows for several optional tags for installation. For example, for agent0:

```
pip install agent0[with-dependencies] # Install individual subpackage
pip install agent0[base] # Install with base packages only (this is what's called in requirements.txt)
```

Note this doesn't change anything on default dev installation, i.e.,
```
pip install -r requirements.txt
```

### DIFF
--- a/lib/chainsync/pyproject.toml
+++ b/lib/chainsync/pyproject.toml
@@ -16,14 +16,14 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 
-dependencies = [
-    # Lateral dependencies across subpackages are pointing to github
-    # TODO test that outermost pip install overwrites this
-    # TODO move this to either this monorepo or seperate repo
-    #"fixedpointmath @ git+https://github.com/delvtech/agent_0.git/#subdirectory=lib/fixedpointmath",
-    # TODO this might not work since it's not on github yet, need to test
-    #"elfpy @ git+https://github.com/delvtech/elf-simulations.git/#subdirectory=lib/elfpy",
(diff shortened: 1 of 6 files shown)


lib/chainsync/pyproject.toml
```diff
@@ -16,14 +16,14 @@ classifiers = [
...
-    # TODO this might not work since it's not on github yet, need to test
-    #"elfpy @ git+https://github.com/delvtech/elf-simulations.git/#subdirectory=lib/elfpy",
-    #"ethpy @ git+https://github.com/delvtech/elf-simulations.git/#subdirectory=lib/ethpy",
+[project.optional-dependencies]
+# This flag installs all dependencies and should be ran when installing this subpackage in isolation
+with-dependencies = [
+    "chainsync[base, lateral]"
+]
+base = [
+    # TODO move fixedpointmath to lateral if this package comes back to this monorepo
```
dpaiton, line 25: I think we decided to move it to its own repo instead of bringing it back here. This monorepo is for packages that are still under development.

dpaiton (approved):
Can you add a note in the `INSTALL.md` file that informs what these [tags] mean? Something similar to the PR description will be fine.

lib/chainsync/pyproject.toml
slundqui: Sounds good. `fixedpointmath` should belong here then.

PR #800, opened 2023-08-08
improve wallet.py readability and debugging

wakamex:
- use the same logic to update long and short positions
- show pre-trade balance
- log same debug statement for both longs and shorts


wakamex:
in deprecated part of codebase

PR #807, opened 2023-08-09
Chainsync cleanup

slundqui:
General cleanup of chainsync (that touches on a couple of other packages). Here's a rundown:

Major changes:
- Moving crash reporting and test from chainsync to agent0
   - This requires some adjustments, this crash report is using postgres data structures, where it should be using a more generic data structure
   - After discussion, removed dead code and referenced in https://github.com/delvtech/elf-simulations/issues/479
- Changing chainsync imports from within a subpackage to be relative imports. This fixes circular import issues
   - Need to go through other packages to match this import structure
- Changing testing structure of chainsync to live close to the file it's testing, and renaming to be `<file_to_be_tested>_test.py`
   - elfpy still follows old testing structure, but the pytest config allows for both

Minor changes:
- Moving chainsync base/hyperdrive to an outer db directory
- Renaming `postgres` to `db_interface`
- Moving hyperdrive specific transactions to hyperdrive dir, and renaming to HyperdriveTransactions
- Moving `get_hyperdrive_contract` to ethpy
- Moving helper functions for dashboard from the bin script to analysis
- Adding checks to dashboard so it runs cleanly when there's no data
- Removing pytest.ini in favor of defining in pyproject.toml


lib/chainsync/chainsync/dashboard/usernames.py
```diff
@@ -0,0 +1,116 @@
+"""Helper functions for mapping addresses to usernames."""
+import pandas as pd
+
+
+def combine_usernames(username: pd.Series) -> pd.DataFrame:
+    """Map usernames to a single user (e.g., combine click with bots)."""
+    # TODO Hard coded mapping, should be a config file somewhere
+    user_mapping = {
```
dpaiton, line 8: now that we're done with the trading competition would it make sense to move this information to a config file that is not tracked by GH? So instead of hard-coding the mapping we can load a file and construct it programatically. Then in future iterations we can modify the files as needed w/o committing code.

lib/chainsync/chainsync/dashboard/usernames.py
```diff
@@ -0,0 +1,116 @@
...
+    username_column = username.name
+    user = username.to_frame().merge(user_mapping, how="left", left_on=username_column, right_index=True)
+    return user
+
+
+def get_click_addresses() -> pd.DataFrame:
+    """Return a dataframe of hard coded click addresses."""
+    # TODO Hard coded mapping, should be a config file somewhere
+    addresses = {
+        "0x004dfC2dBA6573fa4dFb1E86e3723e1070C0CfdE": "Charles St. Louis (click)",
```
dpaiton, line 42: Same comment as above: should we move this to a non-tracked config file and programmatically generate the addresses dict? You could save the config as a json and trivially load it in.

lib/chainsync/chainsync/dashboard/usernames.py
```diff
@@ -0,0 +1,116 @@
...
+    }
+    addresses = pd.DataFrame.from_dict(addresses, orient="index")
+    addresses = addresses.reset_index()
+    addresses.columns = ["address", "username"]
+
+    return addresses
+
+
+def get_user_lookup(agents, bots_map) -> pd.DataFrame:
+    """Generate username to agents mapping.
```
dpaiton, line 69: can you add type hints to these inputs? and add `Arguments` to the docstring.

lib/chainsync/chainsync/dashboard/usernames.py
dpaiton: I'm fine with this coming in a follow-up PR

lib/chainsync/chainsync/dashboard/usernames.py
dpaiton: I'm fine with this coming in a follow-up PR

lib/chainsync/chainsync/dashboard/usernames.py
dpaiton: I'm fine with this coming in a follow-up PR

lib/chainsync/chainsync/db/hyperdrive/db_interface.py
```diff
@@ -2,12 +2,32 @@
...
+    session : Session
+        The initialized session object
+    """
+    for transaction in transactions:
+        session.add(transaction)
+    try:
+        session.commit()
+    except exc.DataError as err:
+        session.rollback()
+        print(f"{transactions=}")
```
dpaiton, line 29: We should be logging, not printing, unless I've missed something?

dpaiton:
might as well remove the `lib/chainsync/tests` folder

dpaiton:
What is the intended difference between `scripts` and `bin`?

dpaiton:
I think it would be reasonable to just remove the `db_` prefix in front of the files in the `db` folders. e.g. `chainsync/db/base/db_schema.py` --> `chainsync/db/base/schema.py`. Same for the ones in the `db/hyperdrive` folder.

slundqui:
> What is the intended difference between `scripts` and `bin`?

`bin` are entrypoints to various services, e.g., from docker. `scripts` are one-offs, probably should be in jupytext format, and likely shouldn't really be tested or maintained, e.g., calculating previous leaderboards from trading competitions.

I don't know what the best way to keep something like scripts around, because for the time being, some of the functionality in those scripts can definitely be generalized once we get some time to look at it.

lib/chainsync/chainsync/db/hyperdrive/interface.py
```diff
@@ -423,8 +478,8 @@ def get_wallet_deltas(
     return pd.read_sql(query.statement, con=session.connection(), coerce_float=coerce_float)
 
 
-def get_agents(session: Session, start_block: int | None = None, end_block: int | None = None) -> list[str]:
-    """Get the list of all agents from the WalletInfo table.
+def get_all_traders(session: Session, start_block: int | None = None, end_block: int | None = None) -> list[str]:
```
dpaiton, line 481: the docstring still says "agent addresses". But tbh, I think a more appropriate name would be `get_all_addresses` since there seems to be no distinction in the code about whether the address belongs to a bot ("agent") or human (what I assume you mean by "trader")

lib/chainsync/chainsync/db/base/__init__.py
```diff
@@ -1,19 +1,17 @@
...
     drop_table,
     get_latest_block_number_from_table,
-    get_transactions,
     get_user_map,
     initialize_engine,
     initialize_session,
     query_tables,
 )
+from .schema import Base, UserMap
+from .utils import convert_scaled_value_to_decimal
```
dpaiton, line 17: @sentilesdal has beef with "utils" because it's too generic. I don't have strong opinions here, but I think he's probably more right than not. (true in general lol) Since this is the only function in the file can we just name it as such? `from .convert_scaled_value_to_decimal import convert_scaled_value_to_decimal` is annoying, but would only need to be done within the `db/base` folder. Elsewhere it is covered by this`__init__` file.

lib/chainsync/chainsync/db/hyperdrive/interface.py
slundqui: This function is supposed to be getting all addresses that have made a trade, regardless of bot or human. This information is gathered from transactions themselves

slundqui: Hence, "trader"

lib/chainsync/chainsync/db/base/__init__.py
slundqui: We've gone back and forth on this so many times, this function was intended to be a private method within convert_data.py, not to be used by anyone else. Perhaps when we find that this function is more useful generically, we can revisit this. For now, moving this function back as a private function in `convert_data.py`

lib/chainsync/chainsync/db/hyperdrive/convert_data.py
```diff
@@ -86,6 +87,28 @@ def _convert_object_hexbytes_to_strings(obj: Any) -> Any:
     return obj
 
 
+def _convert_scaled_value_to_decimal(input_val: int | None) -> Decimal | None:
```
dpaiton, line 90: 👍 I like this solution best

dpaiton (approved):
lgtm thanks for doing this & dealing with my many nits!

PR #808, opened 2023-08-09
Removing rate limit check and going back to using default github token

slundqui:


.github/workflows/static.yml
```diff
@@ -11,30 +11,20 @@ jobs:
...
           python -m pip install -r requirements-dev.txt
 
       - name: Analysing code with pyright
-        run: python -m pyright $(git ls-files '*.py' '*.pyi')
-
-      - name: Check GitHub rate limit at end
-        uses: wakamex/rate-limit-action@master
-        with:
-          token: ${{ secrets.GH_GITHUB_COM_TOKEN || github.token}}
+        run: python -m pyright $(git ls-files '*.py' '*.pyi')
```
dpaiton, line 30: newline went away -- do we need it in the yamls?

dpaiton (approved):

.github/workflows/static.yml
slundqui: CI ran fine, so I guess not

PR #815, opened 2023-08-11
Initial system test

slundqui:
This PR does the following:

- Abstracts various test fixtures across the repo
- Implements various ethpy helper functions
- Adds test fixture for running local chain
- Adds test fixtures for initializing hyperdrive

Currently, the hyperdrive initialization is defined in a test fixture (as it hard codes initial parameters). Moving this to be a general ethpy function is future work.

Actual system tests coming in a future PR.


lib/ethpy/ethpy/base/abi/load_abis.py
```diff
@@ -4,9 +4,20 @@
 import json
 import logging
 import os
+from typing import Literal, overload
 
 
-def load_all_abis(abi_folder: str) -> dict:
+@overload
```
dpaiton, line 10: overloading this is very cool!

lib/ethpy/ethpy/base/abi/load_abis.py
```diff
@@ -20,21 +31,27 @@ def load_all_abis(abi_folder: str) -> dict:
...
         except AssertionError as err:
             logging.debug("JSON file %s did not contain an ABI.\nError: %s", abi_file, err)
     logging.debug("Loaded ABI files %s", str(loaded))
+    if return_bytecode:
+        return abis, bytecodes
     return abis
 
 
-def load_abi_from_file(file_name: str) -> dict:
+def load_abi_from_file(file_name: str, return_bytecode: bool = False) -> dict | tuple[dict, dict]:
```
dpaiton, line 54: shouldn't you have the `@overload` decorator for this func too?

lib/ethpy/ethpy/base/contract/deploy_contract.py
```diff
@@ -0,0 +1,77 @@
...
+        The contract bytecode
+    deploy_addr:
+        The address of the account that's deploying the contract
+    args: list[Any] | None:
+        List of arguments to pass to the contract constructor
+        No arguments if None
+
+    Returns
+    -------
+    str
```
dpaiton, line 35: wrong return type & desc

lib/ethpy/ethpy/base/contract/deploy_contract.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Arguments
+    ---------
+    web3: Web3
+        web3 provider object
+    abi: list[Any]
+        The contract abi
+    bytecode: str
+        The contract bytecode
+    deploy_addr:
```
dpaiton, line 27: no type listed

lib/ethpy/ethpy/base/contract/deploy_contract.py
```diff
@@ -0,0 +1,77 @@
...
+
+from typing import Any
+
+from eth_typing import Address
+from web3 import Web3
+from web3.contract.contract import Contract
+
+
+def deploy_contract_and_return(
+    web3: Web3, abi: list[Any], bytecode: str, deploy_addr: str, args: list | None = None
```
dpaiton, line 13: args should be `list[Any]` I guess?

lib/ethpy/ethpy/base/contract/deploy_contract.py
```diff
@@ -0,0 +1,77 @@
...
+        List of arguments to pass to the contract constructor
+        No arguments if None
+
+    Returns
+    -------
+    str
+        The deployed contract address
+
+    """
+    contract_addr: Address = deploy_contract(web3, abi, bytecode, deploy_addr, args)  # type: ignore
```
dpaiton, line 39: Can you add a comment explaining why you had to add the `# type: ignore` pls?

lib/ethpy/ethpy/base/contract/deploy_contract.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Arguments
+    ---------
+    web3: Web3
+        web3 provider object
+    abi: list[Any]
+        The contract abi
+    bytecode: str
+        The contract bytecode
+    deploy_addr:
```
dpaiton, line 57: no type

lib/ethpy/ethpy/base/abi/load_abis.py
slundqui: Doesn't look like it: https://adamj.eu/tech/2021/05/29/python-type-hints-how-to-use-overload/

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+    --------
+    web3 : Web3
+        web3 provider object
+
+    Returns
+    -------
+    LocalAccount
+        The LocalAccount object
+    """
+    # TODO get private key of this account programatically
```
dpaiton, line 56: I don't think this is possible :( I was trying to do it a while back. 

However, I think one solution is to force the [mnemonic](https://book.getfoundry.sh/reference/anvil/#:~:text=%2Dm%2C%20%2D%2Dmnemonic%20%3CMNEMONIC%3E) when you spin up anvil, and then you can use web3 to generate a private key from that mnemonic (`account = w3.eth.account.from_mnemonic("some dumb list of words"`). This should be the same private key as the one anvil is using. The mnemonic can be set in a `.env` file.

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+    # Ensure this private key is actually matched to the first address of anvil
+    assert web3.eth.accounts[0] == account.address
+    return account
+
+
+def deploy_hyperdrive_factory(rpc_url: str, deploy_account: LocalAccount) -> tuple[Contract, Contract]:
+    """Deploys the hyperdrive factory contract on the rpc_url chain
+
+    Arguments
+    --------
```
dpaiton, line 69: missing a `-`

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+    )
+
+    forwarder_factory_addr, forwarder_factory_contract = deploy_contract_and_return(
+        web3,
+        abi=abis["ForwarderFactory"],
+        bytecode=bytecodes["ForwarderFactory"],
+        deploy_addr=deploy_addr,
+    )
+
+    deployer_addr = deploy_contract(
```
dpaiton, line 122: I went on a goose chase from not noticing the difference between `deployer_addr` and `deploy_addr`. Can you use another name? I _think_ the `deploy_addr` is the address for the account that is funding the gas for the deployment? Are they also paying the initial balance? And the `deployer_addr` is the address of the deployed contract? So would `contract_addr` be more appropriate?

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+        web3,
+        abi=abis["ERC4626HyperdriveDeployer"],
+        bytecode=bytecodes["ERC4626HyperdriveDeployer"],
+        deploy_addr=deploy_addr,
+        args=[pool_addr],
+    )
+
+    factory_config = (
```
dpaiton, line 130: Can you add a comment here explaining the same arg being used? I assume the `deploy` account is funding all 3 contracts?

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+    max_flat_fee = int(0.0015e18)  # 0.15%
+    max_governance_fee = int(0.30e18)  # 0.30%
+    # Configuration settings
+    abi_folder = "packages/hyperdrive/src/abis/"
+
+    # Load compiled objects
+    abis, bytecodes = load_all_abis(abi_folder, return_bytecode=True)
+    web3 = initialize_web3_with_http_provider(rpc_url, reset_provider=False)
+    # Convert deploy address to checksum address
+    deploy_addr = Web3.to_checksum_address(deploy_account.address)
```
dpaiton, line 97: these newlines are definitely not separating logical blocks in the code :-P

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+def deploy_and_initialize_hyperdrive(
+    web3: Web3,
+    base_token_contract: Contract,
+    factory_contract: Contract,
+    deploy_account: LocalAccount,
+) -> str:
+    """Calls the hyperdrive factory to deploy and initialize new hyperdrive contract
+
+    Arguments
+    --------
```
dpaiton, line 167: missing a '-'

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -0,0 +1,242 @@
...
+
+    logs = get_transaction_logs(factory_contract, tx_receipt)
+    hyperdrive_address = None
+    for log in logs:
+        if log["event"] == "GovernanceUpdated":
+            hyperdrive_address = log["address"]
+    if hyperdrive_address is None:
+        raise AssertionError("Generating hyperdrive contract didn't return address")
+
+    # TODO do I return the contract or address here?
```
dpaiton, line 241: Do you mean the hyperdrive contract? I don't see that in here. The address is probably fine.

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
dpaiton: You can leave this as a TODO. I made an [issue for it](https://github.com/delvtech/elf-simulations/issues/816)

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
dpaiton: I think this actually is important for sphinx, fwiw

dpaiton (approved):
I have no major concerns, you did a great job! I did have a large number of nits, sorry about that. hmu if you need more context on any of them.

lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
slundqui: Honestly, I don't know what each of these accounts are doing, I'm mirroring what's being done in the solidity script

PR #825, opened 2023-08-15
Agent0 and config refactor

slundqui:
Major changes:
- Consolidates rpc url env configurations across all subpackages
- Splits out postgres and account env configurations
- Adding dataclass and helper functions for all configurations
- Agent0 bots entry points now defines all configurations and executes bots via 2 lines (defined in examples)
- Passing in all individual configuration files to where it's needed

Minor changes:
- Allowing specification of exact agent budget
- Adding in base abstract class for HyperdrivePolicy
- Adding example custom bot

TODO:
- Consolidating all configurations into one variable for passing into various functions
- Wrapper stateful object to ease syntax, reduce imports, and set defaults for running custom bots


.gitignore
```diff
@@ -50,6 +50,9 @@ __pypackages__/
 
 # Environments
 .env
```
dpaiton, line 52: what about `*.env`?

.gitignore
```diff
@@ -50,6 +50,9 @@ __pypackages__/
...
+*account.env
```
dpaiton, line 53: What about just `*.env`?

lib/agent0/agent0/base/policies/base.py
```diff
@@ -25,6 +25,7 @@ class BasePolicy(Generic[MarketState, Wallet]):
     def __init__(
         self, budget: FixedPoint, rng: NumpyGenerator | None = None, slippage_tolerance: FixedPoint | None = None
     ):
+        # TODO budget should have a flag to allow for "the budget is however much this wallet has"
```
dpaiton, line 28: Made an issue for this: https://github.com/delvtech/elf-simulations/issues/827

lib/agent0/agent0/hyperdrive/exec/get_agent_accounts.py
```diff
@@ -15,12 +14,13 @@
 from web3 import Web3
 from web3.contract.contract import Contract
 
-# pylint: disable=too-many-locals
-
 
+# TODO consolidate config
```
dpaiton, line 18: What is this TODO for? I was going to make an issue but I need some more details.

lib/agent0/agent0/hyperdrive/exec/setup_experiment.py
```diff
@@ -1,41 +1,46 @@
...
+    contract_addresses: HyperdriveAddresses,
+) -> tuple[Web3, Contract, list[HyperdriveAgent]]:
     """Get agents according to provided config, provide eth, base token and approve hyperdrive.
 
     Returns
     -------
     tuple[Web3, Contract, Contract, EnvironmentConfig, list[HyperdriveAgent]]
         A tuple containing:
             - The web3 container
-            - The base token contract
```
dpaiton, line 29: We were returning both contracts because we kept needing them for debugging. Maybe the solution here (instead of returning an unused variable) is to return a dict with the key corresponding to the contract name and the value the contract itself. You can use the same key as was used to access the ABIs:`abis[environment_config.base_abi]`

lib/agent0/agent0/accounts_config.py
```diff
@@ -0,0 +1,190 @@
...
+
+@dataclass
+class AccountKeyConfig:
+    """The account config dataclass
+
+    Attributes
+    ----------
+    ARTIFACTS_URL: str
+        The url of the artifacts server from which we get addresses.
+    RPC_URL: URI | str
```
dpaiton, line 26: I guess all of these should be URI and not URL right?

lib/agent0/agent0/accounts_config.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    rng = np.random.default_rng(random_seed)
+    agent_private_keys = []
+    agent_base_budgets = []
+    agent_eth_budgets = []
+    for agent_info in agent_configs:
+        for _ in range(agent_info.number_of_agents):
+            agent_private_keys.append(make_private_key())
+
+            if isinstance(agent_info.eth_budget, Budget):
```
dpaiton, line 122: This cleans up nicely with [structured pattern matching](https://peps.python.org/pep-0622/#rationale-and-goals).
```
match agent_info.base_budget:
    case Budget as sampler_budget: # sample if a Budget object
        kwargs["budget"] = sampler_budget.sample_budget(rng) # TODO: rename this function to `sample()`
    case FixedPoint as scalar_budget: # sample is a fixed-point
        kwargs["budget"] = scalar_budget.scaled_value # TODO: rename this member variable to `fixed_value`
    case _ as budget_type:
        raise ValueError(f"Invalid base_budget type: {budget_type)}")
```

lib/agent0/bin/fund_bots_from_user_key.py
```diff
@@ -1,23 +1,57 @@
...
+    # Load config from env
+    account_key_config = build_account_config_from_env(env_file, user_key)
+    if account_key_config.USER_KEY == "":
+        raise ValueError("User key not provided as argument, and was not found in env file")
+
+    eth_config = build_eth_config()
+    contract_addresses = fetch_hyperdrive_address_from_url(os.path.join(eth_config.ARTIFACTS_URL, "addresses.json"))
+    user_account = HyperdriveAgent(Account().from_key(account_key_config.USER_KEY))
+
+    # TODO We hardcode ERC20Mintable here as the base contract name, since we don't have access to env_config here
```
dpaiton, line 51: See other comment about just hardcoding this everywhere.

dpaiton: There _are_ other contract options besides ERC20Mintable, but we can generalize the variable when we decide to add those (could be a while)

lib/agent0/examples/hyperdrive_bots.py
```diff
@@ -0,0 +1,82 @@
...
+            min_wei=1,  # 1 WEI base
+            max_wei=int(100_000e18),  # 100k base
+        ),
+        eth_budget=Budget(min_wei=int(1e18), max_wei=int(1e18)),
+        init_kwargs={"trade_chance": FixedPoint(0.8), "risk_threshold": FixedPoint(0.9)},
+    ),
+    AgentConfig(
+        policy=Policies.short_sally,
+        number_of_agents=0,
+        base_budget=Budget(
```
dpaiton, line 60: Can you set one of these to a FixedPoint to illustrate how that works?

lib/chainsync/chainsync/postgres_config.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    # default values for local postgres
+    # Matching environment variables to search for
+    # pylint: disable=invalid-name
+    POSTGRES_USER: str = "admin"
+    POSTGRES_PASSWORD: str = "password"
+    POSTGRES_DB: str = "postgres_db"
+    POSTGRES_HOST: str = "localhost"
+    POSTGRES_PORT: int = 5432
```
dpaiton, line 36: do you think we should add the function to write the class as JSON for all of these config dataclasses?

lib/agent0/agent0/accounts_config.py
```diff
@@ -0,0 +1,190 @@
...
+
+    # default values for local contracts
+    # Matching environment variables to search for
+    # pylint: disable=invalid-name
+    USER_KEY: str | None
+    AGENT_KEYS: list[str]
+    AGENT_ETH_BUDGETS: list[int]
+    AGENT_BASE_BUDGETS: list[int]
+
+    def to_env_str(self):
```
dpaiton, line 40: I think you can `return json.dumps(self.__dict__)` since it's a `dataclass`. At that point, maybe it makes sense to just make it a property?
```
@property
def env_str(self) -> str:
    return json.dumps(self.__dict__)
```

dpaiton: I guess you need to handle None. You could add `None` as a case in the [elfpy ExtendedJSONEncoder](https://github.com/delvtech/elf-simulations/blob/main/lib/elfpy/elfpy/utils/json.py#L11). We can package up utility stuff like this together out of elfpy later.

dpaiton (approved):
This is awesome! Really solid work. Next time can you break up your work into smaller PRs? That makes them easier to review & follow the new workflow.

lib/agent0/agent0/accounts_config.py
```diff
@@ -0,0 +1,190 @@
...
+        env_str += "AGENT_BASE_BUDGETS='" + json.dumps(self.AGENT_BASE_BUDGETS) + "'" + "\n"
+        env_str += "AGENT_ETH_BUDGETS='" + json.dumps(self.AGENT_ETH_BUDGETS) + "'" + "\n"
+        return env_str
+
+
+def initialize_accounts(
+    agent_config: list[AgentConfig], env_file: str | None = None, random_seed: int = 1, develop: bool = False
+) -> AccountKeyConfig:
+    """
+    Build or loads an accounts environment file
```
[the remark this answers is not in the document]
wakamex: "Build or load an account environment file."

```
D401 First line of docstring should be in imperative mood: "Builds or loads an accounts environment file."
```

wakamex:
do you think moving the ABIs into the agent0 package would make sense in this PR?

dpaiton:
can you also update the PR title so that it is more descriptive to the few things that are done here? (at least add that it refactors configs)

lib/agent0/agent0/hyperdrive/exec/get_agent_accounts.py
slundqui: I wasn't a fan of passing around three different types of configs everywhere, but maybe it's okay, as long as the agent0 interface is concise (which is part of the stateful wrapper for the interface todo). I've updated the comment.

lib/agent0/agent0/accounts_config.py
slundqui: This function is being used to write an .env file, which I don't think json is compatible with

lib/agent0/agent0/accounts_config.py
slundqui: Won't fix, these env variables match what's in infra: https://github.com/delvtech/infra/blob/main/env/env.common

lib/chainsync/chainsync/postgres_config.py
slundqui: Unsure what this function would be used for, only thing I see it being useful for is to print out a default env file based on this file. Currently, there's no script that will do this for you, and templates are being handled via env.sample files. Let's push this down the road until we see a use case.

lib/agent0/examples/hyperdrive_bots.py
slundqui: Can you take a look about the TODO above?

dpaiton: this?
```
# TODO should these be in units of wei to match units of Budget?
# Or should budget accept FixedPoints in units of base + eth?
```

dpaiton: I don't understand the question. Everything is in wei, I don't know what units of Budget is. Budget generates ints in units of wei

slundqui: that's the one

dpaiton: I think everything should be in units of wei

PR #835, opened 2023-08-16
adds the pyperdrive wheel

dpaiton:
This PR adds the pyperdrive wheel files & includes the package in `requirements.txt`. It is not presently used anywhere.


sentilesdal (approved):
lgtm

PR #837, opened 2023-08-17
System tests

slundqui:
This PR adds the first system test for elf-simulations.

Major changes:
- It's here! Adds a system test that runs from bots to postgres db on a local test chain.
- Adding custom bot for test fixtures that cycles through all trades with known amounts
- Moving `acquire_data.py` to be within chainsync package, and writing minimal entry point to call
   - Breaking change: data capture script renamed from `acquire_data.py` to `run_chainsync.py`
- Parameterizing `acquire_data.py` to allow for system tests
- Explicitly setting a precision and scale for the Numeric postgres type

Minor changes:
- Adding script for making vscode debugging pytest easier
- Moving helper function `get_web3_and_contracts` from agent0 to ethpy
- Adding `coerce_float` to all postgres getter functions
- Docstrings
- Formatting

Bug fixes:
- Fixed a bug that stores `invTimeStretch` as unscaled values
- Fixed a bug with postgres/python schema typing for `maturityTime`

TODO:
- Do more comparisons in system tests, e.g., how much base for closing longs, pool info, etc
- Break out system_test
- There are rounding issues with using in-memory sqlite, should use containerized postgres
   - https://github.com/delvtech/elf-simulations/issues/836

### DIFF
--- a/lib/agent0/examples/example_agent.py
+++ b/lib/agent0/examples/example_agent.py
@@ -24,6 +24,9 @@ ENV_FILE = "example_agent.account.env"
 
 # Build custom policy
 # Simple agent, opens a set of all trades for a fixed amount and closes them after
+# TODO this bot is almost identical to the one defined in test_fixtures for system tests
+# On one hand, this bot is nice for an example since it shows all trades
+# On the other, duplicated code between the two bots
 class CycleTradesPolicy(HyperdrivePolicy):
     """An agent that simply cycles through all trades"""
 
@@ -109,7 +112,7 @@ class CycleTradesPolicy(HyperdrivePolicy):
                         ),
                     )
                 )
-        elif self.counter == 4:
+        elif self.counter == 5:
             # Close All Shorts
             assert len(wallet.shorts) == 1
             for short_time, short in wallet.shorts.items():
@@ -125,7 +128,7 @@ class CycleTradesPolicy(HyperdrivePolicy):
                         ),
                     )
                 )
-        elif self.counter == 5:
+        elif self.counter == 6:
             # Redeem all withdrawal shares
             action_list.append(
                 Trade(
@@ -137,7 +140,7 @@ class CycleTradesPolicy(HyperdrivePolicy):
                     ),
                 )
             )
-        elif self.counter == 6:
+        elif self.counter == 7:
             # One more dummy trade to ensure the previous trades get into the db
             # TODO test if we can remove this eventually
             action_list.append(
--- a/lib/chainsync/chainsync/db/hyperdrive/schema.py
+++ b/lib/chainsync/chainsync/db/hyperdrive/schema.py
@@ -36,7 +36,7 @@ class PoolConfig(Base):
     oracleSize: Mapped[Union[Decimal, None]] = mapped_column(FIXED_NUMERIC, default=None)
     updateGap: Mapped[Union[int, None]] = mapped_column(Integer, default=None)
     invTimeStretch: Mapped[Union[Decimal, None]] = mapped_column(FIXED_NUMERIC, default=None)
-    termLength: Mapped[Union[Decimal, None]] = mapped_column(FIXED_NUMERIC, default=None)
+    updateGap: Mapped[Union[int, None]] = mapped_column(Integer, default=None)
 
 
 class CheckpointInfo(Base):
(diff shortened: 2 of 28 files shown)


lib/agent0/examples/example_agent.py
```diff
@@ -24,6 +24,9 @@
 
 # Build custom policy
 # Simple agent, opens a set of all trades for a fixed amount and closes them after
+# TODO this bot is almost identical to the one defined in test_fixtures for system tests
```
dpaiton, line 27: Where is this bot being used? My vote is to remove it. In the docs we can point to the cycle_trade_policy for people to see the trades.

slundqui: This bot is being used as an example (i.e., this script calls the bot). It's meant as a template for writing custom bots.

lib/chainsync/chainsync/db/hyperdrive/schema.py
```diff
@@ -10,6 +10,11 @@
 
 # pylint: disable=invalid-name
 
+# Postgres numeric type that matches fixedpoint
+# The high precision doesn't actually allocate memory in postgres, as numeric is variable size
+# https://stackoverflow.com/questions/40686571/performance-of-numeric-type-with-high-precisions-and-scales-in-postgresql
+FIXED_NUMERIC = Numeric(precision=1000, scale=18)
```
dpaiton, line 16: Can you add a note to this comment saying "Precision indicates the total number of digits, while scale indicates the the number of digits to the right of the decimal."

I ask because according to the [Python Decimal type](https://docs.python.org/3/library/decimal.html#:~:text=Unlike%20hardware%20based%20binary%20floating%20point%2C%20the%20decimal%20module%20has%20a%20user%20alterable%20precision%20(defaulting%20to%2028%20places)%20which%20can%20be%20as%20large%20as%20needed%20for%20a%20given%20problem%3A), the "precision" specifies how many decimal places we support (i.e. to the right of the decimal point). While sqlalchemy uses the opposite (not well defined in their docs, but I think it is outlined [here](https://www.w3schools.com/sql/sql_datatypes.asp#:~:text=decimal(p%2Cs,and%20scale%20numbers)).

lib/ethpy/ethpy/hyperdrive/interface.py
```diff
@@ -116,9 +116,9 @@ def get_hyperdrive_config(hyperdrive_contract: Contract) -> dict[str, Any]:
     # Ok so, the contracts store the time stretch constant in an inverted manner from the python.
     # In order to not break the world, we save the contract version as 'invTimeStretch' and invert
     # that to get the python version 'timeStretch'
-    # TODO invTimeStretch should be in fixed point notation
-    pool_config["invTimeStretch"] = hyperdrive_config["timeStretch"]
-    pool_config["timeStretch"] = FixedPoint(1) / FixedPoint(scaled_value=hyperdrive_config["timeStretch"])
+    # TODO fix this
```
dpaiton, line 119: fix what? Can you clarify so that this TODO is helpful to a reader at a later time?

slundqui: The comment about this is right above this line, but I'll change it

tests/system_test.py
```diff
@@ -1,17 +1,377 @@
...
+from chainsync.exec import acquire_data
+from eth_account.signers.local import LocalAccount
+from ethpy import EthConfig
+from ethpy.hyperdrive import HyperdriveAddresses
+from ethpy.test_fixtures.deploy_hyperdrive import _calculateTimeStretch
+from fixedpointmath import FixedPoint
+from sqlalchemy.orm import Session
+
+# This pass is to prevent auto reordering imports from reordering the imports below
+pass  # pylint: disable=unnecessary-pass
```
dpaiton, line 22: 🙃

dpaiton (approved):
Awesome job! Everything is really coming together :)

lib/agent0/examples/example_agent.py
dpaiton: I see, well what about simplifying it, then? Like have it return a no-op (empty list) or have it open a long and then immediately close it. We don't need it to complete every possible trade to demonstrate how to write a custom agent. And if it's super long then people might be weary of making it work. The random agent policy shows all trades, and the new test policy shows all trades. So people have examples.

PR #840, opened 2023-08-18
Create MVP Pypechain lib

sentilesdal:
Creates a simple Pypechain library that generates custom web3.py `Contract` classes for a given abi.  For example, ERC20.json will create an ERC20Contract.py file that contains a custom ERC20Contract class along with classes for every public function the contract has.  Each function has the signature typed which will allow tools like VSCode to enable features like autocomplete when using the functions.  Next steps:

Testing
  - add unit tests for utilites
  - add basic test for template outputs from run_pypechain.py
 
Add Events
  - add ContractEvent classes for all events in an ABI.
 
Add Structs
  - add NamedTuple types for all internalTypes that can be found in an ABI.  This will allow us to have types for things like PoolConfig and PoolInfo based off the actual source of the struct.

Nice to haves
  - Add file splitting for template output, group functions, events, structs into their own files.
  - Add bash script or some installable package so that pypechain can be run from the console for any package.


lib/pypechain/pypechain/sample/sample.py
```diff
@@ -0,0 +1,8 @@
+"""Sample usage of erc20Contract."""
+
+from build.ERC20Contract import ERC20Contract
+
+erc20Contract = ERC20Contract(address=None)
+
+fn = erc20Contract.functions.allowance("0xOWNER", "0xSPENDER").call()
```
dpaiton, line 7: typo?

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -0,0 +1,116 @@
...
+    """Generates class files for a given abi."""
+    ### Set up the Jinja2 environment
+    # Determine the absolute path to the directory containing your script.
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+    # Construct the path to your templates directory.
+    templates_dir = os.path.join(script_dir, "templates")
+    env = Environment(loader=FileSystemLoader(templates_dir))
+    template = env.get_template("contract.jinja2")
+
+    _web3 = Web3()
```
dpaiton, line 31: we typically follow Web3's convention to call this var `web3`. I assume you added the underscore bc of an earlier lint namespace error (this has come up a lot in the past), but it looks like you can remove it now.

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -0,0 +1,116 @@
...
+import sys
+from pathlib import Path
+
+from jinja2 import Environment, FileSystemLoader
+from pypechain.utilities import avoid_python_keywords, is_abi_function, solidity_to_python_type
+from web3 import Web3
+from web3.types import ABIFunction
+
+
+def load_abi_from_file(file_path: Path):
```
dpaiton, line 15: might as well add proper docstrings on all of these funcs. I think we're going to turn on that linter soon.

lib/pypechain/pypechain/utilities/abi.py
```diff
@@ -0,0 +1,86 @@
...
+from web3.types import ABIEvent, ABIFunction
+
+
+class Input(NamedTuple):
+    """An input of a function or event."""
+
+    internalType: str
+    name: str
+    type: str
+    indexed: Optional[bool] = None
```
dpaiton, line 16: I believe the `Optional[]` syntax is being deprecated in favor of `indexed: bool | None = None`

lib/pypechain/pypechain/utilities/abi.py
```diff
@@ -0,0 +1,86 @@
...
+    name: str
+    type: str
+
+
+class AbiItem(NamedTuple):
+    """An item of an ABI, can be an event, function or struct."""
+
+    type: str
+    inputs: List[Input]
+    stateMutability: Optional[str] = None
```
dpaiton, line 33: same as above, should be `str | None = None`

lib/pypechain/pypechain/utilities/abi.py
```diff
@@ -0,0 +1,86 @@
...
+    outputs: Optional[List[Output]] = None
+
+
+class AbiJson(NamedTuple):
+    """A JSON representation of a solidity contract's Application Boundary Interface."""
+
+    abi: List[AbiItem]
+
+
+def load_abi(abi_path: str) -> AbiJson:
```
dpaiton, line 45: we have a version of this in ethpy. Not sure if we want pypechain to depend on ethpy, though? I think eventually we'd expect it to be the other way around, ethpy depends on pypechain. So maybe we can just bake that in now (or in a follow-up PR)? Where ethpy uses this func.

dpaiton: no need to act on this now, just musing

dpaiton (approved):
left some minor comments but overall this is awesome! Can't wait to start using it :)

lib/pypechain/pypechain/utilities/abi.py
sentilesdal: oh yeah thanks.

lib/pypechain/pypechain/run_pypechain.py
sentilesdal: 100%!

lib/pypechain/pypechain/utilities/abi.py
sentilesdal: yeah i'll duplicate for now, if there's more than 3 overlapping i'll make the change.

PR #847, opened 2023-08-22
Generate Python types for all internal Solidity contract types.

sentilesdal:
These are not the best commits, but not the worst.  Probably best to look at the total files. 

This PR creates a new file like IHyperdriveTypes.py that has all the internal types of a contract.  There are also improvements to the formatting of the function parameters.  I've also added a very basic test for the contract jinja template.  This will be mostly replaced with better testing that uses a custom abi that i'll create to make sure we are testing all the crazy structs / nested tuples etc that we'll see returned from a contract.  Next steps will be to add output types to all functions and then events!

Example output:

IHyperdriveTypes.py
```python
"""Dataclasses for all structs in the IHyperdrive contract."""

@dataclass
class Checkpoint:
    """Checkpoint struct."""

    sharePrice: int
    longSharePrice: int
    shortBaseVolume: int

@dataclass
class Fees:
    """Fees struct."""

    curve: int
    flat: int
    governance: int


@dataclass
class PoolConfig:
    """PoolConfig struct."""

    baseToken: str
    initialSharePrice: int
    minimumShareReserves: int
    positionDuration: int
    checkpointDuration: int
    timeStretch: int
    governance: str
    feeCollector: str
    Fees: Fees
    oracleSize: int
    updateGap: int

# more types below
```

Example contracts output now:

IHyperdriveContract.py
```python
class IHyperdriveAddLiquidityContractFunction(ContractFunction):
    """ContractFunction for the addLiquidity method."""

    # pylint: disable=arguments-differ
    def __call__(
        self, _contribution: int, _minApr: int, _maxApr: int, _destination: str, _asUnderlying: bool
    ) -> "IHyperdriveAddLiquidityContractFunction":
        super().__call__(_contribution, _minApr, _maxApr, _destination, _asUnderlying)
        return self

    # TODO: add call def so we can get return types for the calls
    # def call()
    
    class IHyperdriveContractFunctions(ContractFunctions):
    """ContractFunctions for the IHyperdrive contract."""

    addLiquidity: IHyperdriveAddLiquidityContractFunction
    # more below
    
    class IHyperdriveContract(Contract):
    """A web3.py Contract class for the IHyperdrive contract."""

    def __init__(self, address: ChecksumAddress | None = None, abi=Any) -> None:
        self.abi = abi
        # TODO: make this better, shouldn't initialize to the zero address, but the Contract's init
        # function requires an address.
        self.address = address if address else cast(ChecksumAddress, "0x0000000000000000000000000000000000000000")

        try:
            # Initialize parent Contract class
            super().__init__(address=address)

        except FallbackNotFound:
            print("Fallback function not found. Continuing...")

    functions: IHyperdriveContractFunctions
```


lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -58,28 +72,41 @@ def local_hyperdrive_chain(local_chain: str) -> dict:
 
     Returns
     -------
-    dict
-        A dictionary with the following key - value fields:
+    LocalHyperdriveChain
+        A tuple with the following key - value fields:
 
-        "web3": Web3
+        web": Web3
```
dpaiton, line 78: this should be `web3` right? There's still a quote

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -179,12 +179,44 @@ def get_input_names(function: ABIFunction) -> list[str]:
...
+def get_outputs(function: ABIFunction) -> list[str]:
+    """Returns function input name/type strings for jinja templating.
+
+    i.e. for the solidity function signature:
+    function doThing(address who, uint256 amount, bool flag, bytes extraData)
+
+    the following list would be returned:
+    ['who', 'amount', 'flag', 'extraData']
+
+    ---------
```
dpaiton, line 191: typo? line ordering

lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -46,8 +49,19 @@ def local_chain() -> Generator[str, Any, Any]:
     anvil_process.kill()
 
 
+class LocalHyperdriveChain(NamedTuple):
```
dpaiton, line 52: nice I like this better

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -1,120 +1,120 @@
...
-    file_path : Path
-        The path to the ABI file.
+    abi_file_path : str
+        Path to the abi json file.
 
-    Returns
-    -------
-    Any
-        An object containing the contract's abi.
+    output_dr: str
```
dpaiton, line 31: missing "Returns" label?

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -1,120 +1,120 @@
...
+    ----------
+    contract_template : Template
+        A jinja template containging types for all structs within an abi.
+    abi_file_path : Path
+        The path to the abi file to parse.
 
+    Returns
+    -------
+    str
+        _description_
```
dpaiton, line 72: I'm not sure what `_description_` means?

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -1,120 +1,120 @@
...
-    return template
+    return contract_template.render(contract_name=contract_name, functions=function_datas)
 
 
-def get_abi_items(file_path: Path) -> list[ABIElement]:
-    """Gets all the
+def render_types_file(contract_name: str, types_template: Template, abi_file_path: Path) -> str:
+    """Returns a string of the types file to be generated.
 
     Parameters
```
dpaiton, line 100: Parameters --> Arguments

lib/pypechain/pypechain/run_pypechain.py
```diff
@@ -1,120 +1,120 @@
...
+    types_template : Template
+        A jinja template containging types for all structs within an abi.
+    abi_file_path : Path
+        The path to the abi file to parse.
 
     Returns
     -------
-    List[Union[ABIFunction, ABIEvent]]
+    str
         _description_
```
dpaiton, line 110: `_description_`?

lib/pypechain/pypechain/sample/ERC20ContractOriginal.py
```diff
@@ -3,22 +3,73 @@
...
+
+        Returns
+        -------
+            Returns the output of the contract.
+        """
+
+        result = super().call(transaction, block_identifier, state_override, ccip_read_enabled)
+        return result
+
+    # def transact(self, transaction: TxParams | None = None) -> HexBytes:
```
dpaiton, line 60: dead code

lib/pypechain/pypechain/utilities/types.py
```diff
@@ -1,6 +1,9 @@
 """Utilities to help with Solidity types."""
 
 
+from logging import warn
```
dpaiton, line 4: I think you want to do `logging.warning` here to use the logging setup everywhere else in the repo. Maybe warn is an alias for that? Either way, probably best to stay consistent with how we do it elsewhere.

dpaiton (approved):
a few typos & such, otherwise good

PR #867, opened 2023-09-13
update to current hyperdrive abi

dpaiton:


dpaiton:
I'm switching this to a draft until I can wrap my head around the solution.

Unfortunately, changing a variable name seems to have far-reaching consequences in `elfpy`. I'd rather just deprecate the elfpy lib, so I'm going towards that solution.

sentilesdal (approved):

dpaiton:
Ok, it's ready. I ran the full bot pipeline with the dashboard on `hyperdrive:0.0.13` and this branch.

PR #869, opened 2023-09-14
cleanly exit trade loop

wakamex:


dpaiton:
The pattern we've been using is to raise a special error and catch it. This avoids needing the extra return statement & check, and allows you to initiate the error at any point in the pipeline. We haven't ported any errors into agent0 yet, but you can see how we used to define them in elfpy: https://github.com/delvtech/elf-simulations/blob/main/lib/elfpy/elfpy/errors/errors.py

and how we define them in ethpy: https://github.com/delvtech/elf-simulations/blob/main/lib/ethpy/ethpy/base/errors/types.py

Can you follow this pattern, such that whenever you would set `exit_flag=True` you instead raise an "AgentFinished" exception? You can then catch this in the outer loop.

Another minor thing, it looks like you have this pattern:
```python
try:
  # a thing
except SomeError:
  # special thing for SomeError
except AllErrors:
  # do something else
```

This might work, I'm not sure, but it seems like a bad practice in general. We're already breaking protocol with the `except Exception` line. Does it make more sense for you to rework it so that you only have one `except` clause?
```python
try:
  # a thing
except Exception as exc:
  if exc is SomeError:
    # special thing for SomeError
  else:
    # do something else
```

wakamex:
I need the code to stop, so I can go on to do other stuff in my scripts. Returning an empty list continues the trade loop forever. I didn't think of using an Exception to signal an end. Perhaps we should have an "I am done trading" flag? Seems better than an exception, as multiple agents may signal they're each done trading, but we may want to wait until they all signal that, before ending the trade loop. It feels like using `while True` isn't ideal. It also means you can't return variables from `run_agents` which I was actually just trying to do yesterday, to avoid having to run `setup_experiments` twice so I can access the generated return outputs, which seems like something we generally shouldn't do, as it includes logging setup and creation of rng objects, among other things.

wakamex:
here's what my flow would look like with a DoneTrading Exception
https://github.com/wakamex/elf-simulations/pull/23

agreed just parsing my indexerror as "ran out of trades" is not good, so should have something clearer

dpaiton:
that DoneTrading exception looks good. Would it make sense to pass the agent address in the error message so that you can parse _who_ was done trading? Then you can track it and exit the loop once all agents are finished.

The new API might help you avoid running `setup_experiments` twice, lets revisit once that lands.

PR #876, opened 2023-09-19
Add events to ContractTypes.py

sentilesdal:
Adds events to the types output file.  The events are instantiated versions of ABIEvent, which is a TypedDict.  The output looks as follows:

```python
AddLiquidity = ABIEvent(
    anonymous = False,
    inputs = [
        ABIEventParams(indexed=False, name="baseAmount", type="uint256"),
    ],
    name = "AddLiquidity",
    type = "event",
)```

### CODE REVIEW

File: lib/pypechain/README.md [hunk 1]
```diff
@@ -1,4 +1,10 @@
-Pypechain generates web3.py Contract classes for a given solidity contract abi.
+# Pypechain
+
+Static python bindings for ethereum smart contracts.
+
+- Parses JSON ABI's to create typesafe web3.py contract instances
+- Functions have typesafe function parameters and return values
+- Smart Contract internal types are exposed as dataclasses
```
dpaiton, line 7: `internal structs`?

dpaiton (approved):
thanks :)

lib/pypechain/README.md
sentilesdal: troo.  will fix in followup.

PR #879, opened 2023-09-19
extract functional code from fixture

wakamex:
allow us to create hyperdrive deployments using the same code as the fixture, but outside of the fixture


lib/ethpy/ethpy/test_fixtures/deploy_hyperdrive.py
```diff
@@ -217,17 +217,20 @@ def deploy_and_initialize_hyperdrive(
...
         update_gap,
     )
-    tx_receipt = smart_contract_transact(
-        web3,
-        factory_contract,
-        deploy_account,
-        "deployAndInitialize",
-        # Function arguments
+    # Function arguments
+    fn_args = (
```
wakamex, line 221: readability change, this allows me to verify the function because I can see which parameters are for the python function and which are passed to the solidity function (fn_args)

lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -59,16 +59,13 @@ class LocalHyperdriveChain(NamedTuple):
...
+def create_hyperdrive_chain(rpc_uri: str, contract_addresses = None, initial_liquidity = None) -> LocalHyperdriveChain:
     """Initializes hyperdrive on a local anvil chain for testing.
-    Returns the hyperdrive contract address.
 
     Arguments
     ---------
-    local_chain: str
-        The `local_chain` test fixture that binds to the local anvil chain rpc url
+    rpc_uri: str
+        The URI of the local RPC node
```
slundqui, line 68: Docstrings of arguments please. Not sure why contract_addreses and initial_liquidity is being passed in

wakamex: my whole rationale for using this is to allow me to change the initial liquidity amount. I use it like this, to deploy the minimum liquidity possible: `create_hyperdrive_chain("http://127.0.0.1:8546", initial_liquidity=int(20*1e18))`

when the contract size wasn't allowed me to deploy the hyperdrive factory, I tried an alternate path: use the deployed factory to deploy a new hyperdrive pool. that's what contract_addresses is for. it passes in the existing deployment addresses, to look up the factory, and use it to deploy a new hyperdrive. it seems like saving that step would be nice to have. although the factory contains a lot of parameters you can't overwrite, like fees. I never got this flow to work, since apparently you also have to deploy a new vault for a new pool, which is currently lumped into `deploy_hyperdrive_factory`. should we remove this path for now, or leave it as a TODO?

lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -88,15 +86,24 @@ def local_hyperdrive_chain(local_chain: str) -> LocalHyperdriveChain:
...
     """
-
-    web3 = initialize_web3_with_http_provider(local_chain, reset_provider=False)
+    web3 = initialize_web3_with_http_provider(rpc_uri, reset_provider=False)
     account = initialize_deploy_account(web3)
-    base_token_contract, factory_contract = deploy_hyperdrive_factory(local_chain, account)
-    hyperdrive_addr = deploy_and_initialize_hyperdrive(web3, base_token_contract, factory_contract, account)
     abi_folder = "packages/hyperdrive/src/abis/"
     abis, _ = load_all_abis(abi_folder, return_bytecode=True)
+    if contract_addresses is None:
```
slundqui, line 93: Good to go after removing this

slundqui (approved):

lib/ethpy/ethpy/test_fixtures/local_chain.py
wakamex: removed unfinished flow, tracking it here: https://github.com/delvtech/elf-simulations/issues/907

wakamex: added docstrings inb822697107835a8f75c2f2e04a39d99dc0bb701f

PR #880, opened 2023-09-20
Streamlit performance optimizations

slundqui:
Adjusting streamlit to be more performant under high volume.

- Reducing dashboard view from 14400 blocks to 5000 blocks
- Moving groupby computation for positions over time to be on the sql side instead of pandas side in streamlit
- Shortening loopback limit so acquire_data doesn't break when chain spins faster than data pipeline
- Smaller fig sizes for faster plotting
- Explicit figure closing and garbage collection
- Removing warning printing on data collection from web3 for performance

Overall, there is still a memory leak, with a description, fixed within infra, and suggestions on longer term fixed detailed here: (https://github.com/delvtech/elf-simulations/issues/891)

### DIFF
--- a/lib/chainsync/chainsync/db/hyperdrive/interface.py
+++ b/lib/chainsync/chainsync/db/hyperdrive/interface.py
@@ -5,7 +5,7 @@ import logging
 
 import pandas as pd
 from chainsync.db.base import get_latest_block_number_from_table
-from sqlalchemy import exc
+from sqlalchemy import exc, func
 from sqlalchemy.orm import Session
 
 from .schema import (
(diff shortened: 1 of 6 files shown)


lib/chainsync/chainsync/db/hyperdrive/interface.py
```diff
@@ -785,6 +778,10 @@ def get_wallet_pnl(
     end_block : int | None, optional
         The ending block to filter the query on. end_block integers
         matches python slicing notation, e.g., list[:3], list[:-3]
+    wallet_address : list[str] | None, optional
+        The wallet addresses to filter the query on. Returns all if None.
+    return_timestamp : bool, optional
+        Returns the timestamp from the pool info table if True. Defaults to True.
```
dpaiton, line 784: 👍 thanks for improving docstrings!

dpaiton (approved):
Really solid code improvements!

PR #888, opened 2023-09-20
add hyperdrive types

dpaiton:
This PR adds the `hyperdrive_types` package to `lib`. This package is comprised of a shell script that recursively parses an `abi` folder for ABI JSON files, and calls `pypechain` to generate the corresponding python files. Due to problems described in the discussion below, we opted to also commit the programmatically generated python files to this repo.

Original description of the problem is in [this comment below](https://github.com/delvtech/elf-simulations/pull/888#issuecomment-1731833315).


slundqui:
I think there's a way to run whatever python code you need when you do pip install. Right now, we're using pyproject.toml, which handles that for you, but I think a replacement using setup.py allows you to run arbitrary python code during a pip install. Hence, you can imagine doing pip install pypechain (or having it in requirements.txt) run the codegen stuff. Unfortuantly, there's still the issue of pointing it to the abis

dpaiton:
> I think there's a way to run whatever python code you need when you do pip install. Right now, we're using pyproject.toml, which handles that for you, but I think a replacement using setup.py allows you to run arbitrary python code during a pip install. Hence, you can imagine doing pip install pypechain (or having it in requirements.txt) run the codegen stuff. Unfortuantly, there's still the issue of pointing it to the abis

This is a good option. We should consider 1) that will require modifying our install process everywhere else to be consistent. 2) this is the older method, and the PEPs linked in the PR explain why it's generally better to move to the new pyproject.toml system. 3) it's not clear that this is safer or more transparent or better in any way than a makefile or install.sh file

Pointing it to the ABIs should be fine since we're standardizing where they live (`<project_root>/packages/hyperdrive/src/abis`).

dpaiton:
@sentilesdal has suggested that we commit the generated code and leave it as a `pyproject.toml` package. This avoids most of the issues. We now have to be diligent about maintaining the correct version of the generated code & hyperdrive ABIs, which adds a step to the release process. This still seems like a better solution than what I had before, though.

slundqui:
I like this. You probably could write a test that regenerates the code and ensures the committed versions are correct.

sentilesdal (approved):

dpaiton:
I'm changing PR description to be of the solution, not the problem. Here's the original description so we have context for the discussion above:

---------------------

This PR is the result of two decision forks. First, when to generate pypechain files:

1. generate pypechain files at run-time -- any "`main.py`" entry-point script that needs pypechain will have to generate them at run-time. Library objects that could be instantiated (e.g. the Hyperdrive API) by external code will need to generate pypechain code on `__init__`. We will need to generate it in CI as well.

2. generate pypechain files when the other packages are installed. This makes it easier for everyone to access them & know where they are.

Both of these options suck. The first one is straight forward -- we don't want all of our library classes to be executing scripts to generate code files any time they're instantiated. The second one is more nuanced -- I don't know how to handle pypechain as a dependency in the event that one of these packages goes on PIP. If one of my library files (e.g. `ethpy`) imports a class generated by pypechain, then I can no longer install that library without first installing _and then running_ pypechain. This is not a typical workflow, and I don't know how to set up our package so that `pip install ethpy` takes care of this. (edit: We discuss using `setup.py` below, I was not aware that this would work if the project is on pypa but I guess that makes sense. So that is a viable option.)

Nevertheless, I'm going to go with the second option. This leads us to the next decision fork:

Importing files from outside a package root (e.g. a file inside `elf-simulations/lib/ethpy/` importing from another location at or above `elf-simulations/lib`) is not recommended at best, and will usually throw a run-time error (assuming the `pyproject.toml` is set up correctly). 

Therefore, if we wish to use files generated by pypechain in our libraries then we have two options:
1. duplicate the generated files inside of each lib at install time, so that they're in a consistent folder for relative imports
2. create an empty python library structure, generate the files in there before install, and then install the library

I decided on the second option to keep the code in a single location. I'm calling the library `hyperdrive_types`.

This solution still presents some challenges:
- pypechain has dependencies that need to be installed first
- pypechain must be installed to execute `run_pypechain.py`
- after `run_pypechain.py` is executed, we can then install `hyperdrive_types`

This means the install process is more complicated. It used to be:

```bash
python -m pip install --upgrade pip
python -m pip install --upgrade -r requirements.txt
```

and now it is:

```bash
python -m pip install --upgrade pip
python -m pip install -e lib/pypechain[base]
sh lib/hyperdrive_types/scripts/build_type_files.sh packages/hyperdrive/src/abis/
python -m pip install --upgrade -r requirements.txt
```

This whole series could be wrapped up in yet another shell script, although that is generally frowned upon to include arbitrary code execution during an install process (it's not explicitly said, but this is an underlying goal of PEPs [517](https://peps.python.org/pep-0517/) and [518](https://peps.python.org/pep-0518/)). What I mean is, we should probably avoid nesting executable shell scripts for a python package installer. A workaround here is to bundle everything into a single install script (so they don't have to dig deeper), or a makefile.

This PR is a first-draft solution for the problem. If anyone takes issue with it then we should discuss, I don't love this solution and am happy to hear alternative ideas.

PR #890, opened 2023-09-20
consistent naming of URIs

dpaiton:
Web3 uses the term "URI" instead of "URL", so this PR renames our variables to be consistent. The two terms are actually [not the same](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). A URL is a type of URI that specifies the asset (also known as a URN) and resource (for example, `http://`). Since we may not need to specify a resource, we use the more general "URI" term.

99% of the time we will be providing a URL and the distinction is all but pedantic. It only starts to matter if we say e.g. `my_url: URI = URI("address")` -- this is immediately contradictory, which occurs when we are creating variables that have the web3py `URI` type.

### DIFF
--- a/lib/ethpy/ethpy/eth_config.py
+++ b/lib/ethpy/ethpy/eth_config.py
@@ -22,9 +22,6 @@ class EthConfig:
         The path to the abi directory.
     """
 
-    # default values for local contracts
-    # Matching environment variables to search for
-    # pylint: disable=invalid-name
     artifacts_uri: URI | str = URI("http://localhost:8080")
     rpc_uri: URI | str = URI("http://localhost:8545")
     abi_dir: str = "./packages/hyperdrive/src/abis"
@@ -54,9 +51,9 @@ def build_eth_config() -> EthConfig:
 
     arg_dict = {}
     if artifacts_uri is not None:
-        arg_dict["ARTIFACTS_URI"] = artifacts_uri
+        arg_dict["artifacts_uri"] = artifacts_uri
     if rpc_uri is not None:
-        arg_dict["RPC_URI"] = rpc_uri
+        arg_dict["rpc_uri"] = rpc_uri
     if abi_dir is not None:
-        arg_dict["ABI_DIR"] = abi_dir
+        arg_dict["abi_dir"] = abi_dir
     return EthConfig(**arg_dict)
--- a/lib/ethpy/ethpy/eth_config.py
+++ b/lib/ethpy/ethpy/eth_config.py
@@ -54,9 +54,9 @@ def build_eth_config() -> EthConfig:
 
     arg_dict = {}
     if artifacts_uri is not None:
-        arg_dict["ARTIFACTS_URL"] = artifacts_uri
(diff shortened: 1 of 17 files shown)


lib/ethpy/ethpy/eth_config.py
```diff
@@ -41,15 +48,15 @@ def build_eth_config() -> EthConfig:
...
+    rpc_uri = os.getenv("RPC_URI")
     abi_dir = os.getenv("ABI_DIR")
 
     arg_dict = {}
-    if artifacts_url is not None:
-        arg_dict["ARTIFACTS_URL"] = artifacts_url
-    if rpc_url is not None:
-        arg_dict["RPC_URL"] = rpc_url
+    if artifacts_uri is not None:
+        arg_dict["ARTIFACTS_URL"] = artifacts_uri
```
slundqui, line 57: Bug here, arg_dict is setting this variable as `ARTIFACTS_URL`, which I found surprising that considering this PR changes the field for EthConfig to `artifacts_uri`

slundqui: Explicit testing says this should fail with unexpected keyword, so again, I have no idea how this is passing CI

lib/ethpy/ethpy/eth_config.py
```diff
@@ -41,15 +48,15 @@ def build_eth_config() -> EthConfig:
...
+    rpc_uri = os.getenv("RPC_URI")
     abi_dir = os.getenv("ABI_DIR")
 
     arg_dict = {}
-    if artifacts_url is not None:
-        arg_dict["ARTIFACTS_URL"] = artifacts_url
-    if rpc_url is not None:
-        arg_dict["RPC_URL"] = rpc_url
+    if artifacts_uri is not None:
+        arg_dict["ARTIFACTS_URI"] = artifacts_uri
```
slundqui, line 57: Bug is still here, `EthConfig` defines the member variables as lowercase right?

lib/ethpy/ethpy/eth_config.py
```diff
@@ -13,20 +14,26 @@ class EthConfig:
...
+        The uri of the artifacts server from which we get addresses.
+    rpc_uri: URI | str
+        The uri to the ethereum node.
+    abi_dir: str
+        The path to the abi directory.
     """
 
     # default values for local contracts
     # Matching environment variables to search for
     # pylint: disable=invalid-name
```
slundqui, line 27: This is no longer needed if member variables are renamed

lib/ethpy/ethpy/eth_config.py
dpaiton: Oops, yes good catch. Also a good example of why we need to bump up our test coverage -- class initializations should be tested pretty thoroughly.

slundqui: I think this is passing CI because we always explicitly set an eth config in tests. I assume this will break when actually running the pipeline.

slundqui (approved):
LGTM. I'll make the corresponding changes in infra for this.

PR #893, opened 2023-09-21
pypechain support for functions with multiple signatures

dpaiton:
A single contract function can have multiple signatures, which is not supported in Python.
For example, before this PR pypechain would produce something like this:
```
class my_class:
   def __init__(arg0: type0, arg1: type1):
       ... # do stuff for condition 0

class my_class:
    def __init__(arg1: type1):
        ... # do stuff for condition 1
```

We need to use optional arguments to support this in Python. For example:
```
class my_class
    def __init__(arg1: type1, arg0: type0 | None = None):
        if arg0 is None:
            ... # do stuff for condition 1
        else:
            ... # do stuff for condition 0
```

To work this out, here is the new template logic: 
```
loop over the signatures for this function 
  does this signature include any optional arguments?
  no:
    py: if all optionals are None, execute this signature
  yes:
    optional_args_in_sig = the optional arguments required by this signature
    optional_args_not_in_sig = optional arguments that are not present in this signature.
    are all of the optional arguments in this signature?
    yes:
      py: if sig_optional_args are not None, execute this signature
    no:
      py: if sig_optional_args are not None AND other_optional_args are None, execute this signature
```

I had to modify the output from run_pypechain.py to get that conditional to work.

I also fixed an ABI bug where a variable holding the dictionary of structs was persisting across multiple calls to `pypechain::run_pypechain::main`.

Smaller stuff: fixed lint errors, added docs, added type hints, fixed install dependencies, improved functionality with black

### DIFF
--- a/lib/pypechain/pypechain/run_pypechain.py
+++ b/lib/pypechain/pypechain/run_pypechain.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import os
+import re
 import sys
 from dataclasses import asdict
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


slundqui (approved):
LGTM

PR #898, opened 2023-09-25
Fix Event Params and conditional imports

sentilesdal:
There was a bug where we weren't including all the event params for events.  This is fixed.  Also, the imports for ABIEvent and ABIEventParams are now conditionally imported.

- make executable
- fix event params
- ignore difference to arguments for __call__
- regenerate


lib/hyperdrive_types/hyperdrive_types/ERC20MintableContract.py
```diff
@@ -23,6 +19,8 @@
 class ERC20MintableAllowanceContractFunction(ContractFunction):
     """ContractFunction for the allowance method."""
 
+    # super() call methods are generic, while our version adds values & types
+    # pylint: disable=arguments-differ
```
dpaiton, line 23: maybe we should just move this comment & disable to the top?

dpaiton (approved):
thanks for cleaning this up!

lib/hyperdrive_types/hyperdrive_types/ERC20MintableContract.py
sentilesdal: I actually do like it here, even though its repeated, so we can identify other areas this might show up.  same for other lint issues unless they really are globally annoying.

PR #908, opened 2023-09-27
Host parameter in example scripts

slundqui:
For ease of use, the script itself specifies the host of the chain to connect to when running bots. Additionally adding required host argument for fund agents script.


lib/agent0/bin/fund_agents_from_user_key.py
```diff
@@ -44,7 +48,8 @@
     if account_key_config.USER_KEY == "":
         raise ValueError("User key not provided as argument, and was not found in env file")
 
-    eth_config = build_eth_config()
```
dpaiton, line 47: Removing this means we no longer can use the .env file for funding agents, correct? Are you sure we want to remove that workflow?

slundqui: This is specifically for pointing to the host (the agents are still reading a generated .env file). I believe this is fine.

dpaiton (approved):

PR #909, opened 2023-09-27
replace elfpy market with ethpy HyperdriveInterface

dpaiton:
This PR modifies the agent0 policies & accounts to use the new HyperdriveInterface API instead of the elfpy HyperdriveMarket.

The new interface lacks some features (e.g. all of the yieldspace math and variable rates) that were required for the smart long and short policies, so we are deleting those for now. I made several new issues to detail follow-up work that is still needed. They are #910 #911 #913 

This PR also gets us very close to completely detaching elfpy from the rest of the monorepo, though the rest of the detachment should also happen in a follow-up PR.


lib/agent0/agent0/hyperdrive/policies/random_agent.py
```diff
@@ -59,14 +57,14 @@ def get_available_actions(
             all_available_actions.append(HyperdriveActionType.CLOSE_SHORT)
         if wallet.lp_tokens:
             all_available_actions.append(HyperdriveActionType.REMOVE_LIQUIDITY)
-        if wallet.withdraw_shares and market.market_state.withdraw_shares_ready_to_withdraw > 0:
+        if wallet.withdraw_shares and interface.pool_info["withdrawSharesReadyToWithdraw"] > 0:
```
slundqui, line 60: A bit confused on the logic here, you're checking the global pool_info parameter to see if there are withdraw_shares ready to withdraw for this wallet?

dpaiton: it is checking if this wallet has withdraw shares, and if there are withdraw shares that are ready to withdraw in the pool itself. It's possible that the wallet expects to get withdraw shares, but they're still tied up in the pool

slundqui (approved):
This looks great.

We chatted about this before, just putting this here for visibility.
One downside of passing in the interface object to agent policies is that our ideal workflow is we return a list of actions from the `action` method of policies, where it may be confusing to a policy implementer, since the interface also supports write operations (e.g., openLong) directly. One way we can fix this may be to split up the interface to be a read/write interface (I believe the frontend interface is doing this under the hood).

Not important for this PR or the trading competition.

PR #922, opened 2023-09-28
Removing develop mode from example bots scripts

slundqui:
https://github.com/delvtech/elf-simulations/issues/899


dpaiton (approved):
nice cleanup, thanks!

PR #929, opened 2023-09-29
check docker desktop socket

wakamex:


lib/chainsync/chainsync/test_fixtures/db_session.py
```diff
@@ -26,7 +27,14 @@ def psql_docker() -> Iterator[PostgresConfig]:
     Iterator[PostgresConfig]
         An iterator that yields a PostgresConfig
     """
-    client = docker.from_env()
+    home_dir = os.path.expanduser("~")
+    socket_path = Path(f"{home_dir}/.docker/desktop/docker.sock")
```
dpaiton, line 31: is this an OS independent path? I think you want to use [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join) instead of `/`

wakamex: it seems you're right, hardcoding forward slash is bad
Path is the more modern library and their `/` operator chooses the correct type of slash based on your OS
gonna use `socket_path = Path(home_dir) / ".docker" / "desktop" / "docker.sock"`

wakamex:
merged in #1076

PR #930, opened 2023-09-29
wrap json in StringIO so pandas doesn't complain

wakamex:
avoids this warning: `FutureWarning: Passing literal json to 'read_json' is deprecated and will be removed in a future versio
n. To read from a literal string, wrap it in a 'StringIO' object.`

checked to see if this was the reason my test was failing. it wasn't.


slundqui (approved):
LGTM

PR #931, opened 2023-09-29
Anaylsis batch hotfix

slundqui:
Previously, we look at the minimum block number of transactions, wallet deltas, and pool info to see the last block updated. However, there's a case where a large number of blocks tick up without any transactions. In this case, the next transaction will result in analysis attempting to process a large number of blocks, which breaks the analysis pipeline.

This PR resolves this issue, along with the race condition solved in https://github.com/delvtech/elf-simulations/pull/906 by ensuring we write pool_info as the last item, and using pool_info as the table to see how far along the data process is.

This also deprecates all current wallet info tables from the db in favor of using wallet deltas, as current wallet info needed pool_info data for its calculations.


dpaiton (approved):
Awesome! 🧹

PR #934, opened 2023-09-30
api hotfix for updating cache

dpaiton:


slundqui (approved):
LGTM

slundqui:
Unsure why this test is failing now and not before, was the hyperdrive abis updated between the writing of the test and now?

PR #936, opened 2023-10-01
Update to support Hyperdrive 0.0.14

dpaiton:


lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -587,39 +588,40 @@ def get_max_long(self, budget: FixedPoint) -> FixedPoint:
...
+            minimumShareReserves=str(self._contract_pool_config["minimumShareReserves"]),
+            minimumTransactionAmount=str(self._contract_pool_config["minimumTransactionAmount"]),
+            positionDuration=str(self._contract_pool_config["positionDuration"]),
+            checkpointDuration=str(self._contract_pool_config["checkpointDuration"]),
+            timeStretch=str(self._contract_pool_config["timeStretch"]),
             governance=self._contract_pool_config["governance"],
-            fee_collector=self._contract_pool_config["feeCollector"],
-            fees=Fees(
+            feeCollector=self._contract_pool_config["feeCollector"],
+            Fees=Fees(
```
slundqui, line 601: Is there a reason `Fees` here is capitalized?

slundqui (approved):
LGTM

lib/ethpy/ethpy/hyperdrive/api.py
slundqui: We can follow up with this later

PR #943, opened 2023-10-06
Supports clean exits from agents when done trading

slundqui:


lib/agent0/agent0/base/agents/eth_agent.py
```diff
@@ -88,4 +90,8 @@ def get_trades(self, interface: MarketInterface) -> list[Trade[MarketAction]]:
         list[Trade]
             List of Trade type objects that represent the trades to be made by this agent
         """
-        return self.policy.action(interface, self.wallet)
+        actions = list[Trade[MarketAction]]
```
wakamex, line 93: no need to instantiate this to an empty list

lib/agent0/agent0/base/agents/eth_agent.py
```diff
@@ -88,4 +90,8 @@ def get_trades(self, interface: MarketInterface) -> list[Trade[MarketAction]]:
...
+        done_trading: bool
```
wakamex, line 94: no need for type hints on return variables. we get them from the type hints on the results of the `action()` function.

dpaiton (approved):

Issue #896, opened 2023-09-22
create bot garden


wakamex:
MVP to allow end-users to submit their bot for inclusion in the monorepo, given a set of contributing guidelines. clean up the existing bots in preparation for the trading competition. include "info" string in BasePolicy which users have to fill in to subclass, that is sufficiently verbose for someone deciding whether to run this bot.

PR #974, opened 2023-10-11
model zoo

wakamex:
BasePolicy now has a `describe` method. This is only meant to be called by subclasses, passing in their raw_description which this will sanitize for them. For the subclasses, I've chosen a similarly named classmethod `description` to avoid pylint complaining about mismatched # of parameters (parent being describe(cls, raw_description) and subclass being describe(cls)). I would have liked to throw the `@property` descriptor on `description`, but then it can only be called from instantiated objects. Since the description of an agent has nothing to do with a specific instance, I prefer keeping it as a classmethod, with the slightly lengthier `description()` required to access it instead of just `description` without the parentheses. We only call this in the zoo.

renames RandomAgent to Random and ArbitragePolicy to Arbitrage. they live in the policies folder, so it's clear they're policies.

3 ways to use it:

describe all policies in zoo: `agent0.Zoo().describe()`
describe specific policy in zoo: `agent0.Zoo().describe("random")`
get description from specific class: `Random.description()`

output of describe all policies in zoo:
```
=== arbitrage ===
  Take advantage of deviations in the fixed rate from specified parameters.
  The following 3 parameters in Config define its operation:
  - When `high_fixed_rate_thresh`, open shorts are closed, and a long is opened.
  - When `low_fixed_rate_thresh`, open longs are closed, and a short is opened.
  - Trade size is fixed by `trade_amount`.
  Additionally, longs and shorts are closed if they are matured.
=== random ===
  A simple demonstration agent that chooses its actions randomly.
  It can take 7 actions: open/close longs and shorts, add/remove liquidity, and redeem withdraw shares.
  Trade size is randomly drawn from a normal distribution with mean of 10% of budget and standard deviation of 1% of budget.
  A close action picks a random open position of the given type (long or short) and attempts to close its entire size.
  Withdrawals of liquidity and redemption of withdrawal shares is sized the same as an open position: N(0.1, 0.01) * budget.
```

fixes: #896

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ We strive for verbose and readable comments and docstrings.
 Our documentation follows the [Numpy format](https://numpydoc.readthedocs.io/en/latest/format.html).
 The hosted docs are automatically generated using [Sphinx](https://www.sphinx-doc.org/en/master/tutorial/automatic-doc-generation.html).
 
-# Contributor git workflow:
+# Contributor git workflow
 
 We follow a standard [feature branch rebase workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow) that prioritizes short PRs with isolated improvements.
 Commits to `main` should **only** be made in the form of squash-merges from pull requests.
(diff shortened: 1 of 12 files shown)


slundqui (approved):
This looks great, can you put an example of the describe function output in the PR description?

PR #977, opened 2023-10-12
test interface

wakamex:
~~add~~duplicate test for hyperdrive interface
~~fix interface to use snake case instead of camel case~~
make `calc_fixed_rate` work with FixedPoint or int
code reuse through _stringify_pool_pool/config helper functions

### DIFF
--- a/lib/ethpy/tests/test_interface.py
+++ b/lib/ethpy/tests/test_interface.py
@@ -1,9 +1,7 @@
-from chainsync.analysis import calc_fixed_rate_fp, calc_spot_price
-import pytest
-import logging
-
+"""Test the hyperdrive interface."""
 from typing import cast
 
+from chainsync.analysis import calc_fixed_rate_fp, calc_spot_price
 from eth_typing import URI
 from ethpy import EthConfig
 from ethpy.hyperdrive.api import HyperdriveInterface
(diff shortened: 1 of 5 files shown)


lib/ethpy/tests/test_interface.py
```diff
@@ -0,0 +1,119 @@
...
+    expected_spot_price = calc_spot_price(
+        share_reserves=api_pool_info["shareReserves"],
+        share_adjustment=api_pool_info["shareAdjustment"],
+        bond_reserves=api_pool_info["bondReserves"],
+        initial_share_price=api_pool_config["initialSharePrice"],
+        time_stretch=api_pool_config["timeStretch"],
+    )
+
+    api_fixed_rate = interface.fixed_rate
+    expected_fixed_rate = calc_fixed_rate_fp(
```
slundqui, line 110: I'm hesitant to depend on a function in analysis in a test, I suspect this function will get replaced by the interface at some point.

wakamex: fixed in c906830f59f6d8e8d35d417b38818d4b431e9e5f

wakamex:
not sure anything here is needed:

1. thought there was an error. there wasn't. I just had the wrong pyperdrive installed.
2. didn't see the api_test.py which covers everything in here

switching to draft on Oct12, aiming to close by Oct19 if no use for this arises

wakamex:
deprecated by https://github.com/delvtech/agent0/pull/1184

wakamex:
deprecated by https://github.com/delvtech/agent0/pull/1184

PR #983, opened 2023-10-12
Funding bots now async with retries

slundqui:
This PR allows for funding eth, base, and base approvals with bots in async. This allows setup of bots on a chain mined via time quickly, with a large number of bots.

Part of this was a refactor to smart_contract_write calls to allow for a manual nonce parameter. This is needed for funding, as we launch multiple transactions at the same time. The previous method of getting the transaction count from web3 resulted in nonce errors.

The above also fixes an issue when an agent launches multiple trades on the same block.

Also fixes logging to the `.logging` directory

### DIFF
--- a/lib/agent0/agent0/base/config/environment_config.py
+++ b/lib/agent0/agent0/base/config/environment_config.py
@@ -26,7 +26,7 @@ class EnvironmentConfig(FrozenClass):
     # if halt_on_errors is True, halt_on_slippage controls if we halt when slippage happens
     halt_on_slippage: bool = False
     # optional output filename for logging
-    log_filename: str = "agent0-logs"
+    log_filename: str = ".logging/agent0_logs.log"
     # log level; should be in [logging.DEBUG, logging.INFO, logging.WARNING]
     log_level: int = DEFAULT_LOG_LEVEL  # INFO
     # delete_previous_logs; if True, delete existing logs at the start of the run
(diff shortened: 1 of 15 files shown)


lib/agent0/agent0/base/config/environment_config.py
```diff
@@ -26,7 +26,7 @@ class EnvironmentConfig(FrozenClass):
     # if halt_on_errors is True, halt_on_slippage controls if we halt when slippage happens
     halt_on_slippage: bool = False
     # optional output filename for logging
-    log_filename: str = "agent0-logs"
+    log_filename: str = ".logging/agent0_logs.log"
```
dpaiton, line 29: im surprised you need this. I was under the impression that `.logging/` and `.log` were added automatically by the logging lib. Did you verify that this is putting the log into the right spot?

lib/agent0/agent0/hyperdrive/exec/fund_agents.py
```diff
@@ -38,6 +47,10 @@ def fund_agents(
     contract_addresses: HyperdriveAddresses
         Configuration for defining various contract addresses.
     """
+
+    # Funding contains its own logging as this is typically ran from a script or in debug mode
```
dpaiton, line 51: run*

dpaiton (approved):

lib/agent0/agent0/base/config/environment_config.py
slundqui: Verified. Without the prefix, logging goes into the ether and is lost forever

PR #990, opened 2023-10-13
Adjusting sleep in checkpoint bot for sped up time

slundqui:


lib/agent0/bin/checkpoint_bot.py
```diff
@@ -21,11 +21,16 @@
...
 from fixedpointmath import FixedPoint
 from web3.contract.contract import Contract
+from web3.types import RPCEndpoint
 
 # The portion of the checkpoint that the bot will wait before attempting to
 # mint a new checkpoint.
 CHECKPOINT_WAITING_PERIOD = 0.5
 
+# TODO can't seem to get block time or block timestamp interval from anvil, so we code it here
+BLOCK_TIME = 6
```
jalextowle, line 32: This should probably be an environment variable.

jalextowle, line 32: Same here, can we make this an envvar?

jalextowle (approved):
LGTM once we parameterize this.

PR #992, opened 2023-10-16
Crash reporting now takes optional argument for log level

slundqui:


slundqui:
Fixes https://github.com/delvtech/elf-simulations/issues/967

dpaiton (approved):

PR #994, opened 2023-10-16
Add smart_long policy

sentilesdal:
Adds a new smart_long policy.  In order to do this I updated pyperdrive with new get_out_for_in and get_in_for_out methods.

- and update bonds_given_shares_and_rate- add get_out/in methods
- add new smart_long policy
- add new policy to zoo
- add description to smart_long policy
- add policy config to smart_long


lib/agent0/agent0/hyperdrive/policies/zoo.py
```diff
@@ -3,9 +3,11 @@
 
 from typing import NamedTuple
 
+from .arbitrage import Arbitrage
```
dpaiton, line 6: something got messed up with the import order. Can you delete the comment on line 8 & resort?

lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -11,31 +11,23 @@
...
-from ethpy.base import (
-    BaseInterface,
-    async_smart_contract_transact,
-    get_account_balance,
-    initialize_web3_with_http_provider,
-    load_all_abis,
-    smart_contract_preview_transaction,
-    smart_contract_read,
-)
+from ethpy.base import (BaseInterface, async_smart_contract_transact,
```
dpaiton, line 14: this is a new format that we don't use anywhere else in the code. I think it would be better to keep things consistent by leaving it alone. Not sure what changed on your end...

dpaiton (approved):

wakamex (commented):
lgtm

lib/agent0/agent0/hyperdrive/policies/zoo.py
wakamex: I sorted them in order of smarts 😜

PR #1002, opened 2023-10-17
docker utilities

wakamex:
mostly useful for scripts, where you don't want to repeatedly run manual docker commands as you tweak things.

usage:
```py
from agent0.utilities import check_docker
check_docker(restart=True)
```


lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+    Parameters
+    ----------
+    restart: bool
+        Restart docker even if it is running.
+    """
+    home_infra = Path(os.path.expanduser("~")) / "code" / "infra"
+    if os.path.exists(home_infra):
+        infra_folder = home_infra
+    else:
+        infra_folder = Path("/code/infra")
```
dpaiton, line 21: this should not be hard-coding the `/`, right?

lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+import logging
+import os
+import time
+from pathlib import Path
+
+
+def check_docker(restart: bool = False) -> None:
+    """Check whether docker is running to your liking.
+
+    Parameters
```
dpaiton, line 12: Arguments

lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+            _start_docker(f"{preamble_str}, restarting docker...", infra_folder)
+        else:
+            logging.info("%s, using them.", preamble_str)
+    else:
+        _start_docker("Starting docker.", infra_folder)
+    dockerps = os.popen("docker ps --format 'table {{.Names}}\t{{.Status}}\t{{.Ports}}'").read()
+    logging.info(dockerps)
+
+
+def _start_docker(startup_str: str, infra_folder: Path):
```
dpaiton, line 36: pls add docstrings

lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+    output = subprocess.getoutput(cmd)
+    if output is not None:
+        docker_pull_cmd = f"echo '{output}' | xargs -L1 docker pull"
+        _run_cmd(infra_folder, f" && {docker_pull_cmd}", "Updated docker in ")
+    else:
+        logging.info("No matching images found.")
+    _run_cmd(infra_folder, " && docker-compose up -d", "Started docker in ")
+
+
+def _run_cmd(infra_folder: Path, cmd: str, timing_str: str):
```
dpaiton, line 49: pls add docstrings

lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+    result = time.time()
+    os.system(f"cd {infra_folder}{cmd}")
+    formatted_str = f"{timing_str}{time.time() - result:.2f}s"
+    logging.info(formatted_str)
+    return result
+
+
+def _get_docker_ps_and_log() -> str:
```
dpaiton, line 57: pls add docstrings

lib/agent0/agent0/utilities/docker.py
dpaiton: & return type

lib/agent0/agent0/utilities/docker.py
dpaiton: _run_cmd has a return, so I think you should say `return _run_cmd(...)` and then include that return type in the signature for this func.

lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+
+def check_docker(restart: bool = False) -> None:
+    """Check whether docker is running to your liking.
+
+    Parameters
+    ----------
+    restart: bool
+        Restart docker even if it is running.
+    """
+    home_infra = Path(os.path.expanduser("~")) / "code" / "infra"
```
dpaiton, line 17: I don't think we should be hard-coding this path here if you want this utility to be useful for other people

lib/agent0/agent0/utilities/docker.py
wakamex: numpydoc says Parameters https://numpydoc.readthedocs.io/en/latest/format.html#parameters

lib/agent0/agent0/utilities/docker.py
```diff
@@ -0,0 +1,60 @@
...
+    if os.path.exists(home_infra):
+        infra_folder = home_infra
+    else:
+        infra_folder = Path("/code/infra")
+    dockerps = _get_docker_ps_and_log()
```
dpaiton, line 22: The first thing this should do is check if the caller has docker installed.  something like:
```python
from docker.errors import DockerException

  try:
      client = docker.from_env()
  # Skip this test if docker isn't installed
  except DockerException as exc:
```

lib/agent0/agent0/utilities/docker.py
wakamex: maybe I should pass it in as an optional local_path

dpaiton:
I'm not sure this should live in elf-simulations. It is too specific to your setup (hard-coded paths) and to our general infra setup, which will almost certainly not be shared by the broader user-base. I think this might be better placed inside the `infra` repo. Open to discussion, though.

lib/agent0/agent0/utilities/docker.py
wakamex: I also realize you could do this whole thing with the docker module, probably, instead of system calls... I can add a warning to check that docker's installed

wakamex: actually remembered my setup requires another workaround to make the docker module work https://github.com/delvtech/elf-simulations/commit/e3feaf19af566ba5500361111a41e9c4f65f0224

wakamex:
it seems this is mostly workarounds to workarounds of my local setup...
though if mine is a generic ubuntu setup...
do we want people to use the infra repo?

lib/agent0/agent0/utilities/docker.py
wakamex: removed in favor of passing in optional local path parameter in ea27a0edcbb15a0c5d5f6b03abc8d562274985bc

lib/agent0/agent0/utilities/docker.py
wakamex: added optional local path parameter in ea27a0edcbb15a0c5d5f6b03abc8d562274985bc

lib/agent0/agent0/utilities/docker.py
wakamex: added generic check for docker installation including that workaround in ea27a0edcbb15a0c5d5f6b03abc8d562274985bc

lib/agent0/agent0/utilities/docker.py
dpaiton: Yes, this is the one exception we have made from the numpy doc

lib/agent0/agent0/utilities/docker.py
wakamex: _run_cmd doesn't return anything though

wakamex: added docstrings in cab4ed101521656e90d10b1ce32b9da16dd5411c

lib/agent0/agent0/utilities/docker.py
wakamex: added docstrings in cab4ed101521656e90d10b1ce32b9da16dd5411c

lib/agent0/agent0/utilities/docker.py
wakamex: added docstrings in cab4ed101521656e90d10b1ce32b9da16dd5411c

wakamex:
not wanted at the moment, since I'm using this in only a couple of scripts that aren't committed here.

PR #1007, opened 2023-10-17
don't allow trades if bot doesn't have enough base

sentilesdal:
random bot was trying to LP below the minimum amount.  this fixes this for all trade actions the bot has.


slundqui (approved):
Looks great

PR #1008, opened 2023-10-17
add get_in_out funcs to interface

wakamex:


dpaiton (approved):
Approve,  but lets get input from Alex to verify

lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -259,6 +259,93 @@ def get_in_for_out(
...
+        tuple[FixedPoint, FixedPoint, FixedPoint] consisting of:
+            curve_fee : FixedPoint
+                Curve fee, in shares.
+            flat_fee : FixedPoint
+                Flat fee, in shares.
+            gov_fee : FixedPoint
+                Governance fee, in shares.
+        """
+        if maturity_time is None:
+            maturity_time: int = self.pool_config["positionDuration"]
```
jalextowle, line 292: I'm not sure about this line, but you have more context.

lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -259,6 +259,93 @@ def get_in_for_out(
...
+        normalized_time_remaining = time_remaining_in_seconds / self.pool_config["positionDuration"]
+        curve_fee = (
+            ((FixedPoint(1) / self.spot_price) - FixedPoint(1))
+            * self.pool_config["curveFee"]
+            * self.pool_config["initialSharePrice"]
+            * shares_in
+        )
+        flat_fee = (
+            shares_in * (FixedPoint(1) - normalized_time_remaining) * self.pool_config["flatFee"]
+        ) / self.pool_config["initialSharePrice"]
```
jalextowle, line 345: We don't have this in our implementation because a flat fee is never applied when we use this (it's only used in `openLong` so we know the trade is 100% on the curve): https://github.com/delvtech/hyperdrive/blob/main/contracts/src/HyperdriveBase.sol#L333.

lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -760,107 +847,58 @@ def get_max_long(self, budget: FixedPoint) -> FixedPoint:
...
+                    pool_info_str.sharePrice,
+                    checkpoint_exposure=str(self.latest_checkpoint["longExposure"].scaled_value),
+                    maybe_conservative_price=None,
+                    maybe_max_iterations=None,
+                )
+            )
         )
-        return FixedPoint(scaled_value=int(max_short))
+        bonds_in = self.get_in_for_out(amount_out=max_short / self.pool_config["sharePrice"], shares_out=True)
+        curve_fee, _, _ = self.calculate_fees_out_given_bonds_in(bonds_in=bonds_in)
```
jalextowle, line 902: We shouldn't need to do this if you're using the `get_max_short` from the Rust SDK. That function already takes fees into account. When I mentioned that we should "make sure that we're taking fees into account" I was only talking about the function that calculates the bond amount required to bring the rate to a targe.

lib/ethpy/ethpy/hyperdrive/api.py
wakamex: should be `self.current_block_time + self.pool_config["positionDuration"]`
so that after `time_remaining_in_seconds = maturity_time - self.current_block_time` resolves to `self.pool_config["positionDuration"]`
and after `normalized_time_remaining = time_remaining_in_seconds / self.pool_config["positionDuration"]` resolves to 1
can we skip these steps and just set `normalized_time_remaining` to 1? to avoid mistakes like these, and avoid mental math

dpaiton: I'd rather keep the pattern of setting maturity time to `self.current_block_time + self.pool_config["positionDuration"]`

lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -760,107 +847,58 @@ def get_max_long(self, budget: FixedPoint) -> FixedPoint:
...
+        curve_fee_in_base = curve_fee * self.pool_config["sharePrice"]
+        return max_short - curve_fee_in_base
```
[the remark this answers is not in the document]
dpaiton: Got it, we will remove.

lib/ethpy/ethpy/hyperdrive/api.py
dpaiton: Ok good to know. We decided to leave it in, just in case this gets used in a context outisde of `openLong`)

dpaiton (approved):

PR #1020, opened 2023-10-18
fix api checkpoint acquisition

dpaiton:


lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -205,6 +205,25 @@ def vault_shares(self) -> FixedPoint:
...
+        block_timestamp: int
+            A timestamp for any block. Use the latest block to get the current checkpoint id,
+            or a specific timestamp of a transaction's block if getting the checkpoint id for that transaction.
+
+        Returns
+        -------
+        int
+            The checkpoint id, which can be used as an argument for the Hyperdrive getCheckpoint function.
+        """
+        latest_checkpoint_timestamp = int(block_timestamp) - (
```
slundqui, line 222: Is this right? I thought block_timestamp was a timestamp and not a block number

dpaiton: ohh right good catch! here's the source:
https://github.com/delvtech/hyperdrive-monorepo/blob/64161c06c7ad90a5dcd46ca5cd570180842a4c6a/packages/hyperdrive-sdk/src/pool/getCheckpointId.ts

dpaiton: I opted to move everything to timestamp to be consistent with solidity & typescript API

slundqui (approved):

PR #1040, opened 2023-10-23
adds calc_bonds_given_rate_and_shares api function

dpaiton:


lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -399,14 +355,7 @@ def calc_reserves_to_hit_target_rate(
...
         latest_fixed_rate = calc_apr_local_dict(pool_config, pool_info)
-        target_bonds = calc_bond_reserves(
-            pool_info["shareReserves"],
-            pool_info["shareAdjustment"],
-            pool_config["initialSharePrice"],
-            target_rate,
-            pool_config["positionDuration"],
-            pool_config["invTimeStretch"],
-        )
+        target_bonds = interface.bonds_given_rate_and_shares(target_rate)
```
wakamex, line 358: this doesn't work with a changing `pool_info` object, right?

dpaiton: That's a good point -- I'm going to put it back to using your function. We can revisit after #1042 merges. I think after that it'll be an easy fix to incorporate optional modified pool info or current pool info.

lib/agent0/agent0/hyperdrive/policies/smart_long.py
```diff
@@ -131,7 +131,7 @@ def action(
         # only open a long if the fixed rate is higher than variable rate
         if (interface.fixed_rate - interface.variable_rate) > self.risk_threshold and not has_opened_long:
             # calculate the total number of bonds we want to see in the pool
-            total_bonds_to_match_variable_apr = interface.bonds_given_shares_and_rate(
+            total_bonds_to_match_variable_apr = interface.calculate_bonds_given_shares_and_rate(
```
slundqui, line 134: Didn't you rename these to `calc_...` or was that a different PR?

lib/ethpy/ethpy/hyperdrive/api.py
```diff
@@ -454,15 +468,22 @@ def bonds_given_shares_and_rate(self, target_rate: FixedPoint) -> FixedPoint:
...
-        one_over_tau: FixedPoint = self.pool_config["timeStretch"]
-        adjusted_apr = FixedPoint("1") + target_rate * t
-
-        return mu * z_minus_zeta * adjusted_apr**one_over_tau
+        if target_shares is None:
+            target_shares = self.effective_share_reserves
+        return FixedPoint(
+            scaled_value=int(
+                pyperdrive.calculate_bonds_given_shares_and_rate(  # pylint: disable=no-member
+                    str(target_shares.scaled_value),
```
slundqui, line 479: Is the int -> str mapping required for pyperdrive functions? Is there a way we can handle this under the hood in pyperdrive?

slundqui (approved):

lib/agent0/agent0/hyperdrive/policies/smart_long.py
dpaiton: that's in a follow-up PR reformatting the API.  And tbh I'm not sure if I'll even stick with it. I want it to be consistent across languages, but I dislike how it is in SOL & RS.

lib/ethpy/ethpy/hyperdrive/api.py
dpaiton: yes, discussed here: https://github.com/delvtech/elf-simulations/issues/905

PR #1042, opened 2023-10-23
refactor Hyperdrive API

dpaiton:
All changes are mechanical; no logic has been changed.

API is split from one file to a folder structure. The primary class is defined in `api/api.py`. Within that class, (almost) every function performs no logic outside of calling a corresponding, private, function that lives in another module in `api/`. The modules are organized into three groups:
- ABI calls (block getters)
- contract calls
- mock contract calculations (aka pyperdrive calls)

After everything this saves us about 300 lines of code from the original `api.py`.


lib/ethpy/ethpy/hyperdrive/api/api_test.py
```diff
@@ -9,9 +9,9 @@
 from fixedpointmath import FixedPoint
 from web3 import HTTPProvider
 
-from .addresses import HyperdriveAddresses
-from .api import HyperdriveInterface
-from .deploy import DeployedHyperdrivePool
+from ..addresses import HyperdriveAddresses
+from ..deploy import DeployedHyperdrivePool
```
slundqui, line 13: This looks a bit weird, are we sure we want to do relative imports from a parent directory?

dpaiton: fixed!

slundqui (approved):

PR #1049, opened 2023-10-25
Fixing arguments to smart contract functions in python

slundqui:


dpaiton:
I think there is some naming confusion between a [BlockIdentifier](https://github.com/ethereum/web3.py/blob/2db5fee85915154eac8f1554b90321d2c27db079/web3/types.py#L54) and a [BlockNumber](https://github.com/ethereum/eth-typing/blob/master/eth_typing/evm.py#L16). I think here you want the former, so we should use that type. They come from different sources:
```python
from eth_typing import BlockNumber
from web3.types import BlockIdentifier
```

lib/ethpy/ethpy/base/transactions.py
```diff
@@ -161,14 +171,13 @@ def retry_preview_check(exc: Exception) -> bool:
             retry_preview_check,
             function.call,
             {"from": signer_address},
-            **fn_kwargs,
+            block_identifier=block_identifier,
```
dpaiton, line 174: can you make the transaction dict more explicit and add a comment above line 169? This will help us down the road when we realize we need to add more kwargs to the `call` method, like max gas or whatever.

```python
# the transaction parameter is a kwarg dictionary used for Eth.send_transaction
transaction_kwargs = {"from": signer_address}
 return_values = retry_call(
            READ_RETRY_COUNT,
            retry_preview_check,
            function.call,
            transaction_kwargs,
            block_identifier=block_identifier,
. . .
```

dpaiton (approved):

PR #1062, opened 2023-10-30
mock open long and short

dpaiton:


lib/ethpy/ethpy/hyperdrive/api/api.py
```diff
@@ -482,12 +484,42 @@ async def async_redeem_withdraw_shares(
...
+        ---------
+        short_amount : FixedPoint
+            The amount to of bonds to short.
+
+        Returns
+        -------
+        short_amount : FixedPoint
+            The amount of base required to short the bonds (aka the "max loss").
+        """
+        return _calc_short_deposit(self, short_amount, self.spot_price, self.pool_info["sharePrice"])
```
slundqui, line 515: This seems the only place where this call is dependent on the internal state of the interface. I'd like the option to pass in my own pool information when calling rust's open short.

slundqui (approved):
LGTM

lib/ethpy/ethpy/hyperdrive/api/api.py
dpaiton: yes, agreed. I did it this way to keep things consistent. I have plans for a follow-up API refactor that facilitates ensuring everything is synced across pool info, block, checkpoint. This refactor will also include the ability to pass your own pool info. https://github.com/delvtech/elf-simulations/issues/1022

PR #1063, opened 2023-11-01
API updates for consistent and modifiable state

dpaiton:


lib/agent0/agent0/hyperdrive/crash_report/crash_report.py
```diff
@@ -109,14 +105,25 @@ def build_crash_trade_result(
...
+    Returns
+    -------
+    TradeResult
     """
     trade_result = TradeResult(
         status=TradeStatus.FAIL,
         agent=agent,
         trade_object=trade_object,
     )
+    pool_state = hyperdrive.current_pool_state
```
slundqui, line 124: Crash reporting is very dependent on getting data from the block in the exception if it's provided. This functionality has been removed with these changes.

slundqui (approved):

PR #1069, opened 2023-11-03
easier to adjust api pool state

dpaiton:
A PoolState object can now be constructed one of two ways.

----------------

using the hyperdrive interface itself to grab the actual state:
```python
from eth_typing import BlockNumber
from ethpy.hyperdrive.api import HyperdriveInterface

hyperdrive = HyperdriveInterface()

# get pool state for some arbitrary block number
pool_state = hyperdrive.get_hyperdrive_state(hyperdrive.get_block(BlockNumber(100)))

# get the current pool state
# this is cached, and also guarantees that all class attributes correspond to the same block
current_pool_state = hyperdrive.current_pool_state
```

----------------

or you can build it from scratch using data from any block or your own values:
```python
from ethpy.hyperdrive.transactions import (
    get_hyperdrive_checkpoint,
    get_hyperdrive_pool_config,
    get_hyperdrive_pool_info
)
from ethpy.hyperdrive.api import HyperdriveInterface, PoolState

hyperdrive = HyperdriveInterface()

# Could use any block as a reference
block = hyperdrive.get_current_block()
block_number = hyperdrive.get_block_number(block)

# we will use actual pool values, but these dictionaries can have any value as long as the keys & types are correct
contract_pool_config = get_hyperdrive_pool_config(hyperdrive.hyperdrive_contract)
contract_pool_info = get_hyperdrive_pool_info(hyperdrive.hyperdrive_contract, block_number)
contract_checkpoint = get_hyperdrive_checkpoint(
    hyperdrive.hyperdrive_contract,
    hyperdrive.calc_checkpoint_id(contract_pool_config["checkpointDuration"], hyperdrive.get_block_timestamp(block)),
)
variable_rate = hyperdrive.get_variable_rate(block_number)
vault_shares = hyperdrive.get_vault_shares(block_number)
total_supply_withdrawal_shares = hyperdrive.get_total_supply_withdrawal_shares(block_number)

# could change any of these when constructing the object
pool_state = PoolState(
            block,
            contract_pool_config,
            contract_pool_info,
            contract_checkpoint,
            variable_rate,
            vault_shares,
            total_supply_withdrawal_shares,
        )
```

----------------

Once you have the pool_state, you can call any of the underlying hyperdrive helper functions, e.g.

```python
bonds = hyperdrive.calc_bonds_given_shares_and_rate(
    target_rate=FixedPoint("0.05"),
    target_shares=FixedPoint(1*10**6),
    pool_state=pool_state,
)
```

----------------

I also made some minor updates:
- function & variable names are changed to make it more clear that they're doing RPC calls under the hood
- added some API helper functions to expose more private calculation & getter methods
- moved manual repeated code for calculating the total_supply_withdrawal_shares into a private calculation method

### DIFF
--- a/lib/agent0/agent0/hyperdrive/crash_report/crash_report.py
+++ b/lib/agent0/agent0/hyperdrive/crash_report/crash_report.py
@@ -152,7 +152,7 @@ def build_crash_trade_result(
         }
 
     ## Get the pool state at the desired block number
-    pool_state = hyperdrive.get_hyperdrive_state(hyperdrive.block(trade_result.block_number))
+    pool_state = hyperdrive.get_hyperdrive_state(hyperdrive.get_block(trade_result.block_number))
 
     ## Get pool config
     # Pool config is static, so we can get it from the interface here
--- a/lib/agent0/agent0/hyperdrive/crash_report/crash_report.py
+++ b/lib/agent0/agent0/hyperdrive/crash_report/crash_report.py
@@ -118,7 +118,7 @@ def build_crash_trade_result(
         agent=agent,
         trade_object=trade_object,
     )
-    current_block_number = hyperdrive.block_number(hyperdrive.current_block)
+    current_block_number = hyperdrive.block_number(hyperdrive.get_current_block())
 
     ## Check if the exception came from a contract call & determine block number
     # If it did, we fill various trade result data with custom data from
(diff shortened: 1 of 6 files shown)


slundqui (approved):
LGTM

Wondering if constructing a new pool state should allow for None values with checks in the interface functions on which fields the function needs, so it allows someone to not have to construct an entire state for a certain function that only requires a single piece of information

PR #1074, opened 2023-11-08
Fix Hyperdrive API docstrings

dpaiton:


lib/ethpy/ethpy/hyperdrive/api/api.py
```diff
@@ -777,11 +853,11 @@ def calc_bonds_given_shares_and_rate(
...
         and a given fixed rate.
 
-
-        The calculation is based on the formula:
+        The function does not perform contract calls, but instead relies on the Hyperdrive-rust sdk
+        to simulate the contract outputs. The calculation is based on the formula:
 
         .. math::
-            mu * (z - zeta) * (1 + apr * t) ** (1 / tau)
+            \mu * (z - \zeta) * (1 + \text{apr} * t)^{1 / \tau}
```
wakamex, line 860: we don't define tau here, but neither does the rust sdk. mental note to attempt to remove all mentions of tau in a future cleanup pass-through.

wakamex (approved):
lgtm

lib/ethpy/ethpy/hyperdrive/api/api.py
dpaiton: yeah agreed -- we should make this consistent across sol -> rust -> python

PR #1078, opened 2023-11-09
test bots hitting target rate

wakamex:
open 100k long and short, then hit the target rate
image

### DIFF
--- a/lib/agent0/agent0/hyperdrive/policies/deterministic.py
+++ b/lib/agent0/agent0/hyperdrive/policies/deterministic.py
@@ -1,5 +1,4 @@
 """Deterministically trade things."""
-# pylint: disable=line-too-long, missing-class-docstring
 from __future__ import annotations
 
 import logging
--- a/lib/agent0/agent0/hyperdrive/policies/deterministic.py
+++ b/lib/agent0/agent0/hyperdrive/policies/deterministic.py
@@ -1,0 +1,111 @@
+"""Deterministically trade things."""
+# pylint: disable=line-too-long, missing-class-docstring
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from fixedpointmath import FixedPoint
+
+from agent0.hyperdrive.state import HyperdriveActionType, HyperdriveMarketAction
+from elfpy.types import MarketType, Trade
+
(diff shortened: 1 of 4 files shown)


lib/agent0/agent0/hyperdrive/policies/deterministic.py
```diff
@@ -0,0 +1,111 @@
+"""Deterministically trade things."""
+# pylint: disable=line-too-long, missing-class-docstring
```
dpaiton, line 2: why disable these?

lib/agent0/tests/target_rate_test.py
```diff
@@ -0,0 +1,138 @@
...
+    ]
+    run_agents(
+        env_config,
+        agent_config,
+        account_key_config=build_account_key_config_from_agent_config(agent_config, random_seed=1),
+        eth_config=eth_config,
+        contract_addresses=hyperdrive_contract_addresses,
+    )
+
+    # One deterministic bot to hit the variable rate
```
dpaiton, line 88: should this be "One _arb_ bot to hit the variable rate"?

lib/agent0/tests/target_rate_test.py
```diff
@@ -0,0 +1,138 @@
...
+    ]
+    run_agents(
+        env_config,
+        agent_config,
+        account_key_config=build_account_key_config_from_agent_config(agent_config, random_seed=2),
+        eth_config=eth_config,
+        contract_addresses=hyperdrive_contract_addresses,
+    )
+
+    # Run acquire data to get data from chain to db
```
dpaiton, line 111: I think it makes more sense to use the HyperdriveInterface to get the variable rate & fixed rate. The current approach brings in a lot of other machinery into the test, making it less of a unit test. This means it could be harder to debug test failures, since so much code is being executed.

dpaiton (approved):
Sweet, glad to have this tested. Thanks!

lib/agent0/agent0/hyperdrive/policies/deterministic.py
wakamex: good catch, they're from an old version where this was like 5 (very long) lines

wakamex: good catch, they're from an old version where this was like 5 (very long) lines

PR #1079, opened 2023-11-09
Ethpy transaction hotfixes

slundqui:
1. Adding in custom non-async wait for transaction function that implements exponential backoff.
2. Transactions sometimes don't return logs (e.g., the `checkpoint` call), so don't throw an error when no logs are found.
3. Redeem withdrawal shares with an invalid amount is a valid contract call, so we explicitly check for zero values in the preview call on our side and throw an error if we detect this case
4. Adding test that goes through invalid trades

### DIFF
--- a/lib/ethpy/ethpy/base/transactions.py
+++ b/lib/ethpy/ethpy/base/transactions.py
@@ -147,10 +147,6 @@ def smart_contract_preview_transaction(
     else:
         function = contract.get_function_by_name(function_name_or_signature)(*fn_args, **fn_kwargs)
 
-    # We build the raw transaction here in case of error. Note that we don't call `build_transaction`
-    # since it adds the nonce to the transaction, and we ignore nonce in preview
-    raw_txn = function.build_transaction({"from": signer_address})
-
     # We define the function to check the exception to retry on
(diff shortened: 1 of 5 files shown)


dpaiton (approved):

lib/ethpy/ethpy/base/transactions.py
```diff
@@ -219,6 +221,53 @@ def retry_preview_check(exc: Exception) -> bool:
     return {f"value{idx}": value for idx, value in enumerate(return_values)}
 
 
+def wait_for_transaction_receipt(
+    web3: Web3, transaction_hash: HexBytes, timeout: float = 30, start_latency: float = 1, backoff: float = 2
+) -> TxReceipt:
+    """wait_for_transaction_receipt with exponential backoff
```
dpaiton, line 227: can you make this `web3.wait_for_transaction_receipt`

dpaiton (approved):

lib/ethpy/ethpy/base/transactions.py
slundqui: web3's version doesn't have exp backoff. I'm using that here.

PR #1081, opened 2023-11-09
separate local chain fixture logic

wakamex:
separates logic from fixtures.
fixtures call `launch_` methods which can be used directly from tests:
- `local_chain` calls `launch_local_chain`
- `local_hyperdrive_pool` calls `launch_local_hyperdrive_pool`

### DIFF
--- a/lib/ethpy/ethpy/test_fixtures/local_chain.py
+++ b/lib/ethpy/ethpy/test_fixtures/local_chain.py
@@ -27,7 +27,7 @@ def launch_local_chain(anvil_port: int = 9999, host: str = "127.0.0.1"):
     str
         The local anvil chain URI.
     """
-    anvil_process = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -6,51 +6,88 @@
...
-        Yields the local anvil chain URI
-    """
-    anvil_port = 9999
-    host = "127.0.0.1"  # localhost
+def launch_local_chain(anvil_port: int = 9999, host: str = "127.0.0.1"):
+    """Launch a local anvil chain.
 
-    # Assuming anvil command is accessible in path
-    # running into issue with contract size without --code-size-limit arg
+    Parameters
```
dpaiton, line 18: Arguments

dpaiton (approved):
Awesome, thanks! I think this will be useful for the notebook workflow.

PR #1082, opened 2023-11-09
move trade, market type, WEI, json encoder out of elfpy

dpaiton:


lib/hyperlogs/hyperlogs/__init__.py
```diff
@@ -2,6 +2,8 @@
 
 import logging
 
+from .json_encoder import ExtendedJSONEncoder
```
slundqui, line 5: I don't understand why we need a separate library just for logs, it looks like the only thing using this library is agent0's crash reporting. Shouldn't we just put this in agent0?

slundqui: Oh I see, there was another PR that moved all logging stuff out.

slundqui (approved):

PR #1092, opened 2023-11-15
Agent wallet state test against chain

slundqui:
Adding system test that checks the agent's wallet (calculated from deltas) against the value on the chain

### DIFF
--- a/tests/bot_wallet_test.py
+++ b/tests/bot_wallet_test.py
@@ -1,0 +1,308 @@
+"""System test for checking calculated wallets versus wallets on chain"""
+from __future__ import annotations
+
+import logging
+import os
+from dataclasses import dataclass
+from typing import cast
+
+from agent0 import build_account_key_config_from_agent_config
+from agent0.base import MarketType, Trade
+from agent0.base.config import AgentConfig, EnvironmentConfig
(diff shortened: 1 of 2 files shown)


tests/bot_wallet_test.py
```diff
@@ -0,0 +1,303 @@
...
+
+    Arguments
+    ---------
+    wallet: HyperdriveWallet
+        The HyperdriveWallet object to check against the chain
+    hyperdrive: HyperdriveInterface
+        The Hyperdrive API interface object
+    """
+    # Check base
+    base_from_chain = smart_contract_read(hyperdrive.base_token_contract, "balanceOf", wallet.address)["value"]
```
dpaiton, line 39: FYI there is a [hyperdrive interface function](https://github.com/delvtech/elf-simulations/blob/main/lib/ethpy/ethpy/hyperdrive/api/api.py#L338) for this.

dpaiton: Although looking at the rest of the code it probably makes sense to leave it as a low-level call

slundqui: Right, I think I considered using this, but I didn't have access here to the agent object.

PR #1099, opened 2023-11-16
moves stateful variables into their own folder & centralize types

dpaiton:
This PR cleans up the stateful API objects so that type conversions are hidden from the end user.

- `ethpy/hyperdrive` now includes a `state` folder that holds the FixedPoint versions of Checkpoint, PoolInfo, and PoolState.
- PoolState has been refactored to work primarily with the FixedPoint objects. The contract object is accessed via conversions.
- state type conversion functions have been centralized to a single spot inside the `state` folder.
- state type conversions have been localized to _mock_contracts (FixedPoint -> HyperTypes) and transactions (dict -> FixedPoint)


slundqui (approved):
LGTM

PR #1101, opened 2023-11-17
Interactive Agent0

slundqui:
This PR introduces interactive hyperdrive that is tailored to an end user running experiments in hyperdrive. To this end, we focus on syntax and usability for interacting with hyperdrive. Below is a code snippet that showcases how to use the interactive hyperdrive.

```
# Assumes anvil and docker is installed in your environment
from agent0.hyperdrive.interactive import InteractiveHyperdrive, LocalChain
from datetime import timedelta

# Launches a local anvil chain in a subprocess
chain = LocalChain()

# Initializes hyperdrive on the chain
interactive_hyperdrive = InteractiveHyperdrive(chain)

# Initializes a funded agent
hyperdrive_agent = interactive_hyperdrive.init_agent(base=FixedPoint(100000), eth=FixedPoint(100), name="alice")

# Make trades, returning the emitted event
open_long_event = hyperdrive_agent.open_long(base=FixedPoint(11111))

# Advance time
chain.advance_time(timedelta(weeks=52))

# Close the long
close_long_event = hyperdrive_agent.close_long(open_long_event.maturity_time, bonds=open_long_event.bond_amount)

# View all past trades from the database
print(interactive_hyperdrive.get_ticker())
```

Major changes:
- Implemented interactive hyperdrive and associated classes and tests in `agent0/hyperdrive/interactive/`

Minor changes:
- Exposing `set_max_approval` for usage within interactive hyperdrive
- Moving known error checking functions to be within crash reporting directory
- Moving known error checks one level in to `async_execute_agent_trades`
- Expanding receipt breakdown to include all possible fields events emit
- `async_match_contract_call_to_trade` now also returns the receipt breakdown (i.e., the events)
- Adding `maturity_time` to the `Long` and `Short` dataclasses
- Flag to make the database within postgres if it doesn't exist
- Removing extra sleeps in data pipeline

Bug fixes:
- `set_max_approval` now properly throws an error if it errors out
- Surrounding script files in `lib/agent0/examples` with `if __name__ == "__main__"`


lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
```diff
@@ -68,7 +61,9 @@ async def async_execute_single_agent_trade(
     # TODO preliminary search shows async tasks has very low overhead:
     # https://stackoverflow.com/questions/55761652/what-is-the-overhead-of-an-asyncio-task
     # However, should probably test what the limit number of trades an agent can make in one block
-    wallet_deltas_or_exception: list[HyperdriveWalletDeltas | BaseException] = await asyncio.gather(
+    wallet_deltas_or_exception: list[
+        Tuple[HyperdriveWalletDeltas, ReceiptBreakdown] | BaseException
```
dpaiton, line 65: are you sure you need `Tuple` from typing here? I think `tuple[` will work as long as `annotations` is imported.

lib/agent0/tests/interactive_hyperdrive_test.py
```diff
@@ -0,0 +1,190 @@
...
+        interactive_hyperdrive: InteractiveHyperdrive,
+        agent_wallet: HyperdriveWallet,
+    ):
+        # NOTE this function is assuming only one agent is making trades
+
+        # Test against db
+        current_wallet_df = interactive_hyperdrive.get_current_wallet(coerce_float=False)
+
+        # Check base
+        # TODO when current wallet gets fixed to represent current position, we can check base
```
dpaiton, line 24: can you add an issue citation for this? I'm not sure what you're referring to.

lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
dpaiton: Confirmed, you can replace `Tuple` with `tuple` in this entire file and remove the typing import `Tuple` because we have `from __future__ import annotations`. In python 3.11 this is the expected behavior, and we no longer need to import `annotations`.

lib/agent0/tests/interactive_hyperdrive_test.py
```diff
@@ -0,0 +1,190 @@
+"""Tests interactive hyperdrive end to end"""
```
dpaiton, line 1: we're moving to a pattern of putting tests with the source files they are testing, unless they are integrations or systems tests. I think this qualifies as one that should live in the `interactive/` folder, but if you disagree I'm fine with leaving it here.

lib/agent0/examples/interactive_hyperdrive_example.py
```diff
@@ -0,0 +1,104 @@
+"""Example script for using interactive hyperdrive."""
+import datetime
+
+from fixedpointmath import FixedPoint
+
+from agent0.hyperdrive.interactive import InteractiveHyperdrive, LocalChain
+
+if __name__ == "__main__":
```
dpaiton, line 8: I think you should pull this out of a `__name__ == "__main__"` conditional and instead wrap sections in `#%%` comment blocks so that they can be executed as a jupyter notebook in vscode _or_ as a python script.

dpaiton: I think the same about the other examples fwiw. I'd like to lean into the vscode environment because that's what everyone at delv uses. We can should discuss an easier solution for sharing this more broadly, but for now I like being able to run the examples via a notebook in vscode or via `python examples/file.py`.

lib/agent0/agent0/utils.py
```diff
@@ -0,0 +1,12 @@
+"""Various agent0 utility functions."""
```
dpaiton, line 1: I think we should rename `agent0/test_fixtures` to `agent0/test_utils` and then add a file called `assert_never.py` which contains the contents of this file. So the tree would be
```
agent0/
  > test_utils/
    - __init__.py # includes assert_never
    - assert_never.py # contents of current agent0/utils.py
    - cycle_trade_policy.py # untouched
```

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_agent.py
```diff
@@ -0,0 +1,208 @@
...
+        eth: FixedPoint
+            The amount of ETH to fund the agent with.
+        name: str
+            The name of the agent. Defaults to the wallet address.
+        pool: InteractiveHyperdrive
+            The pool object that this agent belongs to.
+        """
+        self._pool = pool
+        self.name = name
+        # TODO
```
dpaiton, line 55: dead comment?

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_agent.py
```diff
@@ -0,0 +1,208 @@
...
+        ---------
+        checkpoint_time : int
+            The checkpoint time to mint
+
+        Returns
+        -------
+        CreateCheckpoint
+            The emitted event of the checkpoint call.
+        """
+        # TODO underlying function will raise NotImplementedError, fix
```
dpaiton, line 98: https://github.com/delvtech/agent0/issues/1105

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_agent.py
```diff
@@ -0,0 +1,208 @@
...
+            The amount of longs to open in units of base.
+
+        Returns
+        -------
+        OpenLong
+            The emitted event of the open long call.
+        """
+        return self._pool._open_long(self.agent, base)
+
+    def close_long(self, maturity_time: int, bonds: FixedPoint) -> CloseLong:
```
dpaiton, line 116: This should be in units of base. But, also:

After an wonderful journey through `InteractiveHyperdrive -> async_execute_agent_trades -> async_execute_single_agent_trade -> async_match_contract_call_to_trade -> HyperdriveInterface` you will find that we end up actually calling the `api::_contract_calls::_async_close_long` function via `HyperdriveInterface.async_close_long`. We are passing that interface in as an argument way back at the top of this stack, when we know that it's an individual `close_long` trade. I think the chosen path is so that we can get access to all of the error handling. But that leads me to wonder if we should be taking that path directly from the API in the first place?

If you think it is easy to fix this wiring go ahead, otherwise please make an issue for it.

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_agent.py
```diff
@@ -0,0 +1,208 @@
...
+            The amount of longs to close in units of bonds.
+
+        Returns
+        -------
+        CloseLong
+            The emitted event of the close long call.
+        """
+        return self._pool._close_long(self.agent, maturity_time, bonds)
+
+    def open_short(self, bonds: FixedPoint) -> OpenShort:
```
dpaiton, line 133: this should be units of base

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_agent.py
```diff
@@ -0,0 +1,208 @@
...
+            The amount of shorts to open in units of bonds.
+
+        Returns
+        -------
+        OpenShort
+            The emitted event of the open short call.
+        """
+        return self._pool._open_short(self.agent, bonds)
+
+    def close_short(self, maturity_time: int, bonds: FixedPoint) -> CloseShort:
```
dpaiton, line 148: this should be units of base

dpaiton (approved):
This is great! I left a bunch of comments; I'm happy if you want to convert them into issues instead of resolving.

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_agent.py
dpaiton: [#1107](https://github.com/delvtech/agent0/issues/1107)

PR #1103, opened 2023-11-17
random agent policy test

dpaiton:


sentilesdal (approved):
lgtm, I assume you didn't add asserts on purpose?  You're just checking if the trades cause a revert?

PR #1104, opened 2023-11-20
Update Hypertypes

sentilesdal:
A first step to consolidate contract types including PoolInfo, PoolConfig etc.  Pulled things into the hypertypes library that belong there and cleaned up some imports.  One thing of note is that the __init__.py file in the new types/ dir was written manually, so we'll need to preserve that if we regenerate types.

- update hypertypes
- update imports
- update pool_info and pool_config calls
- remove unused function
- move types to types/ dir
- update imports
- use hypertypes contract
- move conversion utilities
- move fixedpoint types to their own file


lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -145,22 +146,36 @@ def apply_step(
...
+        ---------
 
 
-    Returns
-    -------
-    pool_state : PoolState
-        The updated pool state.
+
+
+    # TODO: These dataclasses are similar to pypechain except for
```
dpaiton, line 155: Can you move this TODO elsewhere? It'll mess up the docs generation

dpaiton: Actually why is this here at all?

lib/agent0/bin/checkpoint_bot.py
```diff
@@ -18,6 +18,7 @@
 from ethpy.hyperdrive import fetch_hyperdrive_address_from_uri, get_hyperdrive_pool_config
 from fixedpointmath import FixedPoint
 from hyperlogs import logs
+from hypertypes.types import IERC4626HyperdriveContract
```
dpaiton, line 21: Shouldn't this just be `from hypertypes import IERC4626HyperdriveContract`?

lib/ethpy/ethpy/hyperdrive/state/__init__.py
```diff
@@ -1,6 +1,5 @@
 """Hyperdrive state classes and conversion helper functions."""
+from hypertypes.fixedpoint_types import FeesFP, PoolConfigFP, PoolInfoFP
```
dpaiton, line 2: I don't think this should be here.

dpaiton: everywhere that this is used (`FeesFP, PoolConfigFP, PoolInfoFP` are imported from `ethpy.hyperdrive.state` should be rewritten to directly import it from `hypertypes.fixedpoint_types`

lib/ethpy/ethpy/hyperdrive/state/conversions.py
```diff
@@ -31,15 +28,15 @@ def snake_to_camel(snake_string: str) -> str:
 
 
 def dataclass_to_dict(
```
dpaiton, line 30: This function is basically "hypertypes_or_fixedpoint_state_to_contract"; in other words, it converts the hypertypes _or_ fixedpoint versions of pool info, pool config, & checkpoint into a dictionary that should look exactly like what the contract web3 read call returns. I think it makes sense to keep this iwth the other conversions. Then you can delete this file all-together and move the test as well.

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
sentilesdal: i think i had a vim snafu, fixed.

lib/ethpy/ethpy/hyperdrive/state/__init__.py
sentilesdal: good catch, this happened when vscode moved the symbols.  i removed this from the init

lib/ethpy/ethpy/hyperdrive/state/conversions.py
sentilesdal: ok sounds good.

lib/ethpy/ethpy/hyperdrive/state/pool_state.py
```diff
@@ -5,17 +5,15 @@
...
+from hypertypes.utilities.conversions import (
     fixedpoint_checkpoint_to_hypertypes,
     fixedpoint_pool_config_to_hypertypes,
     fixedpoint_pool_info_to_hypertypes,
 )
-from .pool_config import PoolConfig
-from .pool_info import PoolInfo
+from web3.types import BlockData
+
+from .....hypertypes.hypertypes.utilities.conversions import dataclass_to_dict
```
dpaiton, line 16: this shouldn't be a relative import

lib/ethpy/ethpy/hyperdrive/transactions.py
```diff
@@ -55,14 +49,13 @@ def get_hyperdrive_pool_info(hyperdrive_contract: Contract, block_number: BlockN
     dict[str, Any]
         A dictionary containing the Hyperdrive pool info returned from the smart contract.
     """
-    return hypertypes_pool_info_to_fixedpoint(
-        contract_pool_info_to_hypertypes(
-            smart_contract_read(hyperdrive_contract, "getPoolInfo", block_number=block_number)
-        )
-    )
+    pool_info = hyperdrive_contract.functions.getPoolInfo().call(None, block_number)
```
dpaiton, line 52: 🤩

lib/ethpy/ethpy/hyperdrive/state/conversions.py
```diff
@@ -2,19 +2,6 @@
...
-from hypertypes.IHyperdriveTypes import PoolConfig as HtPoolConfig
-from hypertypes.IHyperdriveTypes import PoolInfo as HtPoolInfo
-
-from .checkpoint import Checkpoint
-from .fees import Fees
-from .pool_config import PoolConfig
-from .pool_info import PoolInfo
 
 
 def camel_to_snake(snake_string: str) -> str:
```
dpaiton, line 7: `camel_to_snake` is only used here, `ethpy/hyperdrive/addresses.py` and `chainsync/db/hyperdrive/convert_data.py`. `snake_to_camel` is only used here. So I think it makes sense for addresses.py and convert_data.py to use the new `hypertypes` functions and then you can delete this file entirely.

lib/ethpy/ethpy/hyperdrive/transactions.py
slundqui: BTW, this looks very much like what web3 already exposes:
https://web3py.readthedocs.io/en/stable/web3.contract.html#web3.contract.ContractFunction
In other words, I'm pretty sure before this change you can do this same syntax on the underlying `Contract` object

dpaiton (approved):

lib/ethpy/ethpy/hyperdrive/state/conversions.py
sentilesdal: yeah camel_to_snake kinda stuff should be in a top level place.  actually maybe in like ethpy/utilities/conversions.py or something since that would be a common thing people would do when working with smart contracts.

lib/ethpy/ethpy/hyperdrive/state/pool_state.py
sentilesdal: this is fixed

PR #1124, opened 2023-11-27
check postgres status

wakamex:
instead of `time.sleep(3)` check the status of the postgres container by trying to connect to it.
- fix typehints, and lint docstrings.
- short test like lib/chainsync/chainsync/db/hyperdrive/interface_test.py is 30% faster: 6.73s to 4.69s (nice!)

### DIFF
--- a/lib/chainsync/chainsync/test_fixtures/db_session.py
+++ b/lib/chainsync/chainsync/test_fixtures/db_session.py
@@ -1,4 +1,4 @@
-"""Pytest fixture that creates an in memory db session and creates the base db schema"""
+"""Create an in memory db session and creates the base db schema."""
 import logging
 import os
 import re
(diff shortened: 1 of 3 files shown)


lib/chainsync/chainsync/test_fixtures/db_session.py
```diff
@@ -81,29 +83,33 @@ def psql_docker() -> Iterator[PostgresConfig]:
...
-
-    # Wait for the container to start
-    time.sleep(3)
+    assert isinstance(container, Container)
 
     # Get version of postgres
-    version_out = container.exec_run("postgres -V")[1]  # type:ignore
+    connected = False
+    version_out = ""
+    while not connected:
```
dpaiton, line 91: this seems dangerous -- I think we should set a `max_retries` that can be set via an environment variable. @slundqui thoughts?

slundqui: Agreed, this should check a certain amount of times before exiting, with a sleep between checks

dpaiton (approved):

lib/chainsync/chainsync/test_fixtures/db_session.py
wakamex: we have [for _ in range(10):](https://github.com/delvtech/agent0/blob/df58e5ecd5c1b15e6e474df910951e044aea3016/lib/chainsync/chainsync/db/base/interface.py#L99) in interface.py, so i'll just re-use that logic

PR #1127, opened 2023-11-29
Docs linting

slundqui:


wakamex:
numpy format has a space before the : <https://numpydoc.readthedocs.io/en/latest/format.html#parameters>

slundqui:
> numpy format has a space before the `:` for parameters https://numpydoc.readthedocs.io/en/latest/format.html#parameters

Sphinx doesn't seem like it cares, and it's nice for copy/paste from the function definition, in which black sets it to no space

PR #1146, opened 2023-12-01
Load/save snapshots using interactive hyperdrive

slundqui:
Adds functionality to load/save states with the interactive hyperdrive

### DIFF
--- a/lib/agent0/agent0/hyperdrive/interactive/chain.py
+++ b/lib/agent0/agent0/hyperdrive/interactive/chain.py
@@ -144,12 +144,12 @@ class Chain:
         assert "result" in response
         assert response["result"]
 
+        # load snapshot database state
+        self._load_db()
+
         # The hyperdrive interface in deployed pools need to wipe it's cache
         for pool in self._deployed_hyperdrive_pools:
             pool._reinit_state_after_load_snapshot()  # pylint: disable=protected-access
-
-        # load snapshot database state
-        self._load_db()
 
         self.save_snapshot()
 
--- a/lib/agent0/agent0/hyperdrive/interactive/chain.py
+++ b/lib/agent0/agent0/hyperdrive/interactive/chain.py
@@ -63,6 +63,10 @@ class Chain:
         )
         assert isinstance(self.postgres_container, Container)
 
+        # Snapshot bookkeeping
+        self._saved_snapshot_id: str
+        self._deployed_hyperdrive_pools = []
+
     def __del__(self):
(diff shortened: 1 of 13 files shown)


lib/agent0/agent0/hyperdrive/interactive/chain.py
```diff
@@ -99,6 +117,45 @@ def advance_time(self, time_delta: int | timedelta) -> RPCResponse:
 
         return self._web3.provider.make_request(method=RPCEndpoint("evm_mine"), params=[next_blocktime])
 
+    def save_snapshot(self) -> None:
+        """Saves a snapshot using the `evm_snapshot` RPC call.
+        The chain can store one snapshot at a time, saving another snapshot overwrites the previous snapshot.
+        Saving/loading snapshot only persist on the same chain, not across chains.
+        """
+        response = self._web3.provider.make_request(method=RPCEndpoint("evm_snapshot"), params=[])
+        assert "result" in response
```
dpaiton, line 126: I would prefer to check the conditional and raise an error as a rule of thumb instead of assert. 
```python
if "result" not in response:
    raise KeyError("Response did not have a result.")
```

This pattern encourages raising appropriate error types. Probably more importantly, assert statements are [removed when a wheel is built in optimized mode](https://realpython.com/python-assert-statement/#running-python-in-optimized-mode:~:text=The%20main%20results%20of%20running%20Python%20in%20the%20first%20level%20of%20optimization%20is%20that%20the%20interpreter%20sets%20__debug__%20to%20False%20and%20removes%20the%20assertions%20from%20the%20resulting%20compiled%20bytecode) which is something we may want to do soon if we're still having speed issues.

lib/agent0/agent0/hyperdrive/interactive/chain.py
```diff
@@ -99,6 +117,45 @@ def advance_time(self, time_delta: int | timedelta) -> RPCResponse:
...
+        # When reverting snapshots, the chain deletes the previous snapshot, while we want it to persist.
+        # Hence, in this function, we first revert the snapshot, then immediately create a new snapshot
+        # to keep the original snapshot.
+
+        if not self._has_saved_snapshot:
+            raise ValueError("No saved snapshot to load")
+
+        response = self._web3.provider.make_request(method=RPCEndpoint("evm_revert"), params=[self._saved_snapshot_id])
+        assert "result" in response
+        assert response["result"]
```
dpaiton, line 148: same as above; use if/raise pattern

dpaiton (approved):
Awesome!

PR #1149, opened 2023-12-02
make preview calls only when necessary for slippage

wakamex:


slundqui:
These calls are pretty crucial when doing fuzz testing, as we get error messages when previewing calls. We likely should add a `debug` parameter that switches between these cases if this is needed for speed.

lib/ethpy/ethpy/hyperdrive/api/_contract_calls.py
```diff
@@ -400,14 +400,14 @@ async def _async_add_liquidity(
...
     current_block = interface.current_pool_state.block_number
-    _ = smart_contract_preview_transaction(
-        interface.hyperdrive_contract,
-        agent_checksum_address,
-        "addLiquidity",
-        *fn_args,
-        block_number=current_block,
-        read_retry_count=interface.read_retry_count,
-    )
+    # _ = smart_contract_preview_transaction(
```
dpaiton, line 403: just delete this code if you don't want it

lib/ethpy/ethpy/hyperdrive/api/_contract_calls.py
```diff
@@ -450,14 +450,14 @@ async def _async_remove_liquidity(
...
     current_block = interface.current_pool_state.block_number
-    _ = smart_contract_preview_transaction(
-        interface.hyperdrive_contract,
-        agent_checksum_address,
-        "removeLiquidity",
-        *fn_args,
-        block_number=current_block,
-        read_retry_count=interface.read_retry_count,
-    )
+    # _ = smart_contract_preview_transaction(
```
dpaiton, line 453: same, delete dead comments

dpaiton:
> These calls are pretty crucial when doing fuzz testing, as we get error messages when previewing calls. We likely should add a `debug` parameter that switches between these cases if this is needed for speed.

it looks like he is checking if the bot has set the slippage tolerance, and only doing the check if that's the case. So I think for fuzz testing as long as we set slippage in the bot config we will do the preview calls?

dpaiton (changes requested):
looks good! small asks & waiting for convo to resolve before approving.

lib/ethpy/ethpy/hyperdrive/api/_contract_calls.py
dpaiton: Also, what's going on here? It looks like we did the preview call but then never used it? @sentilesdal do you have any insight as to how this preview_transaction command came to be & what it was intended for? I assume since we were just dumping the result into `_` Mihai is right to remove it entirely.

dpaiton:
After talking with @wakamex and @slundqui we have decided to add a "debug" flag for running these preview calls, which defaults to `False` but will need to be `True` for fuzz bots. 

To avoid confusion with the `DEBUG` environment variable (which automatically funds bots), lets call this flag `preview_before_trade`. This should be a member variable of the `agent0/hyperdrive/interactive/interactive_hyperdrive.py::InteractiveHyperdrive::Config` dataclass **AND** the `ethpy/eth_config.py::EthConfig` config dataclass. It should default to `False` in both cases.

When the InteractiveHyperdrive is instantiated, it should pass the Config `preview_before_trade` flag into eth_config, which is then used to create HyperdriveInterface:

```python
def __init__(self, chain: Chain, config: Config | None = None):
    """Constructor for the interactive hyperdrive agent.

    Arguments
    ---------
    chain: Chain
        The chain object to launch hyperdrive on
    config: Config | None
        The configuration for the initial pool configuration
    """
     ... # setup variables
    self.eth_config = EthConfig(artifacts_uri="not_used", rpc_uri=chain.rpc_uri, abi_dir=abi_dir, , preview_before_trade=config.preview_before_trade)
    # Deploys a hyperdrive factory + pool on the chain
    self._deployed_hyperdrive = self._deploy_hyperdrive(config, chain, self.eth_config.abi_dir)
    self.hyperdrive_interface = HyperdriveInterface(
        self.eth_config,
        self._deployed_hyperdrive.hyperdrive_contract_addresses,
    )
    ... # everything else
```

HyperdriveInterface can then forward this variable inside its async_* trades. For example:

```python
async def async_open_long(
    self,
    agent: LocalAccount,
    trade_amount: FixedPoint,
    slippage_tolerance: FixedPoint | None = None,
    nonce: Nonce | None = None,
) -> ReceiptBreakdown:
    """Contract call to open a long position.

    Arguments
    ---------
    agent: LocalAccount
        The account for the agent that is executing and signing the trade transaction.
    trade_amount: FixedPoint
        The size of the position, in base.
    slippage_tolerance: FixedPoint, optional
        Amount of slippage allowed from the trade.
        If given, then the trade will not execute unless the slippage is below this value.
        If not given, then execute the trade regardless of the slippage.
    nonce: Nonce, optional
        An optional explicit nonce to set with the transaction.

    Returns
    -------
    ReceiptBreakdown
        A dataclass containing the maturity time and the absolute values for token quantities changed.
    """
    return await _async_open_long(self, agent, trade_amount, slippage_tolerance, nonce, self.eth_config.preview_before_trade)
```

This will get you to the corresponding contract call functions, which will now check if this variable **OR** slippage is set before doing a preview call:

```python
if slippage_tolerance is not None or preview_before_trade:
    preview_result = smart_contract_preview_transaction(
        interface.hyperdrive_contract,
        agent_checksum_address,
        "openLong",
        *fn_args,
        block_number=current_block,
        read_retry_count=interface.read_retry_count,
    )
```

lib/ethpy/ethpy/eth_config.py
```diff
@@ -61,14 +64,17 @@ def build_eth_config(dotenv_file: str = "eth.env") -> EthConfig:
     rpc_uri = os.getenv("RPC_URI")
     database_api_uri = os.getenv("DATABASE_API_URI")
     abi_dir = os.getenv("ABI_DIR")
+    preview_before_trade = os.getenv("PREVIEW_BEFORE_TRADE")
```
dpaiton, line 67: can you add this to the `eth.env.sample` as well?

lib/ethpy/ethpy/hyperdrive/api/_contract_calls.py
```diff
@@ -377,14 +390,16 @@ async def _async_close_short(
...
+    if preview_before_trade or slippage_tolerance is not None:
+        preview_result = smart_contract_preview_transaction(
+            interface.hyperdrive_contract,
+            agent_checksum_address,
+            "closeShort",
+            *fn_args,
+            block_number=current_block,
+            read_retry_count=interface.read_retry_count,
+        )
     if slippage_tolerance:
```
dpaiton, line 403: should this be `slippage_tolerance is not None` to match the other patterns?

lib/ethpy/ethpy/hyperdrive/api/_contract_calls.py
```diff
@@ -543,6 +562,7 @@ async def _async_redeem_withdraw_shares(
     # before calling smart contract transact
     # Since current_pool_state.block_number is a property, we want to get the static block here
     current_block = interface.current_pool_state.block_number
+    # if preview_before_trade:
```
dpaiton, line 565: dead comment

lib/ethpy/ethpy/hyperdrive/api/_contract_calls.py
wakamex: ya I like it more explicit

dpaiton (approved):

requirements-dev.txt
```diff
@@ -20,4 +20,5 @@ tomli>=2.0.1
 urllib3<2.0
 # Used for tests of db interface
 docker
-pytest-postgresql
\ No newline at end of file
+pytest-postgresql
+rollbar
```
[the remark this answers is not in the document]
wakamex: moved in 4f4c3f3deaa165570faecf5144454c8d86404fe4

PR #1150, opened 2023-12-04
Adjusting path independence script to run

slundqui:


lib/agent0/interactive_fuzz/fuzz_path_independence.py
```diff
@@ -104,22 +118,14 @@
             agent.close_short(maturity_time=trade.maturity_time, bonds=trade.bond_amount)
 
     # Check the reserve amounts; they should be unchanged now that all of the trades are closed
+
+    pool_info_df = interactive_hyperdrive.get_pool_info(coerce_float=False)
+
+    pd.testing.assert_series_equal(
```
dpaiton, line 124: can you add a comment explaining what this is doing? It's checking the first row against the last, which I guess is the initial setting vs the final setting?

lib/agent0/interactive_fuzz/fuzz_path_independence.py
```diff
@@ -104,26 +114,39 @@
...
-        start_gov_fees_accrued == pool_state.gov_fees_accrued
-    ), f"{start_gov_fees_accrued=} != {pool_state.gov_fees_accrued}"
-    assert start_total_shares == pool_state.vault_shares, f"{start_total_shares=} != {pool_state.vault_shares}"
+
+    pool_info_df = interactive_hyperdrive.get_pool_info(coerce_float=False)
+
+    # On first run, save final state
+    if first_run:
+        check_data["check_pool_info_df"] = pool_info_df[check_columns].iloc[-1].copy()
+        # TODO add these to pool info
```
dpaiton, line 123: we don't add anything to pool info on purpose, the `pool_info` fields should exactly reflect what the contract returns. The `pool_state` container is used to hold any additional variables, like `vault_shares`

sentilesdal (approved):
lgtm

lib/agent0/interactive_fuzz/fuzz_path_independence.py
slundqui: Fixed in https://github.com/delvtech/agent0/pull/1153

PR #1155, opened 2023-12-05
Add Latest Block LRU Cache to web3.py

sentilesdal:
Found in the web3.py docs that there is a ready-to-go latest block cache that uses an LRU cache!


sentilesdal:
I'm going to close this for now.  I'm not confident this will be able to keep up with our fast tests

PR #1170, opened 2023-12-06
Fix up some interactive fuzz tests to unify logging & format

dpaiton:


lib/agent0/interactive_fuzz/fuzz_hyperdrive_balance.py
```diff
@@ -74,6 +88,12 @@
         raise AssertionError(f"{trade_type=} is not supported.")
     trade_events.append((agent, trade_event))
 
+# %%
+# Advance some time
+chain.advance_time(
+    rng.integers(low=0, high=interactive_hyperdrive.hyperdrive_interface.pool_config.position_duration - 1)
+)
```
slundqui, line 95: Probably want to make sure checkpoints are created here (pass in `create_checkpoints` to `advance_time`)

slundqui (approved):
LGTM

PR #1179, opened 2023-12-08
Fuzz bot add logs

slundqui:


lib/agent0/agent0/hyperdrive/crash_report/crash_report.py
```diff
@@ -49,11 +49,13 @@ def setup_hyperdrive_crash_report_logging(log_format_string: str | None = None)
 
 
 # pylint: disable=too-many-statements
+# pylint: disable=too-many-branches
```
sentilesdal, line 52: I propose putting the limits on these to double their current values instead of playing this game where we disable all the time

slundqui: I'm all for that

PR #1192, opened 2023-12-10
create hyperdrive interface fixture

wakamex:


lib/ethpy/ethpy/hyperdrive/interface/interface.py
```diff
@@ -140,6 +145,19 @@ def __init__(
         self.read_retry_count = read_retry_count
         self.write_retry_count = write_retry_count
 
+    @property
+    def deployed_hyperdrive_pool(self) -> DeployedHyperdrivePool:
```
sentilesdal, line 149: This will be more useful if we make the deployed pool an attribute of `self` and just return that if it exists, if not, checks if its deployed and creates the DeployedHyperdrivePool, or raises an error if it doesn't exist.

sentilesdal (approved):
Yeah this is nice

lib/ethpy/ethpy/hyperdrive/interface/interface.py
wakamex: done in 672078ae00c6d3edc0355f13504aff07ecd2fa84

PR #1194, opened 2023-12-10
add retry_utils test

wakamex:
test result:

```
diffs: 0.107, 0.108, 0.111, 0.111
Total wait time: 0.44 seconds.
Time to fail: 1.43 seconds.
```

### DIFF
--- a/lib/ethpy/ethpy/base/retry_utils.py
+++ b/lib/ethpy/ethpy/base/retry_utils.py
@@ -1,85 +1,13 @@
 """Wrapper functions for retrying."""
 from __future__ import annotations
 
-import asyncio
 import inspect
 import logging
 import time
-from typing import Awaitable, Callable, NamedTuple, ParamSpec, TypeVar
-
-from web3._utils.threads import Timeout
+from typing import Callable, ParamSpec, TypeVar
 
 P = ParamSpec("P")
 R = TypeVar("R")
-
-RetryCallOptions = NamedTuple(
-    "RetryCallOptions",
-    [
-        ("timeout", float),
-        ("start_latency", float),
-        ("backoff_multiplier", float),
-    ],
-)
-
-
-async def async_retry_call(
-    retry_count: int,
-    retry_exception_check: Callable[[Exception], bool] | None,
-    func: Callable[P, Awaitable[R]],
-    *args: P.args,
-    options: RetryCallOptions = RetryCallOptions(timeout=2, start_latency=0.01, backoff_multiplier=2),  # type: ignore
-    **kwargs: P.kwargs,
-) -> R:
-    """Retry an async function call.
-
-    Arguments
-    ---------
-    retry_count: int
-        The number of times to retry the function
-    retry_exception_check: Callable[[type[Exception]], bool] | None
-        A function that takes as an argument an exception and returns True if we want to retry on that exception
-        If None, will retry for all exceptions
-    func: Callable[P, Awaitable[R]]
-        The function to call.
-    *args: P.args
-        The positional arguments to call func with
-    options: RetryCallOptions
-        Parameters for the exponential backoff.
-    **kwargs: P.kwargs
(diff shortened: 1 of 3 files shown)


wakamex:
had to remove the ParamSpec cus I couldn't figure it how to avoid this pyright error otherwise:

```
  /code/elfpy/lib/ethpy/ethpy/base/retry_utils.py:21:5 - error: Keyword parameter "timeout" cannot appear in signature after ParamSpec args parameter (reportGeneralTypeIssues)
  /code/elfpy/lib/ethpy/ethpy/base/retry_utils.py:22:5 - error: Keyword parameter "start_latency" cannot appear in signature after ParamSpec args parameter (reportGeneralTypeIssues)
  /code/elfpy/lib/ethpy/ethpy/base/retry_utils.py:23:5 - error: Keyword parameter "backoff_multiplier" cannot appear in signature after ParamSpec args parameter (reportGeneralTypeIssues)
  /code/elfpy/lib/ethpy/ethpy/base/retry_utils.py:88:5 - error: Keyword parameter "timeout" cannot appear in signature after ParamSpec args parameter (reportGeneralTypeIssues)
  /code/elfpy/lib/ethpy/ethpy/base/retry_utils.py:89:5 - error: Keyword parameter "start_latency" cannot appear in signature after ParamSpec args parameter (reportGeneralTypeIssues)
  /code/elfpy/lib/ethpy/ethpy/base/retry_utils.py:90:5 - error: Keyword parameter "backoff_multiplier" cannot appear in signature after ParamSpec args parameter (reportGeneralTypeIssues)
```

lib/ethpy/ethpy/base/retry_utils.py
```diff
@@ -55,27 +70,29 @@ async def async_retry_call(
...
     raise exception
 
 
 def retry_call(
     retry_count: int,
     retry_exception_check: Callable[[Exception], bool] | None,
-    func: Callable[P, R],
-    *args: P.args,
-    **kwargs: P.kwargs,
+    func: Callable[..., R],
```
sentilesdal, line 88: I would take the time to organize these better.  func, *args and **kwargs should come last.  Put the options first with the other options.  Optionally we could make an options object to hold all the options:
```python
def retry_call(
    options: RetryCallOptions,
    func: Callable[P, R],
    *args: P.args,
     **kwargs: P.kwargs
```

lib/ethpy/ethpy/base/retry_utils.py
```diff
@@ -55,27 +70,29 @@ async def async_retry_call(
...
+    *args: Any,
```
sentilesdal, line 89: why `Any`, why not `P.args`?

sentilesdal (approved):
let's take a seconds to think about the args to retry_call and async_retry_call

sentilesdal (changes requested):
oops, meant to request changes, not approve

lib/ethpy/ethpy/base/retry_utils.py
sentilesdal: Oh I just saw your comment, lemme take a look at that.

sentilesdal: Oh, I think if we reorg the params it should be ok, try my suggestion above.

lib/ethpy/ethpy/base/retry_utils.py
wakamex: ah! adding `  # type: ignore` stops pyright from complaining about the latter!

wakamex: removed the part

lib/ethpy/ethpy/base/retry_utils.py
wakamex: removed that stuff, is just test now

sentilesdal (approved):

PR #1204, opened 2023-12-12
Minimum txn amount check and test

slundqui:


lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
```diff
@@ -151,8 +156,10 @@ async def async_execute_agent_trades(
             # these issues
             is_invalid_balance, trade_result = check_for_invalid_balance(trade_result)
             is_slippage, trade_result = check_for_slippage(trade_result)
+            is_min_txn_amount, trade_result = check_for_min_txn_amount(trade_result)
             trade_result.is_invalid_balance = is_invalid_balance
```
dpaiton, line 160: `trade_result` is being modified inside the `check_for_*` functions, but then flag (e.g. `is_min_txn_amount`) is being set outside the function. It seems to me like the function should either make modifications in-place or not, instead of the mixed version we have here.

tests/min_txn_amount_test.py
```diff
@@ -0,0 +1,488 @@
...
+        action_list = []
+        done_trading = False
+        if self.counter == 0:
+            # Add liquidity
+            action_list.append(
+                Trade(
+                    market_type=MarketType.HYPERDRIVE,
+                    market_action=HyperdriveMarketAction(
+                        action_type=HyperdriveActionType.ADD_LIQUIDITY,
+                        trade_amount=FixedPoint(10000),
```
dpaiton, line 107: should this be `SMALL_TRADE_AMOUNT`?

tests/min_txn_amount_test.py
```diff
@@ -0,0 +1,488 @@
...
+        action_list = []
+        done_trading = False
+        if self.counter == 0:
+            # Open Long
+            action_list.append(
+                Trade(
+                    market_type=MarketType.HYPERDRIVE,
+                    market_action=HyperdriveMarketAction(
+                        action_type=HyperdriveActionType.OPEN_LONG,
+                        trade_amount=FixedPoint(10000),
```
dpaiton, line 239: should this be `SMALL_TRADE_AMOUNT`?

tests/min_txn_amount_test.py
dpaiton: oh I see you're adding a bunch and then removing a small amount

dpaiton (approved):

dpaiton (approved):

PR #1207, opened 2023-12-12
improve calc_pnl functionality

wakamex:
- add `calc_pnl` flag to turn off PNL calculation during an experiment, useful only using terminal values in analysis
- value LP and Withdrawal Shares at `lp_share_price` without any smart contract calls


lib/chainsync/chainsync/analysis/calc_pnl.py
```diff
@@ -91,78 +94,40 @@ def calc_single_closeout(position: pd.Series, contract: Contract, pool_info: pd.
...
         # Should never get here
         raise ValueError(f"Unexpected token type: {tokentype}")
     return out_pnl
 
 
 def calc_closeout_pnl(
-    current_wallet: pd.DataFrame, pool_info: pd.DataFrame, hyperdrive_contract: Contract
+    current_wallet: pd.DataFrame,
+    hyperdrive_contract: Contract,
+    hyperdrive_interface: HyperdriveReadInterface,
```
dpaiton, line 109: I think it's better to still get `pool_info` here instead of `hyperdrive_interface`, and then grab the `lp_share_price` entry from that to pass to `calc_single_closeout`. The reason is that whenever we call `hyperdrive_interface.current_pool_state` it is checking the cache and possibly updating the state. We want the post-processing (basically anything in chainsync) to be able to lock _when_ it is doing whatever it is doing (like `calc_pnl`). So the better workflow is for the caller of this `calc_closeout_pnl` script to handle which pool state to use (whether it be current or at some block number) and then pass the pool info for that corresponding state.

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -50,7 +50,10 @@ def calc_shares_needed_for_bonds(
         _shares_to_gov: FixedPoint
             The associated shares going to governance.
     """
-    _shares_to_pool = interface.calc_shares_out_given_bonds_in_down(abs(bonds_needed), pool_state)
+    try:
```
dpaiton, line 53: why are you adding this? or, what are you trying to sovle here? The difference is rounding, yes? This seems like a bad idea to me...

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -104,14 +107,25 @@ def calc_reserves_to_hit_target_rate(
...
+        divisor = FixedPoint(2)
+        bonds_needed = FixedPoint(0)
+        avoid_negative_share_reserves = False
+        while avoid_negative_share_reserves is False:
+            bonds_needed = (target_bonds - pool_state.pool_info.bond_reserves) / divisor
+            try:
+                shares_to_pool, shares_to_gov = calc_shares_needed_for_bonds(bonds_needed, pool_state, interface)
+                # save bad first guess to a temporary variable
+                temp_pool_state = apply_step(deepcopy(pool_state), bonds_needed, shares_to_pool, shares_to_gov)
+                predicted_rate = interface.calc_fixed_rate(temp_pool_state)
```
dpaiton, line 120: this predicted_rate can be outside the try/except & while blocks right? it has no impact on the loop computation

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -104,14 +107,25 @@ def calc_reserves_to_hit_target_rate(
...
+                avoid_negative_share_reserves = temp_pool_state.pool_info.share_reserves >= 0
+            except:  # pylint: disable=bare-except
```
dpaiton, line 122: if you're going to broad except (which you shouldn't) then please add a comment indicating what sort of issue you're expecting to have to avoid. And honestly it'd be best if you figured out what type of error that issue causes and catch that explicitly.

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -130,6 +130,8 @@ class Config:
             The upper bound on the flat fee that governance can set.
         max_governance_fee: FixedPoint
             The upper bound on the governance fee that governance can set.
+        calc_pnl: bool
```
dpaiton, line 133: I might be misremembering from before the holiday break, but I thought @slundqui and I said we did not want to add a `calc_pnl` flag here. iirc, the reasoning was that we don't want to establish a pattern that leads to bloat in this config (adding a flag for every modular change you want in the interface).

You are adding this flag because it is too slow when calculating PNL at every trade, yes? If so, I think the better solution is to remove the PNL calculation entirely as a default action, and then explicitly calculate a PNL column when it is needed and adding that column to the existing database (which I assume only happens when building the dashboard?)

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
wakamex: I'm going to test whether I can get away with not using it

lib/chainsync/chainsync/analysis/calc_pnl.py
```diff
@@ -8,11 +8,14 @@
 from eth_typing import ChecksumAddress, HexAddress, HexStr
 from ethpy.base import smart_contract_preview_transaction
 from ethpy.hyperdrive import BASE_TOKEN_SYMBOL
+from ethpy.hyperdrive.interface import HyperdriveReadInterface
 from fixedpointmath import FixedPoint
 from web3.contract.contract import Contract
 
 
-def calc_single_closeout(position: pd.Series, contract: Contract, pool_info: pd.DataFrame, min_output: int) -> Decimal:
+def calc_single_closeout(
```
dpaiton, line 16: FYI if speed is a concern, you should consider changing all of the `assert` statements in this function

1) don't use asserts -- they're slow; it's better to do `if condition; raise error`. Also, if we compile with optimizations on then [they don't get executed at all](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement:~:text=The%20current%20code%20generator%20emits%20no%20code%20for%20an%20assert%20statement%20when%20optimization%20is%20requested%20at%20compile%20time), so if they're important asserts we should be doing the if/raise pattern anyway.

2) Even after changing them, they should be more concise & only when necessary (for example, that `assert isinstance(maturity, int)` line is unnecessary).

lib/chainsync/chainsync/analysis/calc_pnl.py
```diff
@@ -91,78 +94,40 @@ def calc_single_closeout(position: pd.Series, contract: Contract, pool_info: pd.
...
-            ),
-        )
-        try:
-            # For PNL, we assume all withdrawal shares are redeemable
-            # even if there are no withdrawal shares available to withdraw
-            # Hence, we don't use preview transaction here
-            out_pnl = Decimal(amount * pool_info["lp_share_price"].values[-1]) / Decimal(1e18)
-        except Exception as exception:  # pylint: disable=broad-except
-            logging.warning("Exception caught, ignoring: %s", exception)
+    elif tokentype in ["LP", "WITHDRAWAL_SHARE"]:
```
dpaiton, line 97: probably would be helpful to preserve this comment; I was confused for a sec why you didn't have to preview:

```python
# For PNL, we assume all withdrawal shares are redeemable
# even if there are no withdrawal shares available to withdraw
# Hence, we don't use preview transaction here
```

dpaiton (changes requested):
Let's discuss this at standup tomorrow, but I think before merging you should modify the data analysis pipeline to _not_ do calc pnl at all. It would be better to have an api call to explicitly calc pnl (something like `interface.add_pnl_to_db`) and then the dashboard (and whoever else uses PNL) calc do this explicitly.

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
dpaiton: We decided that this flag isn't ideal, but the alternative solution is too difficult to handle right now. So we're going to leave the flag in and @wakamex will make an issue to remind us of the desired pattern.

dpaiton:
We decided that this flag isn't ideal, but the alternative solution is too difficult to handle right now. So we're going to leave the flag in and @wakamex will make an issue to remind us of the desired pattern.

lib/chainsync/chainsync/analysis/calc_pnl.py
wakamex: nice, and pool_info is already being called in data_to_analysis.py right before this, so it's free

lib/chainsync/chainsync/analysis/calc_pnl.py
wakamex: I remember doing this for typechecking, but I just removed them, and didn't get any errors

wakamex:
created https://github.com/delvtech/agent0/issues/1237

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
wakamex: ran a sweep without this try/catch, so assuming it works for now. will add back later with more detail, if needed.

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
wakamex: took that out, may be obsolete since adding `avoid_negative_share_reserves`

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
wakamex: removed

slundqui (approved):

wakamex:
sanity check to ensure new valuation method is equivalent to previous:

```python
    if tokentype == "LP":
        fn_args = (
            amount,
            min_output,
            (  # IHyperdrive.Options
                address,  # destination
                True,  # asBase
                bytes(0),  # extraData
            ),
        )
        preview_result = smart_contract_preview_transaction(
            contract, sender, "removeLiquidity", *fn_args, block_number=position["block_number"]
        )
        base_proceeds = preview_result["baseProceeds"]
        withdrawal_shares = preview_result["withdrawalShares"]
        alternate_pnl = Decimal(base_proceeds + withdrawal_shares * lp_share_price) / Decimal(1e18)
        assert out_pnl == alternate_pnl, f"Expected {out_pnl} but got {alternate_pnl}, diff={out_pnl - alternate_pnl}, diff(%)={100*(out_pnl - alternate_pnl)/out_pnl}"
```

passes after removing an unnecessary cast to float

PR #1218, opened 2023-12-15
Use pypechain's new deploy method

sentilesdal:
Use this instead of `deploy_contract`, which is now no longer used.  We can delete it or leave it since its still a utility that ethpy could provide.

I've also totally removed the dependency of `abi_dir` in eth_config throughout the repo... I think.  I'm not taking the final step of removing that either in this PR in case there is some use case I don't know about.

### DIFF
--- a/lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
+++ b/lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
@@ -34,7 +34,7 @@ from eth_utils.address import to_checksum_address
 from ethpy import EthConfig
 from ethpy.base import set_anvil_account_balance, smart_contract_transact
 from ethpy.hyperdrive import BASE_TOKEN_SYMBOL, DeployedHyperdrivePool, ReceiptBreakdown, deploy_hyperdrive_from_factory
-from ethpy.hyperdrive.interface import HyperdriveInterface
+from ethpy.hyperdrive.interface import HyperdriveReadWriteInterface
 from fixedpointmath import FixedPoint
 from hypertypes import Fees, PoolConfig
 from numpy.random._generator import Generator
@@ -201,7 +201,7 @@ class InteractiveHyperdrive:
         )
         # Deploys a hyperdrive factory + pool on the chain
         self._deployed_hyperdrive = self._deploy_hyperdrive(config, chain)
-        self.hyperdrive_interface = HyperdriveInterface(
+        self.hyperdrive_interface = HyperdriveReadWriteInterface(
             self.eth_config,
             self._deployed_hyperdrive.hyperdrive_contract_addresses,
             web3=chain._web3,
@@ -520,7 +520,7 @@ class InteractiveHyperdrive:
         Returns
         -------
         InteractiveHyperdriveAgent
-            An object that contains the HyperdriveInterface, Agents,
+            An object that contains the HyperdriveReadWriteInterface, Agents,
             and provides access to the interactive Hyperdrive API.
         """
         # pylint: disable=too-many-arguments
--- a/lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
+++ b/lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
@@ -34,7 +34,7 @@ from eth_utils.address import to_checksum_address
 from ethpy import EthConfig
 from ethpy.base import set_anvil_account_balance, smart_contract_transact
 from ethpy.hyperdrive import BASE_TOKEN_SYMBOL, DeployedHyperdrivePool, ReceiptBreakdown, deploy_hyperdrive_from_factory
-from ethpy.hyperdrive.interface import HyperdriveReadWriteInterface
+from ethpy.hyperdrive.interface import HyperdriveInterface
 from fixedpointmath import FixedPoint
 from hypertypes import Fees, PoolConfig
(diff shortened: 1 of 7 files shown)


dpaiton:
I'm fine with removing `abi_dir` completely. I agree that `deploy_contract` could have some use as an ethpy util, but only if we have good tests around it so it doesn't rot.

lib/ethpy/ethpy/hyperdrive/deploy.py
```diff
@@ -229,79 +226,62 @@ def _deploy_hyperdrive_factory(
...
     """
-    # args = [name, symbol, decimals, admin_addr, isCompetitionMode]
-    args = ["Base", "BASE", 18, ADDRESS_ZERO, False]
-    base_token_contract_addr, base_token_contract = deploy_contract(
-        web3,
-        abi=abis["ERC20Mintable"],
-        bytecode=bytecodes["ERC20Mintable"],
-        deploy_account_addr=deploy_account_addr,
-        args=args,
+    erc20args = ERC20MintableContract.ConstructorArgs("Base", "BASE", 18, ADDRESS_ZERO, False)
```
dpaiton, line 229: I like the pattern of commenting the solidity arg name next to each of these. It makes it easy to cross-reference against the sol file when debugging. I was doing just that yesterday. So, for example, with this line:

```python
erc20args = ERC20MintableContract.ConstructorArgs(
    "Base", # name
    "BASE", # symbol
    18, # decimals
    ADDRESS_ZERO, # admin_addr
    False # isCompetitionMode
)
```

lib/ethpy/ethpy/hyperdrive/deploy.py
```diff
@@ -229,79 +226,62 @@ def _deploy_hyperdrive_factory(
...
+    base_token_contract = ERC20MintableContract.deploy(w3=web3, account=deploy_account_addr, constructorArgs=erc20args)
+
+    pool = MockERC4626Contract.deploy(
+        w3=web3,
+        account=deploy_account_addr,
+        constructorArgs=MockERC4626Contract.ConstructorArgs(
```
dpaiton, line 235: And this would be:

```python
constructorArgs=MockERC4626Contract.ConstructorArgs(
    base_token_contract.address, # asset
    "Delvnet Yield Source", # name
    "DELV", # symbol
    initial_variable_rate.scaled_value, # initialRate
    ADDRESS_ZERO, # admin 
    False, # isCompetitionMode
),
 ```

File: lib/ethpy/ethpy/hyperdrive/deploy.py [hunk 3]
```diff
@@ -198,8 +201,6 @@ def _initialize_deployment_account(web3: Web3, account_private_key: str) -> Loca
 def _deploy_hyperdrive_factory(
     web3: Web3,
     deploy_account_addr: ChecksumAddress,
-    abis: dict,
-    bytecodes: dict,
     initial_variable_rate: FixedPoint,
     pool_config: PoolConfig,
     max_fees: Fees,
```
dpaiton, line 206: The return types on this func are different now, right?

dpaiton:
some more instances of ABI:

- ABI_DIR is an env variable in `eth.env.sample`

dpaiton (changes requested):
might as well purge the rest of abi_dir. Overall, though, this is great! I think the code is much cleaner & more intuitive now.

lib/ethpy/ethpy/hyperdrive/deploy.py
sentilesdal: but now you can just hover the ConstructorArgs to see what they are:

image

but I'll do a solid here and just declare like this:
image

lib/ethpy/ethpy/hyperdrive/deploy.py
sentilesdal: using kwarg style instead.

lib/ethpy/ethpy/hyperdrive/deploy.py
sentilesdal: hah you beat me to it.  i just changed the return types to be more specific

lib/ethpy/ethpy/hyperdrive/deploy.py
dpaiton: oh wow that's awesome! Ok cool yeah I like using the kwargs but the argument names on hover is 100x better than commented names 🤩

lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -184,7 +184,6 @@ def launch_local_hyperdrive_pool(
     )
     return deploy_hyperdrive_from_factory(
         local_chain_uri,
-        abi_folder,
```
dpaiton, line 187: reminder to delete this variable from above (lines 145-147)

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -505,7 +504,7 @@ def init_agent(
         Returns
         -------
         InteractiveHyperdriveAgent
-            An object that contains the HyperdriveReadWriteInterface, Agents,
+            An object that contains the HyperdriveInterface, Agents,
```
dpaiton, line 507: This change should be reverted (back to `HyperdriveReadWriteInterface`)

dpaiton (approved):

PR #1220, opened 2023-12-15
split hyperdrive interface into read & readwrite versions

dpaiton:
The Hyperdrive Interface object has read operations (like getting block information, or pool info) and write operations (like opening a long or short). We want to be able to restrict the type to read-only or read-write versions of the interface to remove any ambiguity about who should and should not be executing trades. Specifically, the bot policies all share an `action` function that is expected to return a list of Trades. The action function is not supposed to execute trades itself, since that is executed elsewhere to account for database updates, error handling, etc. Now all policies specifically receive the ReadInterface to remove any ambiguity about the allowed action set.

I decided to use an inheritance model instead of class attributes, which is how I originally proposed it in the [issue](https://github.com/delvtech/agent0/issues/1128). I decided on this because it removes a step when importing & declaring types (`HyperdriveInterface.ReadInterface` versus only `HyperdriveReadInterface`).


wakamex:
what's the intended usage here? I can't think when I'd prefer ReadInterface, if ReadWriteInterface is a super-set.

dpaiton:
> what's the intended usage here? I can't think when I'd prefer ReadInterface, if ReadWriteInterface is a super-set.

The intent is to be restrict the expected type. For example, bot policies should not be able to execute trades in the `action` function, which is why they now receive a `ReadInterface`. Before, when we had a single interface object, the bots could technically make trades in their action policy, which would have caused trouble down-stream for the database and wallet systems.

lib/agent0/agent0/hyperdrive/exec/run_agents.py
```diff
@@ -158,7 +158,7 @@ def setup_agents(
         contract_addresses = fetch_hyperdrive_address_from_uri(os.path.join(eth_config.artifacts_uri, "addresses.json"))
 
     # create hyperdrive interface object
-    interface = HyperdriveInterface(
+    interface = HyperdriveReadWriteInterface(
```
slundqui, line 161: I wonder if we should just keep the name here to be `HyperdriveInterface` to avoid verboseness

slundqui (approved):

PR #1221, opened 2023-12-18
Adds interface endpoints for building Trade objects

dpaiton:


lib/ethpy/ethpy/hyperdrive/interface/read_interface.py
```diff
@@ -389,6 +393,215 @@ def get_gov_fees_accrued(self, block_number: BlockNumber | None = None) -> Fixed
         """
         return _get_gov_fees_accrued(self.hyperdrive_contract, block_number)
 
+    @staticmethod
+    def open_long_trade(
```
sentilesdal, line 397: shouldn't these be in the read/write interface?

dpaiton: It's in the read interface because it is only returning a Trade object, as opposed to actually executing the trade. I was unsure about the best names for these. Right now if we want to execute a trade we would use ReadWriteInterface.async_open_long and if we want to return a trade object (to be executed by other agent0 machinery) then we would use ReadInterface.open_long_trade. The later is the parimary form of communication from policies to the agent0 trade exec pipeline.

sentilesdal (approved):
ship it!

PR #1222, opened 2023-12-18
Randomize liquidation

slundqui:


dpaiton (approved):

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -515,6 +531,11 @@ def init_agent(
             base = FixedPoint(0)
         if eth is None:
             eth = FixedPoint(10)
+
+        # If the underlying policy's rng isn't set, we use the one from interactive hyperdrive
+        if policy_config is not None and policy_config.rng is None and policy_config.rng_seed is None:
```
wakamex, line 536: judging by codecov saying the following line isn't tested, this condition is never true. I think because the rng object is created in the agent and interactive hyperdrive initialization, it's never None.

wakamex: be the change you want to see in the world, they said https://github.com/delvtech/agent0/pull/1223

wakamex: fuck yaaaaaaaaaa
image

PR #1223, opened 2023-12-19
additional interactive hyperdrive tests

wakamex:


lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_test.py
```diff
@@ -506,3 +528,32 @@ def test_random_liquidate(chain: LocalChain):
...
+    interactive_hyperdrive = InteractiveHyperdrive(chain, interactive_config)
+    agent_policy = Zoo.random.Config()
+    agent_policy.rng = None
+    alice = interactive_hyperdrive.init_agent(
+        base=FixedPoint(10_000),
+        name="alice",
+        policy=Zoo.random,
+        policy_config=agent_policy,
+    )
+    assert alice.agent.policy.rng is not None
```
dpaiton, line 546: I guess in this case we use an internally set random seed. can you assert this value? it's probably also an attribute or the policy.

dpaiton (approved):
small request for an additional assert, otherwise always appreciate more tests!

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_test.py
wakamex: looks like we use a random seed, if one isn't provided:

https://github.com/delvtech/agent0/blob/69ba04d9f67c8674600e430be7838f72a806aaf4/lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py#L139

https://github.com/delvtech/agent0/blob/69ba04d9f67c8674600e430be7838f72a806aaf4/lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py#L162-L163

from `default_rng`:

```
Parameters
seed : {None, int, array_like[ints], SeedSequence, BitGenerator, Generator}, optional
    A seed to initialize the BitGenerator. If None, then fresh, unpredictable entropy will be pulled from the OS.
```

wakamex: agreed to leave default behavior as "random by default", requiring user to explicitly set a seed for deterministic behavior

PR #1225, opened 2024-01-04
pin fixedpointmath version to 0.1.6

wakamex:


wakamex:
doesn't actually help us avoid old version issues. updating to latest should happen even without a version pin, as long as we remember to do `pip install --upgrade -r requirements.txt`.

PR #1240, opened 2024-01-10
improve LPandArb bot

wakamex:
- avoid crash on negative share reserves
- use self.minimum_trade_amount
- add tests for hitting target when closing positions
- improve logic to hit target when closing positions


lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_test.py
```diff
@@ -17,6 +17,8 @@
 # pylint: disable=redefined-outer-name
 # ruff: noqa: PLR2004 (comparison against magic values (literals like numbers))
 
+YEAR_IN_SECONDS = 31_536_000
```
dpaiton, line 20: why add this line? It is not used anywhere, right?

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -92,7 +94,7 @@ def calc_reserves_to_hit_target_rate(
     total_shares_needed = FixedPoint(0)
     total_bonds_needed = FixedPoint(0)
     # pylint: disable=logging-fstring-interpolation
-    logging.debug(f"Targeting {float(target_rate):.2%} from {float(interface.calc_fixed_rate()):.2%}")
+    logging.warning(f"Targeting {float(target_rate):.2%} from {float(interface.calc_fixed_rate()):.2%}")
```
dpaiton, line 97: I don't think this is a warning, as there is nothing wrong when it is posted. If it's not helpful in your debug then maybe we should just delete it all together? Lots of non-helpful logs can be worse than few logs.

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -126,12 +141,12 @@ def calc_reserves_to_hit_target_rate(
             f"iteration {iteration:3}: {float(predicted_rate):22.18%}"
             + f" d_bonds={float(total_bonds_needed):27,.18f} d_shares={float(total_shares_needed):27,.18f}"
         )
-        logging.debug(formatted_str)
+        logging.warning(formatted_str)
```
dpaiton, line 144: same as above; this is not a warning

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
```diff
@@ -126,12 +141,12 @@ def calc_reserves_to_hit_target_rate(
...
         if iteration >= MAX_ITER:
             break
     convergence_speed = time.time() - start_time
     formatted_str = f"predicted precision: {float(abs(predicted_rate-target_rate))}, time taken: {convergence_speed}s"
-    logging.debug(formatted_str)
+    logging.warning(formatted_str)
```
dpaiton, line 149: same as above; this is not a warning

lib/agent0/agent0/hyperdrive/policies/lpandarb_test.py
```diff
@@ -0,0 +1,253 @@
...
+    # report starting fixed rate
+    logging.warning("starting fixed rate is %s", interactive_hyperdrive.hyperdrive_interface.calc_fixed_rate())
+
+    # arbitrage it back (the only trade capable of this is a long)
+    arbitrage_andy.execute_policy_action()
+
+    # report results
+    fixed_rate = interactive_hyperdrive.hyperdrive_interface.calc_fixed_rate()
+    variable_rate = interactive_hyperdrive.hyperdrive_interface.current_pool_state.variable_rate
+    logging.warning("ending fixed rate is %s", fixed_rate)
```
dpaiton, line 104: why are these warnings? they should be info or debug

dpaiton (changes requested):
Not needed in this PR, but lpandarb.py is getting complicated enough that I'm having a hard time following. I think it would be valuable to follow up with a cleanup that includes more comments, chunking repeated or complex code into unit functions, etc.

Other than that, pls adjust all of the logging to be a correct level. Warning should be reserved for situations where the code is behaving in an unexpected way, user input must be ignored or adjusted, deprecated code was used, a default is taken due to a workaround, etc. For example, "transaction failed, skipping" or "trade amount too low, adjusting to the minimum".

We have ourselves taken time to locate the source of and eliminate warnings coming from other libs, it's something developers pay attention to and attempt to resolve. I think all of your new warning logs should probably be debug, but I'll leave it up to you if you want to make some info.

lib/agent0/agent0/hyperdrive/policies/lpandarb.py
wakamex: I changed this because running pytest with log level of debug returns like 10,000 lines

PR #1245, opened 2024-01-10
interactive fuzz trades use correct max amount

dpaiton:


lib/agent0/agent0/interactive_fuzz/helpers/execute_random_trades.py
```diff
@@ -0,0 +1,160 @@
...
+        trade_amount = _get_open_trade_amount(trade_type, rng, interactive_hyperdrive)
+        # extra base accounts for fees
+        # the short trade amount is technically bonds, but we know that will be less than the required base
+        agent = interactive_hyperdrive.init_agent(base=trade_amount * 2, eth=FixedPoint(100))
+        trade_event = _execute_trade(trade_type, trade_amount, agent)
+        trade_events.append((agent, trade_event))
+        if advance_time:
+            # Advance a random amount of time between opening trades
+            chain.advance_time(
+                time_diffs[trade_index],  # type: ignore
```
slundqui, line 72: This type ignore is complaining that time_diffs is possibly unbound, because the variable is in an if statement above. Define the variable and run an assertion to avoid this type:ignore

slundqui (approved):

PR #1249, opened 2024-01-11
Various updates to fuzz testing

slundqui:
- Set open trades to be 75% of calc max
- Path independence check effective share reserves has a tolerance of 1e-4
- Path independence sets variable interest to 0 for closing trades only
- Remove hyperdrive balance test


dpaiton (approved):
lgtm

Issue #1123, opened 2023-11-27
fuzzing: add invariant tests in fuzz pipeline


dpaiton:
These issues
https://github.com/delvtech/hyperdrive/issues/253
https://github.com/delvtech/hyperdrive/issues/534

outline a list of "invariants" (assumed truths) that should hold in hyperdrive at any time.

We've sorted them into two categories:
1. Alongside fuzz testing -- After each block is mined these invariants will be checked using the stored database state values inside a parallel cron job
2. Using interactive hyperdrive -- These are not using the fuzz framework, but instead are targeted tests using a controlled set of trades

---------------------------

1.  Invariants to check alongside fuzz trades:
- [x] total shares == share reserves + shorts outstanding/c + _governanceFeesAccrued + _withdrawalPool.proceeds + epsilon
- [x] the system should always be solvent (solvency = share_reserves - global_exposure - minimum_share_reserves > 0)
- [x] PV always >= idle
- [x] ∆ lp share price should never be more than 0.1%
- [x] share reserves * c - long exposure >= minimum share reserves
- [x] creating a checkpoint should never fail

2. Conditions to check in an interactive setting (aka static tests; not included in fuzz testing)
- [x] opening and immediately closing a trade shouldn't result in profit
- [x] longs and shorts should always be closed at maturity for the correct amounts
- [x] given a set of open positions, the final state of the reserves will be the same no matter what order the positions are closed in (path independence).
- [x] if all of the funds are removed from Hyperdrive, there is no base left in the contract

Present value interactive test
do all trades (add & remove liquidity, open & close longs & shorts)
- [x] For any trade, LP share price (which = present value / number of shares) shouldn't change by more than 0.1% when distributing excess idle (which is called when any action is done)
- [x] open or close trades shouldn't affect PV (within tolerance, maybe 0.1%?)
- [x] removing liquidity shouldn't result in the PV increasing (it should decrease)
- [x] adding liquidity shouldn't result in the PV decreasing (it should increase)
- [x] PV always >= idle

dpaiton:
Input from @jalextowle 

[...]

PR #1251, opened 2024-01-16
Add invariant checks to fuzz tests

sentilesdal:
Adds the following checks:

 1. PV always >= idle
 2. ∆ lp share price should never be more than 0.1%

Part of #1123

### DIFF
--- a/lib/agent0/bin/fuzz_bot_invariant_checks.py
+++ b/lib/agent0/bin/fuzz_bot_invariant_checks.py
@@ -130,99 +130,97 @@ def run_invariant_checks(
     test_epsilon: float
         The tolerance for the invariance checks.
     """
-    # pylint: disable=too-many-locals
     # Get the variables to check & check each invariant
     pool_state = interface.get_hyperdrive_state(latest_block)
-    failed = False
+    any_check_failed = False
     exception_message: list[str] = ["Continuous Fuzz Bots Invariant Checks"]
     exception_data: dict[str, Any] = {}
 
+    eth_balances_results = _check_eth_balances(latest_block_number, pool_state)
+    base_balances_results = _check_base_balances(latest_block_number, pool_state)
+    total_share_results = _check_total_shares(latest_block_number, test_epsilon, pool_state)
+    minimum_share_reserves_results = _check_minimum_share_reserves(latest_block_number, pool_state)
+    solvency_results = _check_solvency(latest_block_number, pool_state)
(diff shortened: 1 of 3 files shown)


lib/agent0/bin/fuzz_bot_invariant_checks.py
```diff
@@ -231,6 +247,84 @@ def run_invariant_checks(
         raise FuzzAssertionException(*exception_message, exception_data=exception_data)
 
 
+def _lp_share_price(
+    block_number: BlockNumber,
+    interface: HyperdriveReadInterface,
+    test_epsilon: float,
+    pool_state: PoolState,
+    exception_message: list[str],
+    exception_data: dict[str, Any],
```
slundqui, line 256: Are you sure these are passed in via reference? Can we return the exception message and exception data just to make it explicit that these two fields are getting adjusted? Better yet, don't pass these in and add the message/data in the caller

lib/ethpy/ethpy/hyperdrive/interface/read_interface.py
```diff
@@ -320,6 +320,29 @@ def get_vault_shares(self, block_number: BlockNumber | None) -> FixedPoint:
             block_number = self.get_block_number(self.get_current_block())
         return _get_vault_shares(self.yield_contract, self.hyperdrive_contract, block_number)
 
+    def get_idle_shares(self, block_number: BlockNumber | None) -> FixedPoint:
```
slundqui, line 323: You'll want to double check with @dpaiton here, but I think the read interface might be strictly contract calls without math on our end

dpaiton: We don't have a rust function for this yet, so I think it's ok to do it in python. However it's not using an RPC to get the balance so I think the docstring is incorrect.

sentilesdal: Added an issue to hyperdrive-sdk: https://github.com/delvtech/hyperdrive-sdk/issues/55

lib/agent0/bin/fuzz_bot_invariant_checks.py
sentilesdal: yeah lists and dicts are always passed by reference but i like what you're saying about making it more explicit anyway.

dpaiton (approved):
awesome!

PR #1252, opened 2024-01-16
Better rollbar logging and path invariance test upgrades

slundqui:
1. Interactive hyperdrive has parameter that allows logs to rollbar
2. Adding log level when crash reporting in interactive hyperdrive
3. Path independence allows for failed paths, logging failures as info to rollbar
4. Adds optional ticker to interactive hyperdrive crash reports
5. Adds optional message prefix to rollbar logs
6. Catching and printing unexpected errors in run all fuzz tests


lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -143,6 +145,7 @@ class Config:
         data_pipeline_timeout: int = 60
         preview_before_trade: bool = False
         log_to_rollbar: bool = False
+        crash_log_level: int | None = None
```
sentilesdal, line 153: if defaults to critical, shouldn't that be reflected here instead of set to None?

sentilesdal (approved):
lgtm

PR #1266, opened 2024-01-18
Setting position duration in interactive fuzz

slundqui:
Due to this bug: https://github.com/delvtech/hyperdrive/issues/732


dpaiton (approved):
don't forget to revert this after HD 0.6.0 drops!

PR #1270, opened 2024-01-19
Hotfixes for fuzz bots

slundqui:
- Develop mode now ignores any existing account.env files and always explicitly uses accounts and budgets defined in runner scripts.
- Fixed a bug where agent wallet bookkeeping wasn't getting updated when we refund fuzz bots

### DIFF
--- a/lib/agent0/agent0/accounts_config.py
+++ b/lib/agent0/agent0/accounts_config.py
@@ -89,25 +89,31 @@ def initialize_accounts(
     if env_file is None:
         env_file = "account.env"
 
-    if not os.path.exists(env_file):
+    # If we're in develop mode, we don't use the env file at all
+    if develop:
+        account_key_config = build_account_key_config_from_agent_config(agent_config, random_seed)
+    # If we're not in develop mode and the env file doesn't exist
+    # we create the env file keeping track of keys and budgets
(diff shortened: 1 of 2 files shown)


lib/agent0/agent0/accounts_config.py
```diff
@@ -89,25 +89,31 @@ def initialize_accounts(
...
-                "Run the following command to fund the accounts, then rerun this script."
-            )
-            # Different commands depending on if default env file is used
-            command_str = "python lib/agent0/bin/fund_agents_from_user_key.py -u <user_private_key>"
-            if env_file != "account.env":
-                command_str += f" -f {env_file}"
-            print(command_str)
-            # Clean exit
-            sys.exit(0)
+        print(
```
dpaiton, line 104: can you make this a logging.info?

dpaiton (approved):
thanks for looking so closely at this stuff!

lib/agent0/agent0/accounts_config.py
slundqui: This print is explicitly a helper command when running bots in e.g., trading competition, which details the next steps needed for funding bots. Hence, I'd like this to always print out regardless of parameters for logging.

PR #1273, opened 2024-01-19
update doc strings and error messaging

dpaiton:


lib/agent0/agent0/interactive_fuzz/fuzz_path_independence.py
```diff
@@ -399,6 +406,7 @@ def invariant_check(
         if not fp_isclose(expected_val, actual_val, abs_tol=epsilon):
             difference_in_wei = abs(expected_val.scaled_value - actual_val.scaled_value)
             exception_message.append(
+                "The pool state has deviated after closing all of the trades.\n"
```
slundqui, line 409: New lines here don't really show up in rollbar, I would rather these be separate arguments, specifically appending two arguments to the exception, one for the message and the other for the amounts.

slundqui (approved):

lib/agent0/agent0/interactive_fuzz/fuzz_path_independence.py
dpaiton: got it!

PR #1275, opened 2024-01-20
fix liquidity present value tests

dpaiton:
initial present value is recalculated for add & remove liquidity

### DIFF
--- a/lib/agent0/agent0/interactive_fuzz/fuzz_present_value.py
+++ b/lib/agent0/agent0/interactive_fuzz/fuzz_present_value.py
@@ -82,11 +82,14 @@ def fuzz_present_value(
     }
     agent = interactive_hyperdrive.init_agent(base=FixedPoint("1e10"), eth=FixedPoint(1_000))
 
+    # Execute the trades and check invariances for each trade
     for trade_type in [
         HyperdriveActionType.OPEN_LONG,
         HyperdriveActionType.CLOSE_LONG,
         HyperdriveActionType.OPEN_SHORT,
         HyperdriveActionType.CLOSE_SHORT,
+        HyperdriveActionType.ADD_LIQUIDITY,
+        HyperdriveActionType.REMOVE_LIQUIDITY,
     ]:
         # Keep the agent flush
         if agent.wallet.balance.amount < FixedPoint("1e10"):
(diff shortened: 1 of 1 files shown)


lib/agent0/agent0/interactive_fuzz/fuzz_present_value.py
```diff
@@ -118,14 +121,37 @@ def fuzz_present_value(
...
+                )
                 trade_amount = FixedPoint(
                     scaled_value=int(
                         np.floor(rng.uniform(low=min_trade.scaled_value, high=agent.wallet.balance.amount.scaled_value))
                     )
                 )
                 trade_event = agent.add_liquidity(trade_amount)
             case HyperdriveActionType.REMOVE_LIQUIDITY:
+                # recompute initial present value for liquidity actions
+                check_data["initial_present_value"] = interactive_hyperdrive.hyperdrive_interface.calc_present_value(
```
slundqui, line 152: Can you double check this? I think this value is getting overwritten: this loop first executes add liquidity and sets this parameter, then removes liquidity and resets this parameter.

dpaiton: the loop recalculates the initial before the trade for each trade; so the steps are
- calculate initial present value
- add liquidity
- check new pv against initial

- calculate initial pv
- remove liquidity
- check new pv against initial

dpaiton: I could remove the loops entirely, they are kind of overcomplicating parsing the test

slundqui: Ah okay, I see, didn't realize we check invariances at the end of every loop. In this case, this var should be set outside of the match

lib/agent0/agent0/interactive_fuzz/fuzz_present_value.py
```diff
@@ -118,14 +121,37 @@ def fuzz_present_value(
             case HyperdriveActionType.CLOSE_SHORT:
                 maturity_time, open_trade = next(iter(agent.wallet.shorts.items()))
                 trade_event = agent.close_short(maturity_time=maturity_time, bonds=open_trade.balance)
+            case _:
+                raise ValueError(f"Invalid {trade_type=}")
```
slundqui, line 125: I believe these trades are not getting checked in invariances, this for loop is simply opening trades. A quick scan shows there are indeed checks against these trades

dpaiton: yeah I just caught that too. one sec

slundqui (approved):
LGTM

PR #1279, opened 2024-01-24
Attempt a preview if a transaction call fails,

slundqui:
Adds the resulting preview error to the crash report


dpaiton (approved):
Awesome, hope this improves error messaging!

PR #1296, opened 2024-01-29
Update agent0 to hyperdrive v0.7.0

slundqui:


lib/ethpy/ethpy/test_fixtures/local_chain.py
```diff
@@ -217,7 +222,7 @@ def launch_local_hyperdrive_pool(
         minimumTransactionAmount=minimum_transaction_amount.scaled_value,
         positionDuration=position_duration,
         checkpointDuration=checkpoint_duration,
-        timeStretch=time_stretch.scaled_value,
+        timeStretch=0,
```
dpaiton, line 225: why is this zero now? if it's not used, shouldn't we delete it from the pool deploy config?

dpaiton (approved):
lgtm aside from confusion about time stretch parameter argument

lib/ethpy/ethpy/test_fixtures/local_chain.py
slundqui: This is what's in the migration script right now. This is required as it's a required field to the solidity call

PR #1307, opened 2024-02-06
Adding an option for running the dashboard in interactive hyperdrive

slundqui:


lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -950,6 +950,39 @@ def get_total_wallet_pnl_over_time(self, coerce_float: bool = False) -> pd.DataF
         ]
         return out
 
+    def get_dashboard_command(self) -> str:
+        """Prints the streamlit dashboard command to run to connect with the interactive hyperdrive.
+        The user can then copy/paste the command into a new terminal to run the dashboard.
+
+        .. note:: While there can be a function in interactive hyperdrive to actually run the dashboard,
```
dpaiton, line 957: you should check the preview deployment for this note, I think you'll need to put a newline after the `.. note::` tag to have it render correctly

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -950,6 +950,39 @@ def get_total_wallet_pnl_over_time(self, coerce_float: bool = False) -> pd.DataF
...
+        streamlit launches a web server under the hood, which doesn't play nice with the interactive process.
+        Here, our options are (1) make the dashboard blocking and (2) run it in a subprocess. If we make the dashboard
+        blocking, control is never passed back to the caller, which may affect interactive hyperdrive cleanup, as well
+        as introducing a deadlock in the interactive hyperdrive script. If we run the dashboard in a subprocess, it's up
+        to the caller to halt execution so that the server stays up and running. Neither of these options is ideal,
+        hence, we simply add a helper function to print the streamlit command and leave it to the user to control
+        the streamlit process.
```
dpaiton, line 964: I'm fine with this option. I think we could do a search for the subprocess (by storing the PID when it is created) and close it when interactive is shut down. But I don't know that this is necessarily a better option...

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
```diff
@@ -950,6 +950,39 @@ def get_total_wallet_pnl_over_time(self, coerce_float: bool = False) -> pd.DataF
...
+
+        .. note:: The interactive hyperdrive script must be in a paused state (before cleanup) for the dashboard to
```
dpaiton, line 966: How would a user make sure this is the case? I assume interactive HD is in a "pause state" much more often than not, but still we can't check this right?

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
slundqui: There's no issues around bookkeeping, running a subprocess returns an object we can access later to close cleanly.

dpaiton (approved):

lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive.py
slundqui: We can't check this. Paused state here really means we're at a breakpoint in a debugger or running in a notebook (without executing the cleanup line)

PR #1316, opened 2024-02-09
add tests of economic intuition

wakamex:
test ideas that may be intuitive, but turn out to be wrong


lib/agent0/tests/econ_tests.py
```diff
@@ -0,0 +1,28 @@
...
+
+from agent0.hyperdrive.interactive import InteractiveHyperdrive
+from agent0.hyperdrive.interactive.chain import Chain
+
+YEAR_IN_SECONDS = 31_536_000
+
+
+@pytest.mark.anvil
+def test_asymmetry(chain: Chain):
+    """Does in equal out?"""
```
dpaiton, line 15: can you add a more informative description for why this test is relevant or interesting?

dpaiton (approved):

PR #1320, opened 2024-02-13
Adding post action stage to bot runner

slundqui:


lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
```diff
@@ -123,6 +125,20 @@ async def async_execute_single_agent_trade(
             )
         trade_results.append(trade_result)
 
+    # TODO to avoid adding a post action in base policy, we only call post action
```
dpaiton, line 128: I'm not following this TODO. We have a base policy that HyperdriveBasePolicy inherits from.

lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
```diff
@@ -123,6 +125,20 @@ async def async_execute_single_agent_trade(
...
+    # way down
+    if isinstance(agent.policy, HyperdriveBasePolicy):
+        # Calls the agent with the trade results in case the policy needs to do bookkeeping
+        # We copy a subset of fields from the trade results to avoid changing the original
+        # trade result for crash reporting
+        # TODO deepcopy may be inefficient here when copying, e.g., trade_result.agent
+        # If this is the case, we can selectively create a new TradeResult object with a subset
+        # of data
+        trade_result_copy = deepcopy(trade_results)
+        # TODO can't put post_action in agent due to circular import, so we call the policy post_action here
```
dpaiton, line 139: what is the TODO? I think this looks fine to me, what would you like to change?

dpaiton (approved):

lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
slundqui: I went down the rabbit hole of adding this to the base agent and policy, and ultimately got to a dead end due to generics. Base policy isn't strictly a hyperdrive base policy, because hyperdrive base policy is explicitly setting the generic types of the market, whereas base policy takes a generic.

lib/agent0/agent0/hyperdrive/exec/execute_agent_trades.py
slundqui: Ideally, post_action would be an agent function, such that exec doesn't need to access the underlying policy. However, I kept running into circular imports when doing this.

PR #1321, opened 2024-02-14
Random hold bots

slundqui:


lib/agent0/agent0/hyperdrive/interactive/interactive_hyperdrive_policy.py
```diff
@@ -144,3 +144,22 @@ def action(
...
+    def post_action(self, interface: HyperdriveReadInterface, trade_results: list[TradeResult]) -> None:
+        """Function that gets called after actions have been executed. This allows the policy
+        to e.g., do additional bookkeeping based on the results of the executed actions.
+        Interactive hyperdrive policy calls the underlying sub policy's post action.
+
+        Arguments
+        ---------
+        interface: MarketInterface
+            The trading market interface.
+        trade_results: list[HyperdriveTradeResult]
```
sentilesdal, line 157: also doesn't match input

lib/agent0/agent0/hyperdrive/policies/random_hold.py
```diff
@@ -0,0 +1,313 @@
...
+        """Custom config arguments for this policy
+
+        Attributes
+        ----------
+        max_open_positions: int
+            The maximum number of open positions
+        min_hold_time: int
+            The minimum hold time in seconds. Defaults to 0
+        min_hold_time: int
+            The minimum hold time in seconds. Defaults to 2 * position_duration.
```
sentilesdal, line 51: is this true? code says zero.

lib/agent0/agent0/hyperdrive/policies/random_hold.py
```diff
@@ -0,0 +1,313 @@
...
+    def __init__(self, policy_config: Config) -> None:
+        """Initializes the bot
+
+        Arguments
+        ---------
+        policy_config: Config
+            The custom arguments for this policy
+        """
+        # Bookkeeping data structure for keeping track of open positions
+        # TODO using a list for now, but likely should use a different data structure
```
sentilesdal, line 80: I think a list is always going to fast enough here, I'd remove the TODO.  If we had a very very large number of positions open maybe, but on the order of 10's just going through the list and filtering is not bad when n is small for  an O(n) operation

sentilesdal: maybe could use an ordered list in the future that even that seems overkill.

lib/agent0/agent0/hyperdrive/policies/random_hold.py
```diff
@@ -0,0 +1,313 @@
...
+            position
+            for position in self.open_positions
+            if position.ready_to_close and position.action_type == HyperdriveActionType.OPEN_LONG
+        ]
+        # Sanity check, we should have at least one, otherwise close long wouldn't be an available action
+        assert len(longs_ready_to_close) > 0
+
+        # Select a random one
+        long_to_close = longs_ready_to_close[self.rng.integers(len(longs_ready_to_close))]
+        # Set flag that the transaction was sent
```
sentilesdal, line 195: why?  so we don't send it again?

lib/agent0/agent0/hyperdrive/policies/random_hold.py
```diff
@@ -0,0 +1,313 @@
...
+        return [close_short_trade(short_to_close.bond_amount, short_to_close.maturity_time, slippage)]
+
+    def post_action(self, interface: HyperdriveReadInterface, trade_results: list[TradeResult]) -> None:
+        """Function that gets called after actions have been executed. This allows the policy
+        to e.g., do additional bookkeeping based on the results of the executed actions.
+        Random hold updates open position bookkeeping if the submitted trade went through
+
+        Arguments
+        ---------
+        interface: MarketInterface
```
sentilesdal, line 250: doesn't match

lib/agent0/agent0/hyperdrive/policies/random_hold.py
```diff
@@ -0,0 +1,313 @@
...
+            The trading market interface.
+        trade_results: list[HyperdriveTradeResult]
```
sentilesdal, line 252: doesn't match

lib/agent0/agent0/hyperdrive/policies/random_hold.py
```diff
@@ -0,0 +1,313 @@
...
+
+        ignore_slippage = self.rng.choice([True, False], size=1) if self.randomly_ignore_slippage_tolerance else False
+        if ignore_slippage:
+            slippage = None
+        else:
+            slippage = self.slippage_tolerance
+        return [close_short_trade(short_to_close.bond_amount, short_to_close.maturity_time, slippage)]
+
+    def post_action(self, interface: HyperdriveReadInterface, trade_results: list[TradeResult]) -> None:
+        """Function that gets called after actions have been executed. This allows the policy
```
sentilesdal, line 244: ok but what is this fn actually doing then? :P

slundqui: The `post_action` function was added in a separate PR here: https://github.com/delvtech/agent0/pull/1320
The comments here do say what this function is doing in `post_action`, but I'll be sure to highlight it here

lib/agent0/agent0/hyperdrive/policies/random_hold.py
slundqui: This flag is for bookkeeping, the end of `post_action` removes positions with transactions that were sent (after making sure they succeed)

lib/agent0/agent0/hyperdrive/policies/random_hold.py
slundqui: While generally I agree with you that a list will likely be sufficient in the majority of cases, I'd argue that `n` in this case isn't on the order of 10's. Specifically, the total number of open positions possible is `(max_hold_time - min_hold_time)/time_per_block`. If you imagine a case where `max_hold_time` is based on position duration, `n` can be quite large. Hence, I'd argue I'll keep this TODO in.

PR #1327, opened 2024-02-21
Update for Hyperdrive v0.8.0

slundqui:


dpaiton (approved):
double check that we're not using the `vault_share_price` from events anywhere. other than that, looks great!

PR #1329, opened 2024-02-22
Load wallet state from chain

slundqui:


lib/agent0/agent0/hyperdrive/exec/run_agents.py
```diff
@@ -380,3 +391,99 @@ def build_wallet_positions_from_data(
...
+        if maturity_time in long_obj:
+            long_obj[maturity_time].balance += long_amount
+        else:
+            long_obj[maturity_time] = Long(balance=long_amount, maturity_time=maturity_time)
+    # Close Longs
+    close_long_events = hyperdrive_contract.events.CloseLong.get_logs(fromBlock=0)
+    for event in close_long_events:
+        maturity_time = event["args"]["maturityTime"]
+        long_amount = FixedPoint(scaled_value=event["args"]["bondAmount"])
+        assert maturity_time in long_obj
```
dpaiton, line 448: can you add error messages to these asserts? like "wallet had close event without corresponding open event" and "wallet had negative long balance, which is not supported"

lib/agent0/agent0/hyperdrive/exec/run_agents.py
```diff
@@ -380,3 +391,99 @@ def build_wallet_positions_from_data(
         longs=long_obj,
         shorts=short_obj,
     )
+
+
+def build_wallet_positions_from_chain(
```
dpaiton, line 396: I think we should put this in a file with `*_from_data`. so `build_wallet_positions.py`. We've been tending in that direction, and it's obvious where the test is (`build_wallet_positions_test.py`).

lib/agent0/agent0/hyperdrive/exec/run_agents.py
```diff
@@ -380,3 +391,99 @@ def build_wallet_positions_from_data(
...
+
+    # Contract call to get withdrawal positions
+    asset_id = encode_asset_id(AssetIdPrefix.WITHDRAWAL_SHARE, 0)
+    withdraw_amount: int = hyperdrive_contract.functions.balanceOf(asset_id, wallet_addr).call()
+    withdraw_obj = FixedPoint(scaled_value=withdraw_amount)
+
+    # We need to gather all longs and shorts from events
+    # and rebuild the current long/short positions
+    # Open Longs
+    open_long_events = hyperdrive_contract.events.OpenLong.get_logs(fromBlock=0)
```
dpaiton, line 433: my understanding from our earlier convo about this is that it is not guaranteed that we can get all events from block 0. I think we should wrap these in a try/except that throws a more detailed error if it fails (e.g. "Unable to load events from block 0, are you using a node that stores all historic data?")

slundqui: My takeaway from this is that we're always to get logs regardless of the chain, it just might be slow.

lib/agent0/agent0/hyperdrive/exec/run_agents.py
slundqui: this is hard to unit test, I need the system as a whole to test wallet states, hence the test is in the outer directory. I can move these to a separate file.

lib/agent0/agent0/hyperdrive/exec/run_agents.py
dpaiton: I don't know how to query what the oldest available block is... if that were possible we could have an alternate flow that gets the earliest possible and posts a warning? The user might not care that we can't access block 0 if the bot is relatively new.

lib/agent0/agent0/hyperdrive/exec/run_agents.py
dpaiton: sure we can punt on tests for now, just mentioning that as a future goal.

lib/agent0/agent0/hyperdrive/exec/run_agents.py
dpaiton: > My takeaway from this is that we're always to get logs regardless of the chain, it just might be slow.

I'm not sure about that, will have to verify.  I think it depends on your provider, which changes with the rpc url given to web3.

dpaiton: even our local devnet with anvil sometimes fails to give us block 0, right?

dpaiton: even if your takeaway is correct we can go with my original proposal (try/except) as a jic, and then put a TODO to more cleverly use earliest available block in a new PR

slundqui: We should revisit this discussion, logs are always available (may be slow), its only if we want to get data from blocks via an anvil specific thing is where it fails. If logs are not always available, then we have bigger problems because frontend depends on this process as well.

slundqui: I don't think we need a try/catch here, we probably want the code to crash if this isn't the case (again, we have a huge issue in frontend if this isn't the case), hence, I'd rather not catch exceptions and have execution fail.

dpaiton (approved):
awesome, this will be very useful!

PR #1330, opened 2024-02-26
LP bot that adds or removes liquidity depending on current PNL

dpaiton:
The bot has this description:

>         This LP bot adds or removes liquidity according to its recent profitability.
>         It has a target PNL that it is aiming for, and takes actions based on how its PNL has changed over time.
>         The bot follows this strategy:
>             - If the PNL is above threshold, and is improving over time, then the bot adds liquidity.
>             - If the PNL is above threshold, and is not improving over time, then the bot takes no action.
>             - If the PNL is below threshold, and is improving over time, then the bot takes no action.
>             - If the PNL is below threshold, and is not improving over time, then the bot pulls liquidity.

The target PNL is set in units of base, which I think is intuitive enough. We also use PNL history to bias our decisions, so it doesn't remove liquidity too early.

The bot can definitely be improved & fine-tuned, but I think this is sufficient to get started with our experiment.

### DIFF
--- a/src/agent0/core/hyperdrive/policies/simple_lp.py
+++ b/src/agent0/core/hyperdrive/policies/simple_lp.py
@@ -48,16 +48,20 @@ class SimpleLP(HyperdriveBasePolicy):
     class Config(HyperdriveBasePolicy.Config):
         """Custom config arguments for this policy."""
 
-        # The target PNL for the bot, in base.
         pnl_target: FixedPoint = FixedPoint("100")
-        # How much liquidity to add or remove, depending on policy outcome, in base.
+        """The target PNL for the bot, in base."""
         delta_liquidity: FixedPoint = FixedPoint("100")
-        # Minimum liquidity the bot will provide.
-        # It will keep this much liquidity in the pool, even if it is losing money.
+        """How much liquidity to add or remove, depending on policy outcome, in base."""
         minimum_liquidity_tokens: FixedPoint = FixedPoint("100")
-        # How many steps back to look when computing PNL progress over time.
-        # Each time `action` is called is a step.
+        """
+        Minimum liquidity the bot will provide.
+        It will keep this much liquidity in the pool, even if it is losing money.
+        """
         lookback_length: int = 10
+        """
+        How many steps back to look when computing PNL progress over time.
+        Each time `action` is called is a step.
+        """
 
     def __init__(
         self,
--- a/src/agent0/core/hyperdrive/policies/simple_lp.py
+++ b/src/agent0/core/hyperdrive/policies/simple_lp.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING
 from fixedpointmath import FixedPoint, FixedPointMath
 
 from agent0.base import Trade
-from agent0.hyperdrive import HyperdriveMarketAction
+from agent0.hyperdrive import HyperdriveMarketAction, TradeResult
 from agent0.hyperdrive.agent import add_liquidity_trade, remove_liquidity_trade
 
 from .hyperdrive_policy import HyperdriveBasePolicy
@@ -48,13 +48,15 @@ class SimpleLP(HyperdriveBasePolicy):
             How many steps back to look when computing past PNL.
             Each time `action` is called is a step.
         pnl_target: FixedPoint
-            The target PNL for the bot to achieve, as a fraction improvement over the pnl at lookback_length
+            The target yearly PNL growth for the bot to achieve,
+            as a fraction improvement over the pnl from previous recordings.
         delta_liquidity: FixedPoint
             How much liquidity to add or remove, depending on policy outcome.
         """
 
-        lookback_length: FixedPoint = FixedPoint("10")  # action calls
-        pnl_target: FixedPoint = FixedPoint("1.0")
+        # pnl_target = 0.1 means the time-weighted average profit_change is 10% increase
+        pnl_target: FixedPoint = FixedPoint("0.1")
+        lookback_length: FixedPoint = FixedPoint("10")  # action calls (usually blocks)
         delta_liquidity: FixedPoint = FixedPoint("100")
 
     def __init__(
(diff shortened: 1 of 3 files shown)


lib/agent0/agent0/hyperdrive/policies/simple_lp.py
```diff
@@ -0,0 +1,177 @@
...
+
+from agent0.base import Trade
+from agent0.hyperdrive import HyperdriveMarketAction, TradeResult
+from agent0.hyperdrive.agent import add_liquidity_trade, remove_liquidity_trade
+
+from .hyperdrive_policy import HyperdriveBasePolicy
+
+if TYPE_CHECKING:
+    from ethpy.hyperdrive import HyperdriveReadInterface
+
```
wakamex, line 18: extra space here?

wakamex:
I just realized percentage change on a pnl value that starts at 0 makes no sense.
How about `pnl/self.total_base_spent*YEAR_IN_SECONDS/(current_time-self.start_time)` which requires no weighted average

dpaiton:
> I just realized percentage change on a pnl value that starts at 0 makes no sense. That metric is independent of how much money you deployed. Making $1 a day on $1 million spent is very different from $1 a day on $10 spent. How about `pnl/self.total_base_spent*YEAR_IN_SECONDS/(current_time-self.start_time)` which requires no weighted average

Agreed; I think the time-weighted average PNL is not making much sense. I'm going to just use the net PNL for now: `pnl = current_balance - total_amount_supplied`. If this is `> 0`, the agent adds LP. If it's `< 0`, they remove LP. If it `= 0`, they sit tight.

src/agent0/core/hyperdrive/policies/simple_lp.py
```diff
@@ -0,0 +1,172 @@
...
+            A tuple where the first element is a list of actions,
+            and the second element defines if the agent is done trading.
+        """
+        # Get the current state of the pool & the bot's position
+        current_block = interface.get_current_block()
+        pool_state = interface.get_hyperdrive_state(current_block)
+        current_holding = wallet.lp_tokens * pool_state.pool_info.lp_share_price
+
+        # Need to be in the game to play it
+        if self.policy_config.minimum_liquidity_tokens > current_holding:
```
wakamex, line 107: compare this to just wallet.lp_tokens or rename the variable to _value

src/agent0/core/hyperdrive/policies/simple_lp.py
```diff
@@ -0,0 +1,172 @@
...
+            trade_amount = self.policy_config.minimum_liquidity_tokens - current_holding
+            return [add_liquidity_trade(trade_amount)], False
```
wakamex, line 109: I think trade_amount is in base, so sounds like you want everything to be in base (not tokens? maybe you were thinking of base with _tokens, but I'd clarify, since I thought of lp_tokens)

src/agent0/core/hyperdrive/policies/simple_lp.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # Get current PNL
+        pnl = current_holding - self.total_base_spent
+        self.pnl_history.append((FixedPoint(interface.get_block_timestamp(current_block)), pnl))
+        # Prune history
+        if FixedPoint(len(self.pnl_history)) > self.policy_config.lookback_length:
+            self.pnl_history = self.pnl_history[-self.policy_config.lookback_length :]
+        # PNL slope is dy/dx, where y is PNL and x is time.
```
wakamex, line 117: trying to think about what this algorithm does, i think a simple average gives equal weight to each observation, even though the time passed may vary drastically. not sure what this adds over just taking dy/dx of the whole sample, which would avoid that issue, while being simpler. but we know the choice of return metric is arbitrary and has many implementations, so as a starting point, this is fine.

src/agent0/core/hyperdrive/policies/simple_lp.py
```diff
@@ -0,0 +1,172 @@
...
+                dx = self.pnl_history[i + 1][0] - self.pnl_history[i][0]
+                dy = self.pnl_history[i + 1][1] - self.pnl_history[i][1]
+                avg_pnl_slope += dy / dx
+            avg_pnl_slope /= len(self.pnl_history) - 1
+
+        # Make trades based on pnl_target and avg_pnl_slope
+        action_list = []
+
+        # I'm doing great, keep putting money in
+        if pnl > self.policy_config.pnl_target and avg_pnl_slope > 0:
```
wakamex, line 132: pnl is a stock and slope is a flow. I think you want the pnl target on the flow measure.

any stock-based measure could become useless as your total wealth moves far away from that point.

wakamex: to keep `SimpleLP` as simple as possible, I'd simplify the logic to just the flow measure, and accept they'll always add or remove. or explicitly define a do-nothing range, which in this case would be in units of base/second I think. or do my original suggestion to make your target liquidity proportional to your rate of return relative to a high target and your budget: `target_liquidity = budget * max(0,min(rate_of_return / 10%,1))` which allows for varying trade sizes and faster convergence.

src/agent0/core/hyperdrive/policies/simple_lp.py
```diff
@@ -0,0 +1,172 @@
...
+            # only add money if you can afford it!
+            if wallet.balance.amount >= self.policy_config.delta_liquidity:
+                action_list.append(add_liquidity_trade(self.policy_config.delta_liquidity))
+
+        # I'm doing bad, time to pull some money out
+        elif pnl < self.policy_config.pnl_target and avg_pnl_slope < 0:
+            # delta_liquidity is in base; convert it to tokens
+            delta_tokens = self.policy_config.delta_liquidity / pool_state.pool_info.lp_share_price
+            # we do not want to pull out so many tokens that we are below the minimum
```
wakamex, line 141: to keep it simple, i think you can take out this check. since you actually want the minimum to be triggered, to act as a "reset" to make the bot keep trying. otherwise it will just sit at just outside the threshold and never trade again.

src/agent0/core/hyperdrive/policies/simple_lp_test.py
```diff
@@ -0,0 +1,98 @@
...
+        base=FixedPoint("1_000_000"), eth=FixedPoint("1_000"), name="Bob"
+    )
+    trade_event_list = []
+    removed_liquidity = False
+    while not removed_liquidity:
+        trade_amount = FixedPoint("1_000")
+        hyperdrive_agent1.add_funds(base=trade_amount)
+        interactive_hyperdrive.set_variable_rate(FixedPoint("0.8"))
+        open_event = hyperdrive_agent1.open_long(trade_amount)
+        chain.advance_time(datetime.timedelta(seconds=100), create_checkpoints=False)
```
wakamex, line 83: not sure why this and the `interactive_hyperdrive.set_variable_rate(FixedPoint("0.8"))` above are needed

we're making LPs lose money by selling fixed rates above the variable rate, right? so setting variable rate to 0 should suffice

wakamex (approved):
looks good, some nits and algorithm discussion, should work as `SimpleLP` either way... but can it be simpler? 🤔

using fundamental theorem of calculus

$$ \quad \lim_{n\to\infty} \sum_{i=1}^{n} f(x_i^*)\Delta x = F(b) - F(a) $$

divide by `b-a`:

$$ \frac{1}{b-a}\lim_{n\to\infty} \sum_{i=1}^{n} f(x_i^*)\frac{b-a}{n} = \frac{F(b) - F(a)}{b-a} $$

$$
\lim_{n\to\infty} \sum_{i=1}^{n} \frac{f(x_i^*)}{n} = \frac{F(b) - F(a)}{b-a}
$$

average slope of n equal pieces = aggregate slope 🙌

src/agent0/core/hyperdrive/policies/simple_lp.py
dpaiton: I opted against using `(last_y - first_y) / (last_x - first_x)` because it puts a lot of emphasis on the first and last points. For example, if the LP had a large spike in PNL in the middle, but at the end points earned approximately the same PNL, then the slope would be ~0. I think We don't want this; we want the metric to be an average of PNL changes over the time span.

Regardless of whether we average each spline or we take the global average, the delta x (time passed) is arbitrary. However, by computing a slope, I think we are normalizing to whatever that time passed may be.

src/agent0/core/hyperdrive/policies/simple_lp.py
dpaiton: hmm this is my first time hearing "stock" vs "flow", but I do think the direction we _were_ going with this was the slope as the metric. However, I had a hard time thinking about what units made sense and so I switched to the "stock" value. I think it's worth considering this metric further, but if agree that this is good enough then I'd like to avoid reworking it again in favor of getting an mvp out the door.

src/agent0/core/hyperdrive/policies/simple_lp.py
dpaiton: It has a minimum liquidity amount, so it'll always have some stake in the game. We want the liquidity in the pool to go down when the LP is unhappy so that traders will also be unhappy due to slippage. So I think we need this to stay in so that the behavior is two-sided.

src/agent0/core/hyperdrive/policies/simple_lp_test.py
dpaiton: we set the rate high so that the open long gets that fixed rate, which means the LP has to supply it. And then we set the variable to zero before advancing a larger amount of time (1 week) so that the LP doesn't get any kickback from the variable while they're providing the fixed. It's a shit scenario for them, which was the intention so that I'd be sure they pulled their liquidity.

dpaiton:
> using fundamental theorem of calculus
> 
> limn→∞∑i=1nf(xi∗)Δx=F(b)−F(a)
> 
> divide by `b-a`:
> 
> 1b−alimn→∞∑i=1nf(xi∗)b−an=F(b)−F(a)b−a
> 
> limn→∞∑i=1nf(xi∗)n=F(b)−F(a)b−a
> 
> average slope of n equal pieces = aggregate slope 🙌

Does this apply when the distance between splines is variable & unknown? I don't think so.

src/agent0/core/hyperdrive/policies/simple_lp.py
wakamex: you're right the statement doesn't hold when the segments vary in size. but the current implementation is a simple average across slopes, not weighted by segment size. so your rate of return (slope, not pnl) over a 1yr period would get the same weight as 1block period. to add in the time weighting would just be multiplying by dx... which means just not dividing by dx in the first place, so might actually be simpler and gives us pnl instead of rate of return. which at that point is the same as the aggregate, since it's just the sum of dy.

also ignoring 1 really bad trade if most of your trades are good is not a good idea. you're thinking of it as an outlier that doesn't represent the actual distribution. but some distributions are actually fat-tailed left-skewed, like selling lottery tickets or [fixed income arbitrage](https://www.johnhcochrane.com/s/Duarte_Longstaff_Yu_nickelsRFS.pdf) which is actually what we're doing here, fixed income just means rates, lovingly referred to as "picking up nickels in front of a steamroller" (you should consider the steamroller!)

src/agent0/core/hyperdrive/policies/simple_lp.py
wakamex: sounds good, fine with not reworking

src/agent0/core/hyperdrive/policies/simple_lp.py
wakamex: oh right, it doesn't need to "add liquidity" for a chance to improve. it just needs to "keep something open" which this does!

src/agent0/core/hyperdrive/policies/simple_lp_test.py
dpaiton: correct, they're independent. I don't know why I had that mixed up! Good catch.

PR #1331, opened 2024-02-26
improve arb bot accuracy

wakamex:
handle 100x bigger trades within same absolute tolerance


lib/agent0/agent0/utilities/predict.py
```diff
@@ -0,0 +1,234 @@
...
+from ethpy.hyperdrive.state import PoolState
+from fixedpointmath import FixedPoint
+
+# need to cover a wide variety of cases
+# pylint: disable=too-many-arguments
+# being very explicit
+# pylint: disable=too-many-locals
+
+YEAR_IN_SECONDS = 31_536_000
+BLOCKS_IN_YEAR = YEAR_IN_SECONDS / 12
```
dpaiton, line 19: I think you either want

`SECONDS_IN_YEAR` & `BLOCKS_IN_YEAR` to mean "the number of X (seconds or blocks) in Y (a year)"

OR

`YEAR_IN_SECONS` & `YEAR_IN_BOCKS` to mean "a year, but in units of X (seconds or blocks)"

You currently mix & match the two

lib/agent0/agent0/utilities/predict.py
```diff
@@ -0,0 +1,234 @@
...
+    hyperdrive_interface: HyperdriveReadInterface
+        Hyperdrive interface.
+    pool_state: PoolState, optional
+        The state of the pool, which includes block details, pool config, and pool info.
+        If not given, use the current pool state.
+    base: FixedPoint, optional
+        The size of the long to open, in base. If not given, converted from bonds.
+    bonds: FixedPoint, optional
+        The size of the long to open, in bonds.
+    for_pool: bool
```
dpaiton, line 61: You have `for_pool = True` in the only use case (`lpandarb.py`) and `for_pool=False` in all of the tests in `predict_trade_test.py`. To simplify the code _and_ have the test more accurately reflect the usage, I recommend you remove this variable entirely, only include the `for_pool=True` branch, and modify the test to assume this is the case.

lib/agent0/agent0/utilities/predict.py
```diff
@@ -0,0 +1,234 @@
...
+        logging.info("predict_long(): bond_fees_to_pool is %s", bond_fees_to_pool)
+        logging.info("predict_long(): bond_fees_to_gov is %s", bond_fees_to_gov)
+    predicted_delta_bonds = -bonds_after_fees - bond_fees_to_gov
+    # gov_scaling factor is the ratio by which we lower the change in base and increase the change in shares
+    # this is done to take into account the effect of the governance fee on pool reserves
+    gov_scaling_factor = FixedPoint(1) - price_discount * curve_fee * governance_fee
+    predicted_delta_base = base_needed * gov_scaling_factor
+    predicted_delta_shares = base_needed / share_price * gov_scaling_factor
+    if verbose:
+        logging.info("predict_long(): predicted pool delta bonds is %s", predicted_delta_bonds)
```
dpaiton, line 120: Can we get rid of these verbose blocks? I think an alternative, if you think an end-user would want to know, is to use `logging.debug` but even then keep in mind our recommended workflow is to use the debugger & breakpoints to inspect variables, so this still seems like it's unnecessary.

lib/agent0/agent0/utilities/predict.py
```diff
@@ -0,0 +1,234 @@
...
+        ),
+        governance=Deltas(
+            bonds=bond_fees_to_gov,
+            base=bond_fees_to_gov * spot_price,
+            shares=bond_fees_to_gov * spot_price * share_price,
+        ),
+    )
+
+
+def predict_short(
```
dpaiton, line 143: don't need `for_pool` or `verbose` here either

lib/agent0/agent0/utilities/predict.py
```diff
@@ -0,0 +1,234 @@
...
+        base_needed = base
+        # this is the wrong direction for the swap, but we don't have the function in the other direction
+        bonds_needed = hyperdrive_interface.calc_bonds_out_given_shares_in_down(base_needed / share_price)
+        # scale down output to account for fees
+        if for_pool is False:
+            bonds_needed /= FixedPoint(1) - price_discount * curve_fee * (FixedPoint(1) - governance_fee)
+        else:
+            bonds_needed /= FixedPoint(1) - price_discount * curve_fee
+    else:
+        raise ValueError("predict_short(): Need to specify either bonds or base, but not both.")
```
dpaiton, line 198: the function name should show up in the error stack trace, so you don't need it here.

dpaiton (approved):
You can convert my requests into issues if you want, but I recommend that you just fix them now. Approved pending those fixes, or issues.

wakamex:
as discussed, let's create a real world example for the alternate code path in predict_trade (`for_pool=False`) or remove it by March 4th

lib/agent0/agent0/utilities/predict.py
wakamex: changed to YEAR_IN_BLOCKS

lib/agent0/agent0/utilities/predict.py
wakamex: created tracking issue, left comment in PR

wakamex:
implemented, just needs review, finishing touches, and to be merged

PR #1340, opened 2024-03-05
adds a function for getting an IFrame to display the dashboard inside jupyter notebooks

dpaiton:
This PR adds the ability to get an IFrame for a deployed dashboard. You can then use this object to display the dashboard inside a Jupyter notebook.

When merged, it will resolve issue https://github.com/delvtech/agent0/issues/1338

### DIFF
--- a/lib/agent0/agent0/hyperdrive/interactive/i_local_hyperdrive.py
+++ b/lib/agent0/agent0/hyperdrive/interactive/i_local_hyperdrive.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import os
 import pathlib
-import re
 import subprocess
 import time
 from dataclasses import asdict, dataclass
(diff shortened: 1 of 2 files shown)


lib/agent0/agent0/hyperdrive/interactive/i_local_hyperdrive.py
```diff
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import os
 import pathlib
+import re
```
slundqui, line 7: Is this import still being used?

slundqui (approved):

lib/agent0/agent0/hyperdrive/interactive/i_local_hyperdrive.py
dpaiton: oops no sorry forgot to run lints

PR #1341, opened 2024-03-05
move isclose check to fixedpointmath; update to new import structure

dpaiton:


lib/agent0/agent0/interactive_fuzz/fuzz_long_short_maturity_values.py
```diff
@@ -322,7 +321,7 @@ def invariant_check(
         expected_long_base_amount = close_trade_event.bond_amount - close_trade_event.bond_amount * flat_fee_percent
 
         # assert with close event bond amount
-        if not fp_isclose(
+        if not FixedPointMath.isclose(
```
slundqui, line 324: from a ui pov, I think exposing the function `isclose` from `fixedpointmath` is cleaner:
```
from fixedpointmath import isclose
isclose(...)
```

dpaiton: Clarifying your request:

In the lib we differentiate `FixedPoint` (the type) from `FixedPointMath` a class with math functions. In the lib `__init__` we have:

```py
from .fixed_point import FixedPoint
from .fixed_point_integer_math import FixedPointIntegerMath
from .fixed_point_math import FixedPointMath
```

To support `from fixedpointmath import isclose` I would need to pull the fns out of the `FixedPointMath` class and modify the `__init__` to include `from .fixed_point_math import *`. This would also lift functions like `exp`, `clip`, etc. Is that what you are recommending?

dpaiton: after thinking about it more I agree what you suggest is better, but it's a big enough change that it should be it's own PR.

tmp.py
```diff
@@ -0,0 +1,61 @@
+# %%
+from fixedpointmath import FixedPoint, FixedPointMath
```
slundqui, line 2: This doesn't look like it should be in the repo

slundqui (approved):

tmp.py
dpaiton: oops crap thanks

slundqui (approved):

PR #1344, opened 2024-03-05
make root folder installable

wakamex:


wakamex:
pytest currently fails with:
```
__________________________________________________________ ERROR collecting lib/hyperlogs/hyperlogs/logs_test.py __________________________________________________________
import file mismatch:
imported module 'hyperlogs.logs_test' has this __file__ attribute:
  /code/elfpy/.venv/lib/python3.10/site-packages/hyperlogs/logs_test.py
which is not the same as the test file we want to collect:
  /code/elfpy/lib/hyperlogs/hyperlogs/logs_test.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

don't see a way around that since the tests are testing the local files but the package is installed in the venv. the solution is to have these libraries all as editable installs, then they're installed in-place. but then we need pyproject.toml's for each one...

slundqui:
Closing in favor of https://github.com/delvtech/agent0/pull/1347

PR #1352, opened 2024-03-08
test managing agent budget

wakamex:
use predict_trade to constrain trade size within target budget, as discussed here: https://github.com/delvtech/agent0/pull/1331#issuecomment-1965509581


dpaiton:
Closing; we will address this when we resolve #1514

wakamex:
the original motivation of this was to show the usefulness of `for_pool=False` code path

PR #1353, opened 2024-03-08
store open share price in Short

wakamex:


src/agent0/ethpy/hyperdrive/transactions.py
```diff
@@ -163,6 +163,13 @@ def parse_logs(tx_receipt: TxReceipt, hyperdrive_contract: Contract, fn_name: st
         if value in log_args and hasattr(trade_result, camel_to_snake(value)):
             setattr(trade_result, camel_to_snake(value), FixedPoint(scaled_value=log_args[value]))
 
+    # special handling for vault_share_price since it's not emitted, but calculated as base_amount / vault_share_amount
+    setattr(
+        trade_result,
+        "vault_share_price",
+        FixedPoint(scaled_value=log_args["baseAmount"]) / FixedPoint(scaled_value=log_args["vaultShareAmount"]),
```
slundqui, line 170: Tests are saying not all results contain `baseAmount` or `vaultShareAmount` (e.g., in `createCheckpoint`). Lets wrap this in a check for these keys.

wakamex:
not needed if we use get_checkpoint to look up the vault share price at the time the short was opened. although that's a bit wasteful of RPC calls.

PR #1356, opened 2024-03-11
cleanup lpandarb bot

dpaiton:
I have been looking over this file closely to spec out the `targeted_long` and `target_rate_deadzone` logic. This is an interim PR since I ended up making some superficial changes:
- reorganized use of temporary PoolState objects to reduce the number of slow `deepcopy` calls
- reorganized arguments & fn placement to be consistent
- extract most of the `action` fn code into new fns
- added docs and type hints
- cleaned up inefficiencies with repeated RPCs
- unified logging format & deleted a time-to-completion log that required a lot of overhead (I would recommend that we further remove a bulk of these logs, and move others from `info` to `debug`, but I didn't want to step on anyone's use case without prior discussion)
- moved some logic to the base class
- added explicit use of the current pool state to avoid accidental race conditions
- switches use of stdlib `min` and `max` to the `fixedpointmath` version; there was a type error before that could have resulted in a loss of precision.
- added a test & fix for a bug where the agent could spend it's money adding liquidity and then try to spend it again arbing

The existing tests were not changed, and all still pass. None of the underlying logic should have changed, aside from the bug fix and avoiding race conditions. However I made changes to when temp variables were used, so I recommend a close look to catch anything the tests may have missed.

On my machine, before this PR the `lpandarb_test.py` file took 25.45 seconds to complete, now it takes 24.99 seconds (excluding the two new tests).


src/agent0/core/hyperdrive/policies/lpandarb_test.py
```diff
@@ -318,3 +318,35 @@ def test_reduce_short(interactive_hyperdrive: ILocalHyperdrive, arbitrage_andy:
...
+    larry_config = PolicyZoo.lp_and_arb.Config(
+        lp_portion=FixedPoint("0.99"),
+        minimum_trade_amount=interactive_hyperdrive.interface.pool_config.minimum_transaction_amount,
+    )
+    larry = interactive_hyperdrive.init_agent(
+        base=larry_base, name="larry", policy=PolicyZoo.lp_and_arb, policy_config=larry_config
+    )
+    # change the fixed rate
+    manual_agent.open_short(bonds=FixedPoint(1_000))
+    larry.execute_policy_action()  # should not be able to arb the full amount after LPing
```
wakamex, line 336: check that no action was done here, I believe, since they don't have the budget to put on a trade above the minimum trade size? without eating into the lp portion

src/agent0/core/hyperdrive/policies/lpandarb_test.py
```diff
@@ -318,3 +318,35 @@ def test_reduce_short(interactive_hyperdrive: ILocalHyperdrive, arbitrage_andy:
...
+    larry_config = PolicyZoo.lp_and_arb.Config(
+        lp_portion=FixedPoint("0.99"),
+        minimum_trade_amount=interactive_hyperdrive.interface.pool_config.minimum_transaction_amount,
+    )
+    larry = interactive_hyperdrive.init_agent(
+        base=larry_base, name="larry", policy=PolicyZoo.lp_and_arb, policy_config=larry_config
+    )
+    # change the fixed rate
+    manual_agent.open_long(base=FixedPoint(10_000))
+    larry.execute_policy_action()  # should not be able to arb the full amount after LPing
```
wakamex, line 352: check for no action being done here also?

src/agent0/core/hyperdrive/policies/lpandarb_test.py
dpaiton: I was anticipating that they do take an arb action, just not the optimal one because they have limited funds remaining (the remaining 0.01 portion). However, I think it's a good idea to check this.

src/agent0/core/hyperdrive/policies/lpandarb.py
```diff
@@ -70,119 +202,186 @@ def calc_shares_needed_for_bonds(
...
-    """Calculate the bonds and shares needed to hit the target fixed rate.
+def calc_delta_reserves_for_target_rate(
+    interface: HyperdriveReadInterface,
+    pool_state: PoolState,
+    target_rate: FixedPoint,
+    minimum_trade_amount: FixedPoint,
+) -> tuple[FixedPoint, FixedPoint]:
+    """Calculate the bonds needed to hit the desired reserves ratio, keeping shares constant.
+
+    delta_bonds tells us the number of bonds to hit the desired reserves ratio, keeping shares constant.
```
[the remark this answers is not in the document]
dpaiton: I clarified the variable names to be explicit about which was wrt the trade and which was wrt the pool

jrhea (approved):

PR #1360, opened 2024-03-13
update README; docs; autoupload

dpaiton:
This PR
- various updates to switch to installing hyperdrivepy via pip
- updates the install, build, readme docs (incl  new disclaimer)
- adds the ability to autoupload agent0 to pypi so it can be pip installed

note that I got rid of the urllib dependency version peg; we're not using vercel anymore so this should be fine but we'll want to double check docs build before merging

### DIFF
--- a/.github/workflows/build_and_upload.yml
+++ b/.github/workflows/build_and_upload.yml
@@ -1,0 +1,77 @@
+name: build and upload wheels
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  detect-version-changes:
+    uses: ./.github/workflows/check_version.yml
+    with:
+      file_path: crates/hyperdrivepy/pyproject.toml
+
+  build-wheels:
+    needs: detect-version-changes
+    # Run on main if version has changed
+    if: needs.detect-version-changes.outputs.version_changed == 'true'
+    name: build wheel
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{github.token}}
+
+      - name: set up python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+          cache: "pip"
+          token: ${{github.token}}
+
+      - name: set up pip
+        run: python -m pip install --upgrade pip
+
+      - name: build hyperdrivepy
+        shell: bash
+        run: source scripts/build_wheels.sh
+
+      - name: upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheels-linux
+          path: ./wheelhouse/*.whl
+
+  build-sdist:
+    needs: detect-version-changes
+    name: Build source distribution
+    runs-on: ubuntu-latest
+    # Run on main if version has changed
+    if: needs.detect-version-changes.outputs.version_changed == 'true'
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -11,8 +11,8 @@ COPY . ./
 RUN apt-get update && \
   apt-get install -y --no-install-recommends gcc python3-dev libssl-dev git && \
   python -m pip install --no-cache-dir --upgrade pip && \
-  python -m pip install --no-cache-dir . && \
-  python -m pip install --no-cache-dir -r requirements-hyperdrivepy.txt && \
+  python -m pip install --no-cache-dir uv && \
+  python -m uv pip install --no-cache-dir agent0@.[all] && \
```
slundqui, line 15: Adding this bloats the docker container to also contain e.g., pytorch

.github/workflows/build_and_upload.yml
```diff
@@ -0,0 +1,77 @@
...
+          path: dist/*.tar.gz
+
+  upload_pypi:
+    needs: [build-wheels, build-sdist, detect-version-changes]
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write
+    # Run on main if version has changed
+    if: needs.detect-version-changes.outputs.version_changed == 'true'
```
slundqui, line 69: I know this is what I did for hyperdrivepy, but I'm leaning more towards releasing on tags better. Gives us a bit more control on when to upload to pypi. Reason why I went with version checking for hyperdrivepy is because of potential collusion with hyperdrive-wasm, which I believe publishes on tags.

scripts/build_wheels.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# agent0 install
+
+echo "install required packages for building wheels"
+python -m pip install --upgrade pip uv build
+uv pip install agent0@.[all]
```
slundqui, line 7: This is installing agent0 from pip, where we're then building a new version. This should be an editable install locally.
```
uv pip install -e .[all]
```

slundqui: scratch that, I think this installs a non-editable install locally, which I think is okay

slundqui (approved):

Dockerfile
dpaiton: agreed, so we should just do [dev] then?

.github/workflows/build_and_upload.yml
dpaiton: I like tags. I can make the change.

Dockerfile
slundqui: IIRC we just did no dependencies and it worked fine, so:
```
python -m uv pip install --no-cache-dir agent0@. &&
```
I did look over this with the uv install PR and purposefully left this as pip, to also avoid having uv in the image as well, but I didn't look too deeply on how big uv was.

dpaiton: uv is huge (~25mb I think); so I'd rather not have it... ok I'll revert this to the original command. We should think about building in a test for validating our INSTALL instructions though.

wakamex (approved):
lgtm, fresh install passes tests without urllib3 peg

PR #1362, opened 2024-03-13
refactor exec folder into separate test and IHyperdrive dependencies

dpaiton:
This PR is strictly a refactor; no new logic was implemented (see caveats below).

A majority of the functions in the `agent0.core.exec` folder were only used by tests. A smaller subset of them were used by `IHyperdrive`. I split these up, and put the IHyperdrive ones in the `interactive` folder. I consolidated the rest into `agent0.core.hyperdrive.utilities`, which now also contains the predict files from `agent0.core.utilities`.

Now that we've severed the connection to interactive hyperdrive, the functions located in `agent0.core.hyperdrive.utilities.run_bots` can be paired down in a follow-up PR. Nearly all of the functions there are not used anywhere else in the repo; strictly in tests. We should set up example flows that would use them, and if we can't come up with any then we should delete them and the corresponding tests. This will be done in a follow-up PR, though.

caveats (functional changes):
- I made some functions 'private' (i.e. with an underscore in the name) if they were only used by other functions inside the same file.
- I made `set_max_approval` non-async. It was not being used as an async function in the first place because it was only used by [`IHyperdrive` and always with the list length of 1](https://github.com/delvtech/agent0/blob/98288dadf6815921bdeb701fe7318ee4ba124ef4/src/agent0/core/hyperdrive/interactive/i_hyperdrive.py#L209). This means as used there would always be only one thread. So I made it non-async to remove unnecessary and unused code, with the idea that we can make it async later if we have a use-case for it.
- the `get_agent_trades` flow was in a similar situation -- we had a function for calling it with a list of agents but then hard-coded a single `[agent]` input whenever it was used. I left parts async that made sense to be (matching contract calls to trades), but the outer loop would have to be made explicitly async in the application space (which hasn't existed yet).


slundqui:
Lets discuss this when you get back. While the description says this is a purely refactor PR, there's changes made here that seem like more than just mechanical changes (e.g., changing functions to no longer be async).

dpaiton:
> Lets discuss this when you get back. While the description says this is a purely refactor PR, there's changes made here that seem like more than just mechanical changes (e.g., changing functions to no longer be async).

Ahh yes that's true -- I'm pretty sure it is fine bc the previous async function had an `await` but I agree we should discuss it'll be faster to go over it all.

src/agent0/core/hyperdrive/interactive/i_hyperdrive.py
```diff
@@ -205,13 +205,12 @@ def _init_agent(
...
     def _set_max_approval(self, agent: HyperdriveAgent):
         # Establish max approval for the hyperdrive contract
-        asyncio.run(
-            set_max_approval(
-                [agent],
-                self.interface.web3,
-                self.interface.base_token_contract,
-                str(self.interface.hyperdrive_contract.address),
-            )
+        set_max_approval(
```
slundqui, line 208: This is confusing me. On main, there are 2 spots where this function is being called: one in `get_agent_accounts.py`, where we pass in a list of agents, and one in `i_hyperdrive.py`, where we pass in a single agent. Making this function to only take a single agent makes me wonder what happened to the other call in this PR

slundqui: It looks like this was redefined in `setup_experiments` as a private function. So this PR is making a change where we were reusing one function to now using two separate functions.

src/agent0/core/hyperdrive/interactive/i_hyperdrive.py
```diff
@@ -245,7 +244,7 @@ def _open_long(self, agent: HyperdriveAgent, base: FixedPoint) -> OpenLong:
         agent.policy.set_next_action(HyperdriveActionType.OPEN_LONG, base)
         # TODO expose async here to the caller eventually
         trade_results: list[TradeResult] = asyncio.run(
-            async_execute_agent_trades(self.interface, [agent], liquidate=False, interactive_mode=True)
+            async_execute_agent_trades(self.interface, agent, liquidate=False, interactive_mode=True)
```
slundqui, line 247: Again, it looks like this function was split up into 3 different places, we have `async_execute_agent_trades`, `async_execute_multi_agent_trades`, and `async_execute_single_agent_trades`. I'd rather have one general purpose function that's being called everywhere instead of copies of functions being called in different places.

slundqui (approved):
Chatted in person about this PR, LGTM

PR #1366, opened 2024-03-18
Upgrade Agent0 to Hyperdrive 0.10.0

slundqui:
Also reworking slippage test to test all slippage trades. This was needed for testing the named return values change.


src/agent0/ethpy/hyperdrive/deploy.py
```diff
@@ -457,7 +460,12 @@ def _deploy_and_initialize_hyperdrive_pool(
         contribution=initial_liquidity.scaled_value,
         fixedAPR=initial_fixed_apr.scaled_value,
         timeStretchAPR=initial_time_stretch_apr.scaled_value,
-        initializeExtraData=bytes(0),
+        # initializeExtraData=bytes(0),
```
dpaiton, line 463: is this dead now? can you delete the comment?

src/agent0/core/hyperdrive/interactive/i_local_hyperdrive.py
```diff
@@ -162,8 +162,12 @@ class Config(IHyperdrive.Config):
         """The duration of a checkpoint (in seconds)."""
         curve_fee: FixedPoint = FixedPoint("0.01")  # 1%
         """The LP fee applied to the curve portion of a trade."""
-        flat_fee: FixedPoint = FixedPoint("0.0005")  # 0.05%
-        """The LP fee applied to the flat portion of a trade."""
+        # 0.05% APR. Here, we divide by 52 because the position duration is 1 week
+        # TODO do we want to default to 0.05% APR always? so we divide by position duration
```
dpaiton, line 166: I'm asking Alex questions to clarify; will report back. I am pretty sure we do not want the user to have to know to divide by position duration, so it should not be divided here in the config. Maybe in `__post_init__`

tests/bot_to_db_test.py
```diff
@@ -210,8 +211,12 @@ def test_bot_to_db(
             "time_stretch": expected_timestretch,
             "governance": deploy_account.address,
             "fee_collector": deploy_account.address,
+            # TODO current bug in solidity that returns zero address for sweep_collector
+            # Fix to look for deploy_account.address once this is fixed
+            # "sweep_collector": deploy_account.address,
+            "sweep_collector": ADDRESS_ZERO,
             "curve_fee": _to_unscaled_decimal(FixedPoint("0.01")),  # 1%
-            "flat_fee": _to_unscaled_decimal(FixedPoint("0.0005")),  # 0.05%
+            "flat_fee": _to_unscaled_decimal(FixedPoint("0.0005")),  # 0.05% APR
```
dpaiton, line 219: Are you sure this is "APR"? (same comment elsewhere that you made the change). The fee is assessed per transaction on the transaction amount, and therefore is inherently not related to time-passed. Our fee is wonky bc the actual amount is interpolated between flat fee and curve fee, according to the position duration, which gives it a "time feel". But nevertheless I think calling it an APR is still inaccurate.

src/agent0/core/hyperdrive/interactive/i_local_hyperdrive.py
slundqui: I'm torn on this. On one hand, I feel like it's good to reflect what's in solidity, but on the other, this fee in particular should perhaps be annualized so that the ux on this is sane. I do want to ensure that this matches the rest of our fees though

tests/bot_to_db_test.py
slundqui: This is indeed APR because the position duration in this configuration is 1 year

dpaiton (approved):

PR #1376, opened 2024-03-23
change fuzz_bots to use ihyperdrive

dpaiton:


src/agent0/core/hyperdrive/interactive/i_hyperdrive.py
```diff
@@ -148,7 +148,7 @@ def init_agent(
         private_key: str,
         policy: Type[HyperdriveBasePolicy] | None = None,
         policy_config: HyperdriveBasePolicy.Config | None = None,
-    ):
+    ) -> IHyperdriveAgent:
```
dpaiton, line 151: @slundqui I added this bc I needed it in an earlier iteration of the script, but I don't now. If there's some reason why you didn't do it in the first place lmk and I can delete it.

scripts/fuzz_bots.py
```diff
@@ -9,103 +9,105 @@
...
+    # The done trading state variable gets set internally
+    # TODO: Right now IHyperdrive always sets done_trading to False
+    # but there are situations where we want it to be true
+    if all(agent.done_trading for agent in agents):
+        break
+    # Execute the agent policies
+    for agent in agents:
+        try:
+            if not agent.done_trading:
+                trade_events = agent.execute_policy_action()
```
slundqui: `trade_events` isn't being used here, can you replace with `_`?

scripts/run_interactive_fuzz.sh
```diff
@@ -1,10 +1,10 @@
 # TODO Run these in the background, assuming these will handle how to notify or log errors
 
 echo "HYPERDRIVE BALANCE"
-python lib/agent0/agent0/interactive_fuzz/fuzz_hyperdrive_balance.py;
+python src/agent0/core/interactive_fuzz/fuzz_hyperdrive_balance.py;
```
slundqui, line 4: I actually don't think this shell script was ever being used, we have a python script that runs interactive fuzz tests

slundqui (approved):

scripts/run_interactive_fuzz.sh
dpaiton: ok I'll delete it

PR #1380, opened 2024-03-26
Removing IHyperdrivePolicy

slundqui:
Previously, IHyperdrive uses a dummy policy to set the next policy action, and execute trades through this dummy policy. Now, we call trades explicitly without this dummy policy, and the underlying policy for the agent is set directly instead of via sub policies.


src/agent0/core/hyperdrive/agent/hyperdrive_actions.py
```diff
@@ -167,8 +167,8 @@ def close_short_trade(
 
 def add_liquidity_trade(
     trade_amount: FixedPoint,
-    min_apr: FixedPoint = FixedPoint(scaled_value=1),
-    max_apr: FixedPoint = FixedPoint(scaled_value=2**256 - 1),
+    min_apr: FixedPoint | None = None,
```
dpaiton, line 170: is this necessary? we made FixedPoint immutable, right? so the `Optional` syntax is not needed because there is no risk of modifying the variable passed in?

src/agent0/core/hyperdrive/interactive/i_hyperdrive.py
```diff
@@ -184,15 +192,18 @@ def _init_agent(
...
+    ) -> HyperdriveAgent[HyperdriveBasePolicy]:
         # Setting the budget to 0 here, we'll update the wallet from the chain
-        agent = HyperdriveAgent(
-            Account().from_key(private_key),
-            initial_budget=FixedPoint(0),
-            policy=IHyperdrivePolicy(
-                IHyperdrivePolicy.Config(sub_policy=policy, sub_policy_config=policy_config, rng=self.config.rng)
-            ),
-        )
+        if policy is None:
```
dpaiton, line 197: we do have a NoOp base policy, which I think is a good default if none is provided.

src/agent0/core/hyperdrive/interactive/i_hyperdrive.py
```diff
@@ -239,91 +250,65 @@ def _add_funds(
...
 
     def _execute_policy_action(
         self, agent: HyperdriveAgent
     ) -> list[OpenLong | OpenShort | CloseLong | CloseShort | AddLiquidity | RemoveLiquidity | RedeemWithdrawalShares]:
-        assert isinstance(agent.policy, IHyperdrivePolicy)
         # Only allow executing agent policies if a policy was passed in the constructor
-        if agent.policy.sub_policy is None:
+        # we check type instead of isinstance to explicitly check for the hyperdrive base class
+        # pylint: disable=unidiomatic-typecheck
+        if type(agent.policy) == HyperdriveBasePolicy:
```
dpaiton, line 309: I think it would be a good idea to add a unit test for this to verify functionality

src/agent0/core/hyperdrive/agent/hyperdrive_actions.py
slundqui: No, this isn't necessary, but it's nice if the caller wants to pass in `None` for a default instead of either defining its own defaults or case switches for these optional values. It also matches how we're doing default values throughout the code

src/agent0/core/hyperdrive/interactive/i_hyperdrive.py
slundqui: There's too many generics down the chain that makes this very clunky to use the base class for this. Keeping this as is.

src/agent0/core/hyperdrive/agent/hyperdrive_actions.py
dpaiton: we discussed; all good now.

dpaiton (approved):

PR #1383, opened 2024-03-29
Making fuzz bots async

slundqui:


scripts/fuzz_bots.py
```diff
@@ -26,78 +28,139 @@
...
+                slippage_tolerance=SLIPPAGE_TOLERANCE,
+                trade_chance=FixedPoint("0.8"),
+                randomly_ignore_slippage_tolerance=True,
+                max_open_positions=2_000,
+            ),
+        )
+        agents.append(agent)
 
+    # Add funds asynchronously
+    print("Funding bots...")
```
dpaiton, line 137: log instead of print?

dpaiton (approved):

PR #1388, opened 2024-04-01
Allows calculating variable rate from checkpoints

slundqui:


src/agent0/ethpy/hyperdrive/interface/read_interface.py
```diff
@@ -273,7 +276,14 @@ def get_hyperdrive_state(self, block: BlockData | None = None) -> PoolState:
         checkpoint_time = self.calc_checkpoint_id(self.pool_config.checkpoint_duration, self.get_block_timestamp(block))
         checkpoint = get_hyperdrive_checkpoint(self.hyperdrive_contract, checkpoint_time)
         exposure = get_hyperdrive_checkpoint_exposure(self.hyperdrive_contract, checkpoint_time)
-        variable_rate = self.get_variable_rate(block_number)
+
+        try:
+            variable_rate = self.get_variable_rate(block_number)
+        except BadFunctionCallOutput:
+            # TODO do we want to throw a warning here?
```
dpaiton, line 283: yes I think a warning is a good call. Consider changing it to be "setting `state.variable_rate` to None".

src/agent0/ethpy/hyperdrive/state/pool_state.py
```diff
@@ -29,7 +29,9 @@ class PoolState:
     checkpoint_time: int
     checkpoint: CheckpointFP
     exposure: FixedPoint
-    variable_rate: FixedPoint
+    # TODO we may want to remove this from pool state and have users explicitly
```
dpaiton, line 32: I think I like this (removing it from state; forcing explicit call) better, but if you'd rather punt that for a later time it's good with me.

dpaiton (approved):

PR #1397, opened 2024-04-09
Parameterizing checkpoint bot

slundqui:


scripts/checkpoint_bot.py
```diff
@@ -100,7 +102,10 @@ def main(argv: Sequence[str] | None = None) -> None:
     else:
         block_timestamp_interval = int(block_timestamp_interval)
 
-    web3 = initialize_web3_with_http_provider(eth_config.rpc_uri, reset_provider=False)
+    if parsed_args.rpc_url == "":
+        web3 = initialize_web3_with_http_provider(eth_config.rpc_uri, reset_provider=False)
+    else:
+        web3 = initialize_web3_with_http_provider(parsed_args.rpc_url, reset_provider=False)
```
dpaiton, line 108: We should just stick to "URI" or change it to "URL" everywhere. Having both will be confusing.

dpaiton (approved):
lgtm, just one comment on the use of URL vs URI

PR #1402, opened 2024-04-10
add timeout

wakamex:


src/agent0/ethpy/base/transactions.py
```diff
@@ -505,6 +512,8 @@ async def async_smart_contract_transact(
         The number of times to retry the transact call if it fails. Defaults to no retries.
     txn_options_value: int | None
         The value field to set for transaction options.
+    timeout: int
```
dpaiton, line 515: use `int | None = None` and then set the default in the fn. Make sure the docstring names the default.

dpaiton (approved):

PR #1406, opened 2024-04-10
add gas limit parameter to trade functions

wakamex:


src/agent0/ethpy/base/transactions.py
```diff
@@ -419,21 +424,30 @@ def build_transaction(
...
     else:
-        transaction_kwargs = TxParams(
-            {
-                "from": signer_checksum_address,
-                "nonce": nonce,
-            }
-        )
+        transaction_kwargs["gas"] = txn_options_gas
+
+    transaction_kwargs["maxFeePerGas"] = Wei(max_fee_per_gas)
```
slundqui, line 449: Did we determine if these fields are needed?

src/agent0/core/hyperdrive/utilities/run_bots/execute_multi_agent_trades.py
```diff
@@ -237,7 +237,11 @@ async def _async_match_contract_call_to_trade(
 
         case HyperdriveActionType.OPEN_LONG:
             trade_result = await interface.async_open_long(
-                agent, trade.trade_amount, trade.slippage_tolerance, nonce=nonce
+                agent,
+                trade.trade_amount,
+                trade.slippage_tolerance,
+                nonce=nonce,
+                gas_limit=trade.gas_limit,
```
dpaiton, line 244: I think we intended for these to be flipped, so the order is `slippage, gas_limit, nonce`.

src/agent0/core/hyperdrive/utilities/run_bots/execute_multi_agent_trades.py
```diff
@@ -255,7 +259,12 @@ async def _async_match_contract_call_to_trade(
             if not trade.maturity_time:
                 raise ValueError("Maturity time was not provided, can't close long position.")
             trade_result = await interface.async_close_long(
-                agent, trade.trade_amount, trade.maturity_time, trade.slippage_tolerance, nonce=nonce
+                agent,
+                trade.trade_amount,
+                trade.maturity_time,
+                trade.slippage_tolerance,
+                nonce=nonce,
```
dpaiton, line 266: I think we intended for these to be flipped, so the order is `slippage, gas_limit, nonce`.

slundqui (approved):

src/agent0/ethpy/base/transactions.py
```diff
@@ -21,6 +22,7 @@
 
 DEFAULT_READ_RETRY_COUNT = 5
 DEFAULT_WRITE_RETRY_COUNT = 1
+GAS_ESTIMATE_MULTIPLIER = 1.1
```
dpaiton, line 25: we don't want this anymore, right?

src/agent0/ethpy/base/transactions.py
```diff
@@ -419,21 +424,30 @@ def build_transaction(
...
+    # Assign gas parameters
+    max_priority_fee = web3.eth.max_priority_fee
+    pending_block = web3.eth.get_block("pending")
+    base_fee = pending_block.get("baseFeePerGas", None)
+    if base_fee is None:
+        raise AssertionError("The latest block does not have a baseFeePerGas")
+    max_fee_per_gas = max_priority_fee + base_fee
+
+    if txn_options_gas is None:
+        transaction_kwargs["gas"] = int(web3.eth.estimate_gas(transaction_kwargs) * GAS_ESTIMATE_MULTIPLIER)
```
dpaiton, line 445: don't need this anymore

src/agent0/ethpy/base/transactions.py
```diff
@@ -8,6 +8,7 @@
 
 from eth_account.signers.local import LocalAccount
 from eth_typing import BlockNumber, ChecksumAddress
+from fixedpointmath import FixedPoint
```
dpaiton, line 11: don't need this anymore

src/agent0/ethpy/base/transactions.py
```diff
@@ -382,6 +384,7 @@ def build_transaction(
     nonce: Nonce | None = None,
     read_retry_count: int | None = None,
     txn_options_value: int | None = None,
+    txn_options_gas: int | None = None,
```
dpaiton, line 387: are we still using this? Is this the arg that the policy config is setting?

src/agent0/ethpy/base/transactions.py
wakamex: is it being used? is it better than nothing?

slundqui: I'm with Dylan, if we're adding an overwrite for gas, let's default to the rpc for the gas estimate without the multiplier

src/agent0/core/hyperdrive/utilities/run_bots/execute_multi_agent_trades.py
wakamex: moved nonce to the end and also added keywords in 521ea3e4c9f233cb8e22bdfc9db29042702e9980

src/agent0/core/hyperdrive/utilities/run_bots/execute_multi_agent_trades.py
wakamex: moved nonce to the end and also added keywords in 521ea3e4c9f233cb8e22bdfc9db29042702e9980

src/agent0/ethpy/base/transactions.py
wakamex: no they're not, they're just being more explicit about default behavior. Pretty sure I took them from somewhere else in our code.

src/agent0/ethpy/base/transactions.py
wakamex: removed in d21497c71bbc02a4271e9db1eb33b11e09655bc6

src/agent0/ethpy/base/transactions.py
wakamex: removed in 6c05738bd101660aa9d51c745e3837906f098736

src/agent0/ethpy/base/transactions.py
wakamex: added clarification that we're exposing default web3py behavior (other than the gas parameter) in 6c05738bd101660aa9d51c745e3837906f098736

src/agent0/ethpy/base/transactions.py
wakamex: yeah this is the lowest-level function that has that policy parameter passed to it, then we add it to transaction kwargs here

src/agent0/ethpy/base/transactions.py
wakamex: deleted the estimate_gas portion here because it breaks on devnet: 39580eb0cd3c07282add2957a6e5258cd4670683

PR #1407, opened 2024-04-11
fix typo in build positions

wakamex:


dpaiton (approved):
awesome, great catch!! I'll make an issue for writing a test for this

wakamex:
> awesome, great catch!! I'll make an issue for writing a test for this

yeah it just wasn't caught because we didn't have any cases of a transfer to a bot account. on testnet we have 1 transfer of Sepolia eth to these accounts.

PR #1413, opened 2024-04-16
add `calc_targeted_long` and `calc_spot_price_after_short`

dpaiton:
- add `calc_targeted_long` and `calc_spot_price_after_short` from rust
- rename some variables & reorders functions to more closely match rust
- reorganize & improve interface tests

### DIFF
--- a/src/agent0/ethpy/hyperdrive/interface/_mock_contract.py
+++ b/src/agent0/ethpy/hyperdrive/interface/_mock_contract.py
@@ -110,16 +110,16 @@ def _calc_max_long(pool_state: PoolState, budget: FixedPoint) -> FixedPoint:
     )
 
 
-def _calc_close_long(pool_state: PoolState, bond_amount: FixedPoint, maturity_time: int) -> FixedPoint:
+def _calc_close_long(
+    pool_state: PoolState, bond_amount: FixedPoint, maturity_time: int, current_time: int
+) -> FixedPoint:
     """See API for documentation."""
-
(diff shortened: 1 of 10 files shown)


src/agent0/ethpy/hyperdrive/interface/_mock_contract.py
```diff
@@ -46,13 +64,25 @@ def _calc_effective_share_reserves(pool_state: PoolState) -> FixedPoint:
...
-    spot_price = hyperdrivepy.calculate_spot_price(
-        fixedpoint_to_pool_config(pool_state.pool_config),
-        fixedpoint_to_pool_info(pool_state.pool_info),
+    if target_shares is None:
+        target_shares = _calc_effective_share_reserves(pool_state)
+    return FixedPoint(
+        scaled_value=int(
+            hyperdrivepy.calculate_initial_bond_reserves(
+                str(target_shares.scaled_value),
+                str(pool_state.pool_config.initial_vault_share_price.scaled_value),
```
wakamex, line 79: should we use current vault share price if we want this function to be used any time other than the beginning of the pool?

dpaiton: good idea; I made an issue for it so that it is fixed at the root level https://github.com/delvtech/hyperdrive/issues/987

at that time we can propagate the change up

wakamex (approved):
LGTM, we can create an issue to create calc_bond_reserves (not initial) later

PR #1433, opened 2024-04-24
add gas price multiples

wakamex:


src/agent0/ethpy/base/transactions.py
```diff
@@ -433,22 +445,20 @@ def build_transaction(
 
     # Assign gas parameters
     # other than the optional gas parameter, this is the default behavior of web3py, exposed here for clarity
-    max_priority_fee = web3.eth.max_priority_fee
+    max_priority_fee = int(web3.eth.max_priority_fee * txn_options_priority_fee_multiple)
     pending_block = web3.eth.get_block("pending")
     base_fee = pending_block.get("baseFeePerGas", None)
+    if base_fee is not None:
```
dpaiton, line 451: can't this be if/else?

src/agent0/ethpy/base/transactions.py
```diff
@@ -533,6 +545,10 @@ async def async_smart_contract_transact(
         The value field to set for transaction options.
     txn_options_gas : int | None
         The gas field to set for transaction options.
+    txn_options_base_fee_multiple: float | None = None
+        The multiple by which to multiply the base fee to set for transaction options. Defaults to 1.
```
dpaiton, line 549: this is a weird phrasing. maybe something  like "the base fee is multiplied by this before it is added to the max priority fee to set gas"?

src/agent0/ethpy/base/transactions.py
wakamex: I think it would be identical, since they're float | None.
Seems more explicit though?
```
if txn_options_value is not None:
if base_fee is not None:
if txn_options_gas is not None:
```
vs.
```
if txn_options_value:
if base_fee:
if txn_options_gas:
```
seems like the latter forces you to think a bit more to understand what's going on, whereas the former makes it clear you're checking for non-default values.
would the latter fail if the value is 0? since it also checks for truthiness?

src/agent0/ethpy/base/transactions_test.py
```diff
@@ -0,0 +1,132 @@
...
+# we're testing here
+# pylint: disable=too-many-locals
+
+
+@pytest.mark.anvil
+def test_gas_price_base_multiple_explicit(chain: ILocalChain):
+    """Set the gas price base multiple explicitly."""
+    # set up config, hyperdrive, interface, web3, and agent
+    base_fee_multiple = 100
+    regular_config = ILocalHyperdrive.Config()
```
dpaiton, line 21: why `regular_`? What are you differentiating from?

src/agent0/ethpy/base/transactions_test.py
```diff
@@ -0,0 +1,132 @@
...
+    multiplied_priority_fee_per_gas = multiplied_built_transaction["maxPriorityFeePerGas"]
+    assert "maxFeePerGas" in multiplied_built_transaction
+    assert isinstance(multiplied_built_transaction["maxFeePerGas"], int)
+    multiplied_base_fee_per_gas = multiplied_built_transaction["maxFeePerGas"] - multiplied_priority_fee_per_gas
+
+    assert multiplied_base_fee_per_gas / regular_base_fee_per_gas == base_fee_multiple
+
+
+@pytest.mark.anvil
+def test_gas_price_priority_multiple_explicit(chain: ILocalChain):
```
dpaiton, line 60: I'm not seeing the difference between this test and the one above it

src/agent0/ethpy/base/transactions.py
wakamex: improved in24d151d63fc5213669327255444ce5c09981bfa8

src/agent0/ethpy/base/transactions_test.py
wakamex: the one above it tests the base fee multiple, this tests the priority fee multiple

src/agent0/ethpy/base/transactions_test.py
wakamex: fixed in 680589b9819148a22dbd8efd774899623da1dbe6

src/agent0/ethpy/hyperdrive/interface/read_interface.py
```diff
@@ -109,6 +113,10 @@ def __init__(
             The number of times to retry the read call if it fails. Defaults to 5.
         txn_receipt_timeout: float | None, optional
             The timeout for waiting for a transaction receipt in seconds. Defaults to 120.
+        txn_options_base_fee_multiple: float | None, optional
```
dpaiton, line 116: I'm not following how you're using these new arguments. It seems like they are passed to the individual write functions in addition to the interface constructor. I think it's better to pass them to the write functions and not be member attributes of the interface object, but I could be missing something.

src/agent0/ethpy/base/transactions.py
dpaiton: sorry, I was talking  about

```py
if base_fee is not None:
  base_fee *= txn_options_base_fee_multiple
if base_fee is None:
  raise AssertionError("The latest block does not have a baseFeePerGas")
```

This should be `if ... else ...`

Regarding the other point you're making, I agree it's better to check the affirmative `is None` than the negative `is not None`. But `if base_fee` is **not** the same as `if base_fee is not None`. The latter would evaluate to `True` even if `base_fee = False`.

src/agent0/ethpy/base/transactions_test.py
dpaiton: Ahhh got it, thanks!

dpaiton (approved):

src/agent0/ethpy/base/transactions.py
wakamex: changed in 48e0212a272a59992b47b0e5495acb93d498b820

PR #1438, opened 2024-04-25
Normalizing test delta in lp share price check

slundqui:


dpaiton:
I think the math makes sense but it is more intuitive to me to normalize the expected value than the tolerance. what do you think?

dpaiton (approved):

PR #1439, opened 2024-04-26
add typing to BasePolicy.config

wakamex:
goal is to implement type narrowing all the way down to the lowest agent and policy level, through the use of generics


wakamex:
closed because #1443 needs a lot more work than this

PR #1441, opened 2024-04-26
Fixing race condition in invariance checks

slundqui:


dpaiton (approved):
great catch! we should always be using the pool_state object to avoid these race conditions.

PR #1442, opened 2024-04-26
improve freezable class

wakamex:
use Freezable as a class mixin instead of a decorator, allowing static typing of methods without stubbing

### DIFF
--- a/src/agent0/core/base/__init__.py
+++ b/src/agent0/core/base/__init__.py
@@ -1,4 +1,4 @@
-"""The agent0 base module exports some types here"""
+"""The agent0 base module exports some types here."""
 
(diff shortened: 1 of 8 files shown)


dpaiton:
Can you add a PR description? I'm not sure what the outcome of these changes should be.

dpaiton (approved):
Add a todo to remove the need to include type stubs for e.g. `def freeze(...` in any class that has the freezable decorator. Otherwise this is a great improvement!

wakamex:
no more type stubs necessary!

image

wakamex:
forgot you were going to review again @dpaiton even though I just requested a review 15 mins ago. we can revert, or add fixes on top.

PR #1465, opened 2024-05-11
Updating agent0 for Hyperdrive v1.0.5

slundqui:


src/agent0/chainsync/db/hyperdrive/interface_test.py
```diff
@@ -133,25 +133,44 @@ def test_checkpoint_time_query_checkpoints(self, db_session):
...
+        checkpoint_3 = CheckpointInfo(
+            checkpoint_time=100, vault_share_price=Decimal("3.4"), weighted_spot_price=Decimal("5.1")
+        )
         with pytest.raises(ValueError):
             add_checkpoint_info(checkpoint_3, db_session)
 
+        # Adding a checkpoint info with the same checkpoint time and vault share price should
+        # update the value
+        checkpoint_4 = CheckpointInfo(
+            checkpoint_time=100, vault_share_price=Decimal("3.1"), weighted_spot_price=Decimal("5.1")
```
sentilesdal, line 166: `vault_share_price` should never change for an existing checkpoint.  its kinda the whole point of the checkpoint.

src/agent0/core/test_utils/cycle_trade_policy.py
```diff
@@ -86,16 +86,19 @@ def action(
             # Remove All Liquidity
             action_list.append(remove_liquidity_trade(wallet.lp_tokens))
         elif self.counter == 4:
+            # Readd liquidity to allow for closing positions
```
sentilesdal, line 89: Readd

dpaiton (approved):

src/agent0/ethpy/hyperdrive/deploy.py
```diff
@@ -230,12 +235,32 @@ def _deploy_hyperdrive_factory(
         account=deploy_account_addr,
         constructorArgs=HyperdriveFactoryContract.ConstructorArgs(factory_deploy_config, "HyperdriveFactory"),
     )
+
+    lp_math_contract = LPMathContract.deploy(w3=web3, account=deploy_account_addr)
```
sentilesdal, line 239: I'm not sure of all the uses cases for this deploy method, but should we allowing passing in an existing lp_math_contract_address or some kinda external libraries obj?

src/agent0/ethpy/hyperdrive/deploy.py
```diff
@@ -422,7 +446,7 @@ def _deploy_and_initialize_hyperdrive_pool(
         )
 
     # There are 5 contracts to deploy, we call deployTarget on all of them
-    for target_index in range(5):
+    for target_index in range(4):
```
sentilesdal, line 449: comment 'There are 4 contracts to deploy', not 5

sentilesdal (approved):
One main concern with updating vault share price in the checkpoint

src/agent0/chainsync/db/hyperdrive/interface_test.py
slundqui: Agreed, and there's checks in the insertion logic that prevents this case from happening (see checkpoint_3 insertion). This update is for the new fields weighted spot price fields that do change when you get the checkpoint multiple times in the same checkpoint.

src/agent0/ethpy/hyperdrive/deploy.py
slundqui: Deploy only gets called when we deploy on a new chain. Agent0 doesn't support deploying on an existing chain, so no need to handle the case when this contract already exists.

src/agent0/chainsync/db/hyperdrive/interface_test.py
sentilesdal: ok nice

sentilesdal (approved):
ok looks good!

PR #1479, opened 2024-05-21
Adding explicit check for registered hyperdrive versions

slundqui:


src/agent0/ethpy/hyperdrive/addresses.py
```diff
@@ -135,6 +135,21 @@ def get_hyperdrive_addresses_from_registry(
...
             unnamed_addresses_dict.pop(hyperdrive_address)
 
+    # Check all versions of registered pools to ensure addresses are correct
+    for address, _ in unnamed_addresses_dict.items():
+        hyperdrive_contract: IHyperdriveContract = IHyperdriveContract.factory(w3=web3)(
+            web3.to_checksum_address(hyperdrive_address)
+        )
+        try:
+            hyperdrive_version = hyperdrive_contract.functions.version().call()
+            correct_version = "v1.0.6"
```
dpaiton, line 145: this should probably not be hardcoded here right? Is this the hyperdrive version we are synced with? If so, you could follow the pattern that we use in `hyperdrive-rs` to include a [`hyperdrive.version` file](https://github.com/delvtech/hyperdrive-rs/blob/main/crates/hyperdrive-wrappers/hyperdrive.version)

dpaiton (approved):
approved pending resolution of my comment

PR #1480, opened 2024-05-21
lp share price fuzz testing updates

slundqui:
- LP share price now is a separate run with the `--lp-share-price-test` flag that sets its own fuzz parameters.
  - If this flag is set, we only run the lp share price test with 0 - 10% variable rate and no fees.
  - If this flag isn't set, we skip the lp share price test

- Updating lp profitability invariant check with more epsilon

### DIFF
--- a/scripts/local_fuzz_bots.py
+++ b/scripts/local_fuzz_bots.py
@@ -118,6 +118,3 @@ def parse_arguments(argv: Sequence[str] | None = None) -> Args:
 
 if __name__ == "__main__":
     main()
-
-if __name__ == "__main__":
-    main()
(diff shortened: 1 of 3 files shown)


scripts/local_fuzz_bots.py
```diff
@@ -43,10 +58,66 @@ def main() -> None:
...
+    )
+
+    # Use system arguments if none were passed
+    if argv is None:
+        argv = sys.argv
+
+    return namespace_to_args(parser.parse_args())
+
+
+if __name__ == "__main__":
```
dpaiton, line 119: this is duplicated?

dpaiton (approved):
Approved. iirc we are doing this because the fee bounds were too high for this specific test, right? Can you add a note to that effect in the PR desc & also list the different config values? that way if stuff comes up later it'll be easier for protocol peeps to reference this PR without having to dig into the code.

PR #1482, opened 2024-05-22
add test for making sure examples work

dpaiton:
This sets up a framework for testing examples in CI.

We don't always want to run the example files unaltered. As a couple of examples, they may have components that we do not expect to work in CI (like running a dashboard) or they may execute trades in a loop that has a large number of iterations. To allow us to test the files while skipping certain lines, this test uses Python's `ast` to parse the data and throw out any unwanted lines.

The test itself is an adaptation of a [previous test](https://github.com/delvtech/agent0/blob/435f72310e0e6aa044a76adec5d1e61ec6e59327/tests/test_notebooks.py) that was used for example notebooks. The notebook portion could be reintegrated in order to test the `tutorial.ipynb` file.


slundqui (approved):
lgtm

PR #1485, opened 2024-05-22
Adding pause on invariance fail in fuzz bots script

slundqui:


scripts/local_fuzz_bots.py
```diff
@@ -114,6 +127,12 @@ def parse_arguments(argv: Sequence[str] | None = None) -> Args:
         action="store_true",
         help="Runs the lp share price fuzz with specific fee and rate parameters.",
     )
+    parser.add_argument(
+        "--pause-on-invariance-fail",
+        default=False,
+        action="store_true",
+        help="Runs the lp share price fuzz with specific fee and rate parameters.",
```
dpaiton, line 134: incorrect help

scripts/local_fuzz_bots.py
```diff
@@ -53,19 +55,28 @@ def main(argv: Sequence[str] | None = None) -> None:
...
+                log_to_rollbar=log_to_rollbar,
+                run_async=False,
+                random_advance_time=True,
+                random_variable_rate=True,
+                num_iterations=3000,
+                lp_share_price_test=parsed_args.lp_share_price_test,
+            )
+
+        except FuzzAssertionException as e:
+            print(f"Pausing pool on fuzz assertion exception {repr(e)}.")
```
dpaiton, line 78: we use `logging.info` in `testnet_fuzz_bot_invariant_checks`; should we use it here too?

dpaiton (approved):

PR #1487, opened 2024-05-23
Agent refactor

slundqui:
This PR is the 3rd PR in a series of PRs for agent0 to support multiple pools.

The agent classes in agent0 is reworked in this PR. Here, the main changes include `init_agent` being a function of the chain as opposed to the pool. Additionally, many functions now take a `pool` argument to specify what pool the agent is making a trade on. There are helper functions for setting an active pool to avoid passing this argument in various functions, by calling it either in `init_agent`, or by calling `agent.set_active(pool=...)`.

Additionally, the policy is now untied from the agent, such that an agent can hot-swap policies on the fly by calling the `agent.set_active(policy=...)` function.

- `init_agent` is now a function of a `LocalChain` or `Chain` object.
   - This function now takes an optional `pool` argument for setting an initial active pool.
- Many functions now require an active pool to be set, or an explicit pool to be passed in as an argument.
- `PolicyAgent` has been deprecated in favor of using `LocalAccount`.
- Moving many logging and bookkeeping configs from `Hyperdrive` and `LocalHyperdrive` to `Chain` and `LocalChain` respectively.
- No longer keeping track of wallet deltas, we get the wallet wrt a pool via `get_wallet()`.
- Adding helper functions for `get_long()`, `get_short()`, `get_lp()`, and `get_withdrawal_shares()`.
- `LocalHyperdrive` now has a flag for deploying on constructor, or attach to existing. This is necessary for the forking workflow.
- Added tests for unit and system fuzzing.


examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# Initialize agents from the chain
 # %%
-# Initialize the interactive object with specified initial pool parameters and the chain to launch hyperdrive on
-# An "admin" user (as provided by the Chain object) is launched/funded here for deploying hyperdrive
+# Generate funded trading agents from the interactive object
+# Names are reflected on output data frames and plots later
+# NOTE a local chain can initialize with base and eth.
+# Base is a mock token used for all underlying pools deployed.
+agent0 = chain.init_agent(
+    eth=FixedPoint(100),
```
dpaiton, line 30: I think it makes sense to give them a little base here to illustrate the point

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
-# NOTE: LocalChain is a subclass of Chain
 chain = LocalChain(local_chain_config)
-# Can connect to a specific existing chain
-# existing_chain = Chain("http://localhost:8545")
 
+# Initialize agents from the chain
 # %%
-# Initialize the interactive object with specified initial pool parameters and the chain to launch hyperdrive on
-# An "admin" user (as provided by the Chain object) is launched/funded here for deploying hyperdrive
+# Generate funded trading agents from the interactive object
```
dpaiton, line 25:
```
# Generate funded trading agents from the interactive object
# Names are reflected on output data frames and plots later
# Base is a mock token used for all underlying pools deployed.
#
# NOTE: a local chain can initialize with base and eth.
```

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# Alternatively, you can pass in one of the agents to use as the deployer. If you do this, ensure
+# there is enough base to cover the initial funding (defaults to 100_000 base.), and enough eth to cover gas.
 
 # Parameters for pool initialization. If empty, defaults to default values, allows for custom values if needed
-initial_pool_config = LocalHyperdrive.Config()
 # Launches 2 pools on the same local chain
-hyperdrive = LocalHyperdrive(chain, initial_pool_config)
-hyperdrive_2 = LocalHyperdrive(chain, initial_pool_config)
+
+# The config object is automatically build.
```
dpaiton, line 50: build -> built

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# TODO add argument for deployer
```
dpaiton, line 51: I think we should leave TODOs out of the examples. Better to make an issue or put the TODO deeper in the code.

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
-# Generate funded trading agents from the interactive object
-# Names are reflected on output data frames and plots later
-agent0 = hyperdrive.init_agent(base=FixedPoint(100000), eth=FixedPoint(100), name="alice")
-hyperdrive_agent1 = hyperdrive_2.init_agent(base=FixedPoint(100000), eth=FixedPoint(100), name="bob")
-# Omission of name defaults to wallet address
-hyperdrive_agent2 = hyperdrive.init_agent(base=FixedPoint(100000), eth=FixedPoint(10))
 
-# Add funds to an agent
-agent0.add_funds(base=FixedPoint(100000), eth=FixedPoint(100))
+# We also add an active pool to this agent as well.
```
dpaiton, line 58: remove "as well" because you already say "also"

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# We can initialize an agent with a custom policy - more on that later
+agent1 = chain.init_agent(
+    eth=FixedPoint(100),
+    name="agent1",
+    policy=PolicyZoo.random,
+    policy_config=PolicyZoo.random.Config(),
+)
+
+# %%
+# Initialize the interactive object with specified initial pool parameters and the chain to launch hyperdrive on.
```
dpaiton, line 42: This first sentence reads weird to me. Maybe:

`# Deploy a specified pool parameters on a specified chain, return an initialized LocalHyperdrive object.`

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+close_long_event_1 = agent0.close_long(
+    maturity_time=open_long_event_1.maturity_time,
+    bonds=open_long_event_1.bond_amount,
+    pool=hyperdrive0,
+)
+
+# Close previous longs from the wallet
+# This wallet is a wallet object that contains positions relevant to a specific pool.
+# This takes a snapshot of the current wallet positions.
+agent0.get_wallet(pool=hyperdrive0)
```
dpaiton, line 94: these `get_wallet` calls are going to create outputs that show the wallet info, right? If so, I think they should be their own cells so that the user can clearly see the different outputs for each agent. Then the closing events in its own cell.

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# Shorts
+open_short_event = agent1.open_short(bonds=FixedPoint(33333))
+agent1_shorts = agent1.get_shorts()
+close_short_event = agent1.close_short(maturity_time=agent1_shorts[0].maturity_time, bonds=agent1_shorts[0].balance)
+
+# LP
+add_lp_event = agent1.add_liquidity(base=FixedPoint(44444))
+remove_lp_event = agent1.remove_liquidity(shares=agent1.get_lp())
+
+# The above trades doesn't result in withdraw shares, but the function below allows you
```
dpaiton, line 110:
```
# The above trades don't result in withdraw shares, but if they did you could use the function below
# to withdraw shares from the pool.
# withdraw_shares_event = agent1.redeem_withdraw_share(shares=agent1.get_withdraw_shares())
```

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# to withdrawal shares from the pool.
+# withdraw_shares_event = agent1.redeem_withdraw_share(shares=agent1.get_withdraw_shares())
```
dpaiton, line 112: would be nice if a user can call this (or an equivalent close_long or close_short) event and it would provide an informative error "Agent has no withdrawal shares; nothing to redeem".  Then we can leave this line un-commented and also it seems like a common thing to do in a loop and ignore the error.

dpaiton: would be even better if we threw a custom `InsufficientBalanceError` that the user could catch!

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+
 # %%
 
-# Another long with a different maturity time
-open_long_event_2 = agent0.open_long(FixedPoint(22222))
+# Agents can also execute policies, which encapsulates actions to take on a pool.
+# This requires initializing a policy class from the agent.
+# For example, we set a policy that makes random trades.
+# We can either initialize a policy on initialization (see agent1's initialization)
```
dpaiton, line 119: `# We can either provide a policy on initialization (see agent1's initialization),`

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# or we can explicitly call `set_policy` to set a policy on an agent.
+# NOTE: `set_policy` overwrites the existing policy.
+# TODO we may be able to set multiple policies on an agent and hot-swap them
```
dpaiton, line 122: again no TODOs in examples

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+agent0.set_active(
+    policy=PolicyZoo.random,
+    policy_config=PolicyZoo.random.Config(rng_seed=123),
+)
+
+# Execute policy trade on a pool
+# Output event is one of the possible trade events
+agent0_trades = []
+for i in range(10):
+    # NOTE Since a policy can execute multiple trades per action, the output events is a list
```
dpaiton, line 132: `    # NOTE: Since a policy can execute multiple trades per action, the output events is a list.`

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# %% [markdown]
+#####################
+# Simulation Management
+#####################
 
+# %%
+# The local chain and hyperdrive objects provide various simulation tools, such as
+# advancing time, snapshotting, and controlling the underlying variable yield rate.
+
+# Advancing time.
```
dpaiton, line 149:
```
# Advancing time.
# The `advance_time` function accepts timedelta or seconds.
# The option `create_checkpoints` creates hyperdrive checkpoints,
# but this call may be slow when advancing a large amount of time.
# NOTE: these calls are chainwide calls, so all pools connected to this chain gets affected.
```

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+    agent1_trades.extend(agent1.execute_policy_action())
 
+# %% [markdown]
+#####################
+# Simulation Management
+#####################
 
+# %%
+# The local chain and hyperdrive objects provide various simulation tools, such as
+# advancing time, snapshotting, and controlling the underlying variable yield rate.
```
dpaiton, line 147: can you add an example for controlling the underlying variable yield rate? This is pool-specific, correct?

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# Non-blocking mode keeps the server running until cleanup.
+chain.run_dashboard(blocking=False)
 
 # %%
-# Get data from database under the hood
-pool_config = hyperdrive.get_pool_config()
-# The underlying data is in Decimal format, which is lossless. We don't care about precision
-# here, and pandas need a numerical float for plotting, so we coerce decimals to floats here
-pool_info = hyperdrive.get_pool_info(coerce_float=True)
+# These functions queries the underlying database to get data. All functions here return
```
dpaiton, line 178:
```
# These functions query the underlying database to get data.
# All functions here return a Pandas dataframe.
```

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# a pandas dataframe.
 
-checkpoint_info = hyperdrive.get_checkpoint_info()
+# Agent level analysis
 
-# Change this to get wallet
-wallet = agent0.get_wallet()
+# Get the raw trade events in pandas across all pools
```
dpaiton, line 183: You dropped using `.` at the end of your comments here. I personally think it's weird to require comments to be full sentences, but we do have that requirement in Solidity & Rust. More importantly, imo, is that we are consistent.

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
 
-pool_positions = hyperdrive.get_positions()
-trade_events = hyperdrive.get_trade_events()
-pool_positions_over_time = hyperdrive.get_historical_positions()
-total_wallet_pnl_over_time = hyperdrive.get_historical_pnl(coerce_float=True)
+# Get data from database wrt a hyperdrive pool.
+pool_config = hyperdrive0.get_pool_config()
+# The underlying data is in Decimal format, which is lossless. We don't care about precision
+# here, and pandas need a numerical float for plotting, so we coerce decimals to floats here
+# TODO this pool state contains pool_info + pool_analysis, which is different from the
```
dpaiton, line 201: no TODOs

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
+# %%
+# Pool level analysis
 
-pool_positions = hyperdrive.get_positions()
-trade_events = hyperdrive.get_trade_events()
-pool_positions_over_time = hyperdrive.get_historical_positions()
-total_wallet_pnl_over_time = hyperdrive.get_historical_pnl(coerce_float=True)
+# Get data from database wrt a hyperdrive pool.
+pool_config = hyperdrive0.get_pool_config()
+# The underlying data is in Decimal format, which is lossless. We don't care about precision
```
dpaiton, line 199:
```
# The underlying data is in Decimal format, which is lossless.
# Pandas need a numerical float for plotting, so we coerce decimals to floats.
```

examples/interactive_local_hyperdrive_example.py
dpaiton: (especially in the examples)

examples/interactive_local_hyperdrive_example.py
```diff
@@ -1,118 +1,228 @@
...
-# An "admin" user (as provided by the Chain object) is launched/funded here for deploying hyperdrive
+# Generate funded trading agents from the interactive object
+# Names are reflected on output data frames and plots later
+# NOTE a local chain can initialize with base and eth.
+# Base is a mock token used for all underlying pools deployed.
+agent0 = chain.init_agent(
+    eth=FixedPoint(100),
+    name="agent0",
+)
+# We can initialize an agent with a custom policy - more on that later
```
dpaiton, line 33: I dont think the custom policy discussion ever came up in this example; maybe just don't mention it here? or point them to the custom policy example file?

examples/interactive_remote_hyperdrive_example.py
```diff
@@ -33,63 +25,142 @@
...
-    private_key=private_key,
+private_key_0 = make_private_key()
+private_key_1 = make_private_key()
+
+# Init from private key
+agent0 = chain.init_agent(
+    private_key=private_key_0,
+    name="agent0",
+)
+# We can initialize an agent with a custom policy - more on that later
```
dpaiton, line 44: No custom policy discussion in this file like the local example. Probably makes sense to point them to the other example.

examples/interactive_remote_hyperdrive_example.py
```diff
@@ -1,30 +1,22 @@
...
-# Set the address of the hyperdrive pool
-# hyperdrive_address = "0x0000000000000000000000000000000000000000"
-
-# Alternatively, look up the list of registered hyperdrive pools
-# This is the registry address deployed on sepolia.
-# registry_address = "0xba5156E697d39a03EDA824C19f375383F6b759EA"
-#
-# hyperdrive_address = Hyperdrive.get_hyperdrive_addresses_from_registry(chain, registry_address)["sdai_14_day"]
-
 # For this example, we launch a chain and local hyperdrive, and set the rpc_uri and hyperdrive address from these.
```
dpaiton, line 20: I don't think `hyperdrive_address` or `rpc_uri` are defined in the code. But more importantly:

Will was confused about what is going on with this example vs the local one. Here are some changes I'd suggest:

- give the file a header that explains what it is demonstrating (for example, "This example demonstrates how to connect to a remote chain for analysis. It is a demo, however, so we will spoof a remote chain using similar steps to the local hyperdrive example.")
- put the local chain stuff (lines 21-22) and the pool stuff (lines 53-67) together
- add a little more description around the comments on lines 53-67

Here's an example:

```
# In a typical remote setup we would want to set the rpc_uri to a remote chain,
# such as Sepolia testnet:
# rpc_uri = "http://uri.to.sepolia.testnet"

# For this example, we instead launch a chain and local hyperdrive, and set the rpc_uri and hyperdrive address from these.
local_chain = LocalChain()
local_hyperdrive = LocalHyperdrive(local_chain)

# Now we can pretend the above is a remote chain and pool for our demonstration.

# NOTE: We need to set different db port here to avoid port collisions with local chain.
rpc_uri = local_chain.rpc_uri # I don't actually know the syntax for this
chain = Chain(rpc_uri, config=Chain.Config(db_port=1234))
hyperdrive_address = # I don't know what this should be, but it's missing as far as I can tell.

# There are two ways to connect to a remote pool:
#
# 1. directly provide a pool address:
# hyperdrive_address = "0x0000000000000000000000000000000000000000"
#
# 2. provide a registry address and select a pool from the list:
# hyperdrive_address = Hyperdrive.get_hyperdrive_addresses_from_registry(
#     chain,
#     registry_address="0xba5156E697d39a03EDA824C19f375383F6b759EA"
# )["sdai_14_day"]
#
# We will mimic #1 and use the previously deployed pool:
hyperdrive_config = Hyperdrive.Config()
hyperdrive_pool = Hyperdrive(chain, hyperdrive_address, hyperdrive_config)
```

examples/interactive_remote_hyperdrive_example.py
```diff
@@ -33,63 +25,142 @@
...
+# %% [markdown]
+#####################
+# Executing Trades
+#####################
+
+# We set agent1's active pool to avoid passing in pool for functions.
+agent1.set_active(pool=hyperdrive_pool)
 
+# Set max approval for the agent on a specific pool.
+# TODO this is overly permissive, we may want to expose setting approval for a specific
```
dpaiton, line 86: No TODOs in examples

examples/interactive_remote_hyperdrive_example.py
```diff
@@ -33,63 +25,142 @@
...
+
+# %%
 # Make trades
+
 # Return values here mirror the various events emitted from these contract calls
-# These functions are blocking, but relatively easy to expose async versions of the
-# trades below
-open_long_event = agent0.open_long(base=FixedPoint(11111))
-close_long_event = agent0.close_long(maturity_time=open_long_event.maturity_time, bonds=open_long_event.bond_amount)
+# TODO These functions are blocking, expose async versions of the trades below
```
dpaiton, line 96: no TODOs on examples

examples/interactive_remote_hyperdrive_example.py
```diff
@@ -33,63 +25,142 @@
...
+open_long_event = agent0.open_long(base=FixedPoint(11111), pool=hyperdrive_pool)
+close_long_event = agent0.close_long(
+    maturity_time=open_long_event.maturity_time,
+    bonds=open_long_event.bond_amount,
+    pool=hyperdrive_pool,
+)
+
+open_short_event = agent1.open_short(bonds=FixedPoint(33333))
```
dpaiton, line 104: We probably don't need lines 104-110, since it's redundant with the local example? I'm not sure how much we should repeat ourselves. Your call.

src/agent0/core/base/agent/policy_agent.py
```diff
@@ -1,107 +0,0 @@
...
-            address=HexBytes(self.address),
-            balance=Quantity(amount=initial_budget, unit=TokenType.BASE),
-        )
-
-    @property
-    def checksum_address(self) -> ChecksumAddress:
-        """Return the checksum address of the account."""
-        return Web3.to_checksum_address(self.address)
-
-    def get_trades(self, interface: MarketInterface) -> list[Trade[MarketAction]]:
```
dpaiton, line 92: We need a `preview_agent_trades` endpoint to complement `execute_agent_trades` that allows the user to see what the policy would suggest without actually doing it. This supports human-in-the-loop behavior.

src/agent0/core/hyperdrive/crash_report/crash_report.py
```diff
@@ -67,10 +71,14 @@ def build_crash_trade_result(
...
-        Object containing a wallet address and Agent for determining trades. If None, won't report the agent.
+    account: LocalAccount | None, optional.
+        The LocalAccount object that made the trade. If None, won't report the agent.
+    wallet: HyperdriveWallet | None, optional
+        The agent's wallet. If None, won't report the wallet.
+    policy: HyperdriveBasePolicy | None, optional
+        The agent's policy. If None, won't report the policy.
     trade_object: Trade[HyperdriveMarketAction] | None, optional
-        A trade provided by a HyperdrivePolicyAgent. If None, won't report the trade object.
+        A trade provided by a LocalAgent. If None, won't report the trade object.
```
dpaiton, line 81: `LocalAccount`

src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
```diff
@@ -144,85 +148,138 @@ def add_funds(
             The amount of base to fund the agent with. Defaults to 0.
         eth: FixedPoint | None, optional
             The amount of ETH to fund the agent with. Defaults to 0.
+        pool: LocalHyperdrive | None, optional
+            The pool to interact with. Defaults to the active pool.
         signer_account: LocalAccount | None, optional
             The signer account to use to call `mint`. Defaults to the agent itself.
         """
+
+        if pool is None and self._active_pool is not None:
```
dpaiton, line 157: I think there is a situation where both `pool` and `self._active_pool` are `None`. The user initializes with `pool=None` (allowed), and then calls `add_funds` with `pool=None` (allowed). But this combo would break, correct?

src/agent0/core/hyperdrive/interactive/local_chain.py
```diff
@@ -412,7 +431,56 @@ def load_snapshot(self) -> None:
...
-        pool_filename = self._snapshot_dir + "/" + self.chain_id + "-pools.pkl"
+        # NOTE: existing pool objects initialized after snapshot will no longer be valid.
+        self._load_pool_bookkeeping(self._snapshot_dir)
+        # load snapshot database state
+        self._load_db(self._snapshot_dir)
+        # Update pool's agent bookkeeping
+        # Note this will wipe the agent's active pool.
+        self._load_agent_bookkeeping(self._snapshot_dir)
+
+        # The hyperdrive interface in deployed pools need to wipe it's cache
```
dpaiton, line 442: `its`

dpaiton: or really "their" since we're talking about the plural "pools"

src/agent0/core/hyperdrive/interactive/local_chain.py
```diff
@@ -103,10 +121,11 @@ def __init__(self, config: Config | None = None, fork_uri: str | None = None, fo
 
         # Snapshot bookkeeping
         # TODO snapshot dir will be clobbered if you run multiple chains simultaneously
-        self._snapshot_dir = config.snapshot_dir
+        self._snapshot_dir = config.snapshot_dir + "/" + self.chain_id + "/"
```
dpaiton, line 124: path.sep or something right? to be os agnostic

dpaiton:
Note for followup. Left side is current dependency, right is what we want
image

src/agent0/core/hyperdrive/interactive/local_hyperdrive.py
```diff
@@ -340,7 +348,7 @@ def _create_checkpoint(
 
         try:
             tx_receipt = self.interface.create_checkpoint(
-                self._deployed_hyperdrive.deploy_account, checkpoint_time=checkpoint_time
+                self.chain.get_deployer_account(), checkpoint_time=checkpoint_time
```
dpaiton, line 351: This could potentially be a different deployer account per pool, and is fixed once the pool is deployed. So I think the chain should pass it as an argument on instantiation and then the pool can just own the address as a member variable.

src/agent0/core/hyperdrive/interactive/local_hyperdrive.py
```diff
@@ -193,12 +196,19 @@ def __init__(self, chain: LocalChain, config: Config | None = None, name: str |
...
+            _deployed_hyperdrive = self._deploy_hyperdrive(self.config, chain)
+            hyperdrive_address = _deployed_hyperdrive.hyperdrive_contract.address
+        else:
+            if hyperdrive_address is None:
+                raise ValueError("Must specify a hyperdrive address if not deploying a Hyperdrive contract.")
+
+        self._initialize(chain, hyperdrive_address, name)
 
-        self._initialize(chain, self._deployed_hyperdrive.hyperdrive_contract.address, name)
+        self.calc_pnl = self.chain.config.calc_pnl
```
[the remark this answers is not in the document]
dpaiton: or have `calc_pnl` member attribute in pool that is set via argument from chain, and config only lives in chain.

src/agent0/hyperfuzz/unit_fuzz/helpers/setup_fuzz.py
```diff
@@ -62,36 +62,38 @@ def setup_fuzz(
...
-        low=1, high=99999999
-    )  # No seed, we want this to be random every time it is executed
-    rng = np.random.default_rng(random_seed)
-
     # Parameters for pool initialization.
     # Using a day for checkpoint duration to speed things up
     if crash_log_level is None:
         crash_log_level = logging.CRITICAL
 
+    random_seed = np.random.randint(
```
dpaiton, line 70: move comment:
```
    # No seed, we want this to be random every time it is executed
    random_seed = np.random.randint(low=1, high=99999999)  
    rng = np.random.default_rng(random_seed)
 ```

dpaiton (approved):
Yay!

examples/interactive_local_hyperdrive_example.py
slundqui: Can't mint base here, this agent doesn't have an active pool

examples/interactive_local_hyperdrive_example.py
slundqui: This currently throws an "Can't call this function with an argument of 0"

examples/interactive_local_hyperdrive_example.py
slundqui: It's there a couple of lines below

examples/interactive_local_hyperdrive_example.py
slundqui: Meant to say "active policy" here

examples/interactive_remote_hyperdrive_example.py
slundqui: Again, I meant active policy here

examples/interactive_remote_hyperdrive_example.py
slundqui: These analysis functions are a subset of the supported ones in local, so I wanted to give examples of the allowed queries in the remote case

examples/interactive_local_hyperdrive_example.py
dpaiton: oh, well then maybe the answer is to move that comment to elsewhere? I read it to say that I can initialize with base and eth but it sounds like you're saying I can't in this calse

src/agent0/core/base/agent/policy_agent.py
slundqui: Will implement in a future pr.
https://github.com/delvtech/agent0/issues/1491

examples/interactive_local_hyperdrive_example.py
dpaiton: hmm ok yeah I guess that's fine for now

src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
slundqui: This gets checked when we attempt to mint base

src/agent0/core/hyperdrive/interactive/local_hyperdrive.py
slundqui: Will solve in future PR:
https://github.com/delvtech/agent0/issues/1492

src/agent0/core/hyperdrive/interactive/local_hyperdrive.py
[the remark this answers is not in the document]
slundqui: https://github.com/delvtech/agent0/issues/1492

examples/interactive_local_hyperdrive_example.py
slundqui: Yeah, I moved the comment

PR #1498, opened 2024-05-29
Exposing gas limit to chain config and using the parameter for create checkpoint in advance time

slundqui:


dpaiton (approved):
I made an issue to remind us what we ultimately want (fewer parameters listed in units of `base`) https://github.com/delvtech/agent0/issues/1500 but this is fine for now to get stuff up & running.

PR #1507, opened 2024-05-31
Deployment cleanup

slundqui:
PR to clean up hyperdrive deployment.

- Splitting up factory deployment and hyperdrive deployment from factory.
- Factory deployment also deploys a registry.
- Cleaning up named tuple outputs for deploy functions

### DIFF
--- a/src/agent0/core/hyperdrive/interactive/local_hyperdrive_test.py
+++ b/src/agent0/core/hyperdrive/interactive/local_hyperdrive_test.py
@@ -1047,12 +1047,13 @@ def test_share_price_compounding_annus(fast_chain_fixture: LocalChain):
     )
     interactive_hyperdrive = LocalHyperdrive(fast_chain_fixture, interactive_config)
     hyperdrive_interface = interactive_hyperdrive.interface
+    beginning_share_price = hyperdrive_interface.current_pool_state.pool_info.lp_share_price
     logging.info(f"Variable rate: {hyperdrive_interface.current_pool_state.variable_rate}")
     logging.info(f"Starting share price: {hyperdrive_interface.current_pool_state.pool_info.lp_share_price}")
     fast_chain_fixture.advance_time(YEAR_IN_SECONDS, create_checkpoints=False)
     ending_share_price = hyperdrive_interface.current_pool_state.pool_info.lp_share_price
     logging.info(f"Ending   share price: {ending_share_price}")
-    assert ending_share_price - 1 == initial_variable_rate
+    assert ending_share_price - beginning_share_price == initial_variable_rate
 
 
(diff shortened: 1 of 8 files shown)


src/agent0/ethpy/hyperdrive/deploy.py
```diff
@@ -47,123 +48,171 @@
...
+    function_name = register_function.fn_name
+    function_args = register_function.args
+    receipt = smart_contract_transact(
+        web3,
+        deployed_factory.registry_contract,
+        deployer_account,
+        function_name,
+        *function_args,
+    )
+    assert receipt["status"] == 1, f"Failed to register Hyperdrive contract.\n{receipt=}"
```
dpaiton, line 212: is this right to be an assert? seems like a runtime check?

dpaiton (approved):

src/agent0/core/hyperdrive/interactive/local_hyperdrive_test.py
```diff
@@ -1047,12 +1047,13 @@ def test_share_price_compounding_annus(fast_chain_fixture: LocalChain):
...
     hyperdrive_interface = interactive_hyperdrive.interface
+    beginning_share_price = hyperdrive_interface.current_pool_state.pool_info.lp_share_price
     logging.info(f"Variable rate: {hyperdrive_interface.current_pool_state.variable_rate}")
-    logging.info(f"Starting share price: {hyperdrive_interface.current_pool_state.pool_info.lp_share_price}")
+    logging.info(f"Starting share price: {beginning_share_price}")
     fast_chain_fixture.advance_time(YEAR_IN_SECONDS, create_checkpoints=False)
     ending_share_price = hyperdrive_interface.current_pool_state.pool_info.lp_share_price
     logging.info(f"Ending   share price: {ending_share_price}")
-    assert ending_share_price - 1 == initial_variable_rate
+    assert ending_share_price - beginning_share_price == initial_variable_rate
```
slundqui, line 1056: @wakamex Can you look over this test and make sure this is still okay? This PR added a transaction between deploying the pool and returning (to register the deployed pool), so I think we can no longer assume the initial share price is 1

wakamex: lgtm if the test passes, but i'd think if starting point isn't 1 we should use

```
(ending_share_price - beginning_share_price) / beginning_share_price = initial_variable_rate
```

wakamex: A transaction actually has to be made for compounding to take place, so it's correct for the rate of return to be applied to the starting share price of 1.

PR #1513, opened 2024-06-04
Adding in agent0 signature to extra data on trade calls

slundqui:


dpaiton:
I think we should make a pool (or chain?) config variable that sets the default of `bytes.fromhex("a0")` for the extra data. Then a user can override to empty if they don't want it.

jrhea (approved):

PR #1524, opened 2024-06-09
Hyperdrive v1.0.10

dpaiton:


src/agent0/ethpy/hyperdrive/addresses.py
```diff
@@ -94,7 +94,7 @@ def get_hyperdrive_addresses_from_registry(
         address=web3.to_checksum_address(hyperdrive_registry_addr)
     )
     # Look for events from the contract
-    events = registry_contract.events.HyperdriveInfoUpdated.get_logs(fromBlock=0)
+    events = registry_contract.events.InstanceInfoUpdated.get_logs(fromBlock=0)
```
slundqui, line 97: We need to not use events here, we lose this event in the infra deployment

src/agent0/hyperdrive.version
```diff
@@ -1,3 +1 @@
-v1.0.6
-v1.0.7
-v1.0.8
\ No newline at end of file
+v1.0.10
```
slundqui, line 1: We still need to support v1.0.6 since that's the version on sepolia

slundqui:
Let's propagate the new event field throughout

slundqui (approved):

PR #1530, opened 2024-06-12
Allow connection of service bots to remote chain. Adding rollbar logging to checkpoint bot. Includes version bump to v0.20.3

slundqui:


dpaiton (approved):
should probably try to keep version bumps to their own pr as a rule of thumb!

PR #1532, opened 2024-06-12
Updates to FullHyperdriveEnv in traiderdaive to fix stale references

wshainin:
Updates keys to reflect new schema changes.
Updates observation step to handle an empty bot wallet on first iteration.
Makes explicit which columns from pool_info to use in observation space.
Changes reward calculation to change in pnl from t-1 -> t (for now).
Remove method that filters for rl bot's wallet from all positions in favor of DB query.

### DIFF
--- a/src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
+++ b/src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 import warnings
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any
 
@@ -15,8 +15,7 @@ from fixedpointmath import FixedPoint
 from gymnasium import spaces
 from scipy.special import expit
 
-from agent0.core.hyperdrive.interactive import LocalChain, LocalHyperdrive
-from agent0.core.hyperdrive.policies import PolicyZoo
+from agent0 import LocalChain, LocalHyperdrive, PolicyZoo
 
 # Global suppression of warnings, TODO fix
 warnings.filterwarnings("ignore")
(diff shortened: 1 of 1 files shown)


src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
```diff
@@ -465,6 +493,7 @@ def _get_info(self) -> dict:
         return {}
 
     def _get_rl_wallet_positions(self, coerce_float: bool) -> pd.DataFrame:
+        # TODO can we use self.rl_bot.get_positions()?
```
dpaiton, line 496: I think yes

src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
```diff
@@ -510,35 +515,59 @@ def _get_observation(self) -> dict[str, np.ndarray]:
...
+            current_block = self.interactive_hyperdrive.interface.get_current_block()
+            timestamp = self.interactive_hyperdrive.interface.get_block_timestamp(current_block)
+            rl_bot_wallet["normalized_time_remaining"] = (
+                rl_bot_wallet["maturity_time"] - timestamp
+            ) / position_duration
+
+            long_orders = rl_bot_wallet[rl_bot_wallet["token_id"] == "LONG"]
+            # Ensure data is the same as the action space
+            long_orders = long_orders.sort_values("maturity_time")
+            # long_orders["pnl"] = long_orders["unrealized_"]
```
dpaiton, line 534: dead comment

dpaiton (approved):
lgtm; thanks for getting this up to speed!

src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
slundqui: Yes, it's the same underlying function, just where you want to put the computation. `self.rl_bot.get_positions()` filters on the db side, whereas the code here puts the computation on the pandas side

slundqui (approved):
This looks great, thanks for fixing the breaking changes from the refactor

src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
wshainin: Which do we prefer? If we want it on the DB side, we could get rid of the method `_get_rl_wallet_positions`, and just replace the three calls with `self.rl_bot.get_positions`

src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
wshainin: Fixed

src/agent0/traiderdaive/gym_environments/full_hyperdrive_env.py
dpaiton: I think you should keep this on the db side and remove the `_get_rl_wallet_positions` method

wshainin: Ok, done.

PR #1533, opened 2024-06-17
pin numpy to 1.26.4

wakamex:
otherwise, uv resolves to numpy 2.0 which forces web3 to install a really old version


wakamex:
fixed by https://github.com/delvtech/agent0/pull/1535

PR #1536, opened 2024-06-17
lpandarb: handle matured positions

wakamex:


src/agent0/core/hyperdrive/policies/lpandarb.py
```diff
@@ -97,22 +97,17 @@ def arb_fixed_rate_down(
...
-        reduce_short_amount = minimum(
-            short.balance, bonds_needed / curve_portion, interface.calc_max_long(max_trade_amount_base, pool_state)
-        )
-        if reduce_short_amount > min_trade_amount_bonds:
-            bonds_needed -= reduce_short_amount * curve_portion
-            logging.debug(
-                "reducing short by %s\nreduce_short_amount*curve_portion = %s",
-                reduce_short_amount,
-                reduce_short_amount * curve_portion,
+        if curve_portion > FixedPoint(0):
```
dpaiton, line 100: Is the motivation behind this flag that it is costly & time-consuming to close matured positions, which is not relevant if the bot has unlimited funds?

If so, I think it would be smart to put this behind a flag. For example, `if curve_portion > FixedPoint(0) or self.config.close_matured_positions`. I assume we want to close matured positions for a bot with fixed (limited) budget.

dpaiton (approved):
Consider putting this behind a bot config flag. Otherwise I think it's a good change. Thanks for writing the tests too!

src/agent0/core/hyperdrive/policies/lpandarb.py
wakamex: I briefly debated having it close all matured positions before doing anything else. I think a previous version actually did that.

Maybe it should close them? I didn't want them to get in the way. But they don't have any market impact. A real user would almost certainly want to close their positions.

wakamex: even in the unlimited money case, closing matured positions doesn't hurt at all. it's just an extra step. but since we want to close matured positions in the real world, let's add it as default behavior.

PR #1539, opened 2024-06-18
Fuzz fixes

slundqui:
- Re-adding `status == 0` error message when no trace exists (this allows fuzz bots to not pause on this error)
- Relaxing lp share price check if a checkpoint was minted on the checked block


wakamex (approved):
lgtm

dpaiton (approved):

PR #1561, opened 2024-06-21
Allow for lazy pnl calculation

slundqui:
This PR allows for only calculating pnl when calling `get_positions` for speed. To utilize this, set `calc_pnl = False` in the `Chain.Config`. If this configuration is set to False, `get_positions` accepts a `calc_pnl` argument, which fills the pnl values in the resulting dataframe.

Includes a bugfix to `get_positions` when getting positions from multiple pools.

### DIFF
--- a/src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
+++ b/src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
@@ -980,8 +980,10 @@ class HyperdriveAgent:
             Whether to show positions closed positions (i.e., positions with zero balance). Defaults to False.
             When False, will only return currently open positions. Useful for gathering currently open positions.
             When True, will also return any closed positions. Useful for calculating overall pnl of all positions.
-        calc_pnl: bool
-            Whether to return the pnl for the current position. Only used if the chain config's `calc_pnl` is False.
+        calc_pnl: bool, optional
+            If the chain config's `calc_pnl` flag is False, passing in `calc_pnl=True` to this function allows for
+            a one-off pnl calculation for the current positions. Ignored if the chain's `calc_pnl` flag is set to True,
+            as every position snapshot will return pnl information.
         coerce_float: bool, optional
             Whether to coerce underlying Decimal values to float when as_df is True. Defaults to False.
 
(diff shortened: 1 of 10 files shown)


src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
```diff
@@ -979,6 +980,8 @@ def get_positions(
             Whether to show positions closed positions (i.e., positions with zero balance). Defaults to False.
             When False, will only return currently open positions. Useful for gathering currently open positions.
             When True, will also return any closed positions. Useful for calculating overall pnl of all positions.
+        calc_pnl: bool
+            Whether to return the pnl for the current position. Only used if the chain config's `calc_pnl` is False.
```
dpaiton, line 984: I think it would be a good idea to add a conditional right off the bat and warn the user if they're setting this flag without setting chain config's `calc_pnl`. e.g.:

```py
if self.chain.config.calc_pnl and calc_pnl:
    logging.warning("calc_pnl parameter is set but will not be used because chain.calc_pnl is False.")
```

src/agent0/core/hyperdrive/interactive/local_hyperdrive.py
```diff
@@ -497,6 +503,8 @@ def get_positions(self, show_closed_positions: bool = False, coerce_float: bool
             When False, will only return currently open positions. Useful for gathering currently open positions.
             When True, will also return any closed positions. Useful for calculating overall pnl of all positions.
             Defaults to False.
+        calc_pnl: bool
+            Whether to return the pnl for the current position. Only used if the chain config's `calc_pnl` is False.
```
dpaiton, line 507: same as above with warning message

src/agent0/core/hyperdrive/interactive/local_hyperdrive_agent.py
```diff
@@ -534,6 +535,8 @@ def get_positions(
             Whether to show positions closed positions (i.e., positions with zero balance). Defaults to False.
             When False, will only return currently open positions. Useful for gathering currently open positions.
             When True, will also return any closed positions. Useful for calculating overall pnl of all positions.
+        calc_pnl: bool
+            Whether to return the pnl for the current position. Only used if the chain config's `calc_pnl` is False.
```
dpaiton, line 539: I've lost track of the flow here but I think you'd want a warning message here as well? Maybe just here and not the other ones?

dpaiton (approved):

PR #1564, opened 2024-06-24
Allows backfilling pool info in the case of forking a pool

slundqui:
Includes cool progress bar when doing slow backfilling.
image

Also includes the following minor fixes:
- Removing unnecessary uv installation in CI.
- Fixing a bug with `test_lazy_calc_pnl` dataframe comparison where pandas `equals` was comparing based on index after sort.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,9 +21,6 @@ jobs:
           python-version: "3.10"
           token: ${{github.token}}
 
-      - name: install uv
-        run: curl -LsSf https://astral.sh/uv/install.sh | sh
-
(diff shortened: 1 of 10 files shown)


.github/workflows/lint.yml
```diff
@@ -21,9 +21,6 @@ jobs:
           python-version: "3.10"
           token: ${{github.token}}
 
-      - name: install uv
-        run: curl -LsSf https://astral.sh/uv/install.sh | sh
```
dpaiton, line 25: oops good catch

dpaiton (approved):

PR #1582, opened 2024-06-28
Various fuzz fixes

slundqui:
- Checkpoint bot logs warnings when retrying, and logs criticals if it fails 10 times in a row.
- Invariant check script checks every block.
- Running invariance checks asynchronously.
- Each invariance check controls what log level to log at in rollbar.
- Changing `check_checkpointing_should_never_fail` to `check_previous_checkpoint_exists`, which ensures the previous checkpoint always exists.
- LP share price check throws warning if price is up past threshold, throws critical if price is down past threshold.
- Removing initial lp profitable check.

### DIFF
--- a/scripts/checkpoint_bots.py
+++ b/scripts/checkpoint_bots.py
@@ -244,7 +244,7 @@ def run_checkpoint_bot(
                 fail_count = 0
             except Exception as e:  # pylint: disable=broad-except
                 if fail_count < FAIL_COUNT_THRESHOLD:
-                    logging_str = f"Checkpoint transaction failed."
+                    logging_str = "Checkpoint transaction failed."
                     log_level = logging.WARNING
                     logging.warning(logging_str + f" {repr(e)}")
                 else:
--- a/scripts/checkpoint_bots.py
+++ b/scripts/checkpoint_bots.py
@@ -246,11 +246,10 @@ def run_checkpoint_bot(
                 if fail_count < FAIL_COUNT_THRESHOLD:
                     logging_str = "Checkpoint transaction failed."
                     log_level = logging.WARNING
-                    term_logging_str = logging_str + f" {repr(e)}"
                 else:
                     logging_str = f"Checkpoint transaction failed over {FAIL_COUNT_THRESHOLD} times."
                     log_level = logging.CRITICAL
-                    term_logging_str = logging_str + f" {repr(e)}"
+                term_logging_str = logging_str + f" {repr(e)}"
                 logging.log(log_level, term_logging_str)
 
                 if log_to_rollbar:
--- a/scripts/checkpoint_bots.py
+++ b/scripts/checkpoint_bots.py
@@ -39,6 +39,7 @@ CHECKPOINT_WAITING_PERIOD = 0.5
 
 # The threshold for warning low funds
 CHECKPOINT_BOT_LOW_ETH_THRESHOLD = FixedPoint(0.1)
+FAIL_COUNT_THRESHOLD = 10
 
 
 # Sets up async nonce manager
@@ -155,6 +156,8 @@ def run_checkpoint_bot(
     # TODO ERC4626 gets split up here, may want to only do this for the position duration string.
     pool_name = "".join([c + "_" if c.isdigit() else c for c in pool_name])
 
+    fail_count = 0
+
     while True:
         # Check if we've reached the block to exit
         if block_to_exit is not None and chain.block_number() >= block_to_exit:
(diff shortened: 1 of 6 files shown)


scripts/checkpoint_bots.py
```diff
@@ -237,17 +240,27 @@ def run_checkpoint_bot(
...
-                )
-                logging.warning(logging_str)
+                if fail_count < FAIL_COUNT_THRESHOLD:
+                    logging_str = "Checkpoint transaction failed."
+                    log_level = logging.WARNING
+                    term_logging_str = logging_str + f" {repr(e)}"
+                else:
+                    logging_str = f"Checkpoint transaction failed over {FAIL_COUNT_THRESHOLD} times."
+                    log_level = logging.CRITICAL
+                    term_logging_str = logging_str + f" {repr(e)}"
```
dpaiton, line 253: looks like this line `term_logging_str = logging_str + f" {repr(e)}"` is only needed once after the conditional.

scripts/invariant_checks.py
```diff
@@ -67,11 +72,14 @@ def main(argv: Sequence[str] | None = None) -> None:
     # Get deployed pools on first iteration
     deployed_pools = Hyperdrive.get_hyperdrive_pools_from_registry(chain, registry_address)
 
+    start_block_number = chain.block_data().get("number", None)
```
dpaiton, line 75: Is there a reason you use `chain.block_data().get("number", None)` here and `chain.block_number()` below? I looked and they seem equivalent to me, maybe the web3 call under the hood for `chain.block_number()`, which uses `RPC.eth_blockNumber` ends up making two http requests? If there's no difference, I like `chain.block_number()` as the standard only because it's more readable.

scripts/invariant_checks.py
```diff
@@ -81,30 +89,45 @@ def main(argv: Sequence[str] | None = None) -> None:
...
+                    log_to_rollbar=log_to_rollbar,
+                    pool_name=hyperdrive_obj.name,
+                )
+                for hyperdrive_obj in deployed_pools
+            ]
+
+            logging.info("Running invariant checks for block %s on pools %s", check_block, deployed_pools)
+            asyncio.run(async_runner(return_exceptions=False, funcs=partials))
+
+        start_block_number = latest_block_number + 1
```
dpaiton, line 125: can you rename this? like uhh... `last_checked_block` or something? I think the `if (latest_block_number - start_block_number) > LOOKBACK_BLOCK_LIMIT:` conditional makes it seem like this only runs for `LOOKBACK_BLOCK_LIMIT` iterations.

scripts/invariant_checks.py
```diff
@@ -81,30 +89,45 @@ def main(argv: Sequence[str] | None = None) -> None:
...
-                log_to_rollbar=log_to_rollbar,
-                pool_name=name,
-            )
+
+        for check_block in range(start_block_number, latest_block_number + 1):
+            if (latest_block_number - start_block_number) > LOOKBACK_BLOCK_LIMIT:
+                error_message = "Unable to keep up with invariant checks."
+                logging.error(error_message)
+                log_rollbar_message(error_message, logging.ERROR)
+                break
```
dpaiton, line 108: I think we should log the error and `continue` instead of `break`. My thinking is if it falls behind that's bad (post an error) but that doesn't mean we should stop checking for future blocks.

src/agent0/ethpy/hyperdrive/interface/read_interface.py
```diff
@@ -209,6 +209,16 @@ def __init__(
         self._current_pool_state = None
         self.last_state_block_number = -1
 
+        # Best effort to find initialize event and set deploy block
+        self.deploy_block: None | int = None
+        initialize_event = self.get_initialize_events("earliest")
```
dpaiton, line 214: Are we sure we want this every time the interface is initialized? If I understand correctly this will pull all logs from the earliest available block. That has to be really time consuming, right? I think we should populate this when requested (`interface.get_deploy_block()`) and make sure to make that call in the invariance checks.

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -100,21 +113,30 @@ def run_invariant_checks(
...
             exception_message.append(message)
+        exception_data = exception_data_template.copy()
         exception_data.update(data)
         exception_data["block_number"] = pool_state.block_number
 
-    # Log additional information if any test failed
-    if any_check_failed:
-        logging.critical("\n".join(exception_message))
+        # Log exception to rollbar
+        assert log_level is not None
```
dpaiton, line 136: is this for type narrowing or is this possible at run time? if run time make it an `if` --> `raise` conditional.

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -395,56 +437,32 @@ def _check_lp_share_price(
...
+    # We check if lp share price jumps higher than expected
+    # if we're doing a full check. In this case, we warn
+    difference_in_wei = abs(previous_lp_share_price.scaled_value - current_lp_share_price.scaled_value)
     if not currently_minting_checkpoint and not closing_mature_position:
-        if not isclose(previous_lp_share_price, current_lp_share_price, abs_tol=test_tolerance):
-            failed = True
-    # Relaxed check
-    else:
-        if (previous_lp_share_price - current_lp_share_price) >= test_tolerance:
+        if (current_lp_share_price - previous_lp_share_price) >= test_tolerance:
```
dpaiton, line 444: should this also be `abs`? like on line 442

dpaiton (approved):
This is great, thanks for turning around these TODOs so quickly. I think we should talk about testing this out on a mainnet fork w/ fuzz bots. We should do things like underfund the checkpoint bot and make sure discord pings the critical error. We could also spin up anvil w/ a known insolvent pool state and run the checkpoints to verify errors show up.

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
slundqui: This is type narrowing, log_level should always be set in this case

scripts/invariant_checks.py
slundqui: Moved this check outside of the loop for clarity

PR #1603, opened 2024-07-11
Fixing lp share price check to check against pending block vs the same block once it's mined

slundqui:


src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -390,46 +388,59 @@ def _check_lp_share_price(
     exception_data: dict[str, Any] = {}
     log_level = None
 
-    block_number = pool_state.block_number
+    # TODO we hack in a stateful variable into the interface here, since we need
```
dpaiton, line 391: you could add the variable as a function arg and include it in the InvariantCheckResults, so that it maintains state across subsequent calls

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -390,46 +388,59 @@ def _check_lp_share_price(
...
 
-    # This is known to fail when checking the first block, as block - 1 doesn't exist.
-    try:
-        previous_pool_state = interface.get_hyperdrive_state(block_number - 1)
-    except BlockNotFound:
-        # Not a failure
-        return InvariantCheckResults(
-            failed=False, exception_message=exception_message, exception_data=exception_data, log_level=log_level
-        )
+    # The mined pool state is always 12 seconds ahead of the previous pool state.
```
dpaiton: is it always 12 sec when we're fuzzing with a `--block-time=12` flag set for anvil, right? But on mainnet & testnet it's not always 12. Is that going to matter?

dpaiton (approved):

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
slundqui: Resolved via flag when running in simulation mode

dpaiton (approved):

PR #1607, opened 2024-07-13
Various fixes and features for rl bots training

slundqui:
This PR adds various changes to support rl bots training. Some of these changes helps take a step in allowing async trades in agent0, such that multiple transactions can be submitted on the same block.

# Breaking changes
- Renaming `agent.liquidate` to `agent.execute_liquidate`.

# Changes
- Splitting up getting an agent's actions (i.e., `get_policy_action` and `get_liquidate_action`) and executing an agent's actions (i.e., `execute_action`) into two separate steps, with `execute_policy_action` and `execute_liquidate` calling the two steps sequentially.
- Adding `manual_database_sync` to `LocalChain.Config` to allow for manual database syncing instead of syncing after every trade.
- Executing transactions (e.g., `agent.open_long`) no longer eagerly gets the wallet, and instead only gets wallet information if the trade crashes (the wallet was only being used for crash reporting).
- Only adds the addr -> username mapping to the database if we're not using the default username.
- Adds thread safe nonce handlers in the agent object.
- Don't set nonce for `build_transaction`, and set nonce after. This fixes an issue with requiring transactions to be built in order.

### DIFF
--- a/src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
+++ b/src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
@@ -154,7 +154,7 @@ class HyperdriveAgent:
 
     def _get_nonce_safe(self) -> Nonce:
         with self.nonce_lock:
-            # Since we're handing nonces here, we assume this wallet isn't making other trades
+            # Since we're handling nonces here, we assume this wallet isn't making other trades
             # so we always use the latest block
             chain_nonce = self.chain._web3.eth.get_transaction_count(self.address, "latest")
             if chain_nonce > self.current_nonce:
--- a/src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
+++ b/src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
@@ -76,6 +76,7 @@ class HyperdriveAgent:
     """Interactive Hyperdrive Agent."""
 
     # pylint: disable=too-many-public-methods
+    # pylint: disable=too-many-instance-attributes
 
     ################
     # Initialization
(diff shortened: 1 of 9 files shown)


src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
```diff
@@ -647,69 +630,81 @@ def execute_policy_action(self, pool: Hyperdrive | None = None) -> list[BaseHype
...
+        list[Trade[HyperdriveMarketAction]]
+            The liquidate actions of the underlying policy.
         """
-        if self.account is None:
-            raise ValueError("Must initialize agent with private key for transactions.")
         if pool is None:
             pool = self._active_pool
         if pool is None:
-            raise ValueError("Liquidate requires an active pool.")
+            raise ValueError("Executing policy action requires an active pool.")
```
dpaiton, line 660: executing -> determining?

src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
```diff
@@ -128,13 +139,35 @@ def __init__(
...
+            # Register the username if it was provided
+            add_addr_to_username(self.name, [self.address], self.chain.db_session)
+
+        # The agent object itself maintains it's own nonce for async transactions
+        self.nonce_lock = threading.Lock()
+        self.current_nonce = 0
+
+    def _get_nonce_safe(self) -> Nonce:
+        with self.nonce_lock:
+            # Since we're handing nonces here, we assume this wallet isn't making other trades
```
dpaiton, line 157: "handling"

dpaiton (approved):

PR #1609, opened 2024-07-15
Better nonce handling

slundqui:
There's an issue in nonce handling in that calling a trade that e.g., previews the call before submitting can fail itself (e.g., slippage), which ends up updating the internal nonce count before the transaction gets submitted. We solve this by passing in a nonce function instead of the actual nonce, and call this nonce function at the last possible moment (i.e., when sending the built transaction) to ensure nonce bookkeeping isn't incrementing prematurely.


src/agent0/core/hyperdrive/interactive/exec/execute_agent_trades.py
```diff
@@ -157,8 +155,10 @@ async def async_execute_agent_trades(
         The policy being executed.
     preview_before_trade: bool
         Whether or not to preview the trade before it is executed.
-    base_nonce: Nonce, optional
-        The base nonce to use for this set of trades.
+    nonce_func: Callable[[], Nonce] | None
+        A callable function to use to get a nonce. This function is useful for e.g.,
+        passing in a safe nonce getter tied to an agent.
+        Defaults to setting it to the result of `get_transaction_count`.
```
dpaiton, line 161: It doesn't look like there is a default

src/agent0/core/hyperdrive/interactive/exec/execute_agent_trades.py
```diff
@@ -255,28 +246,23 @@ async def async_execute_single_trade(
     policy: HyperdriveBasePolicy | None, optional
         The policy attached to the agent. This is only used to potentially call `post_action`
         of the policy.
-    nonce: Nonce | None, optional
-        The nonce for the trade.
+    nonce_func: Callable[[], Nonce] | None
+        A callable function to use to get a nonce. This function is useful for e.g.,
+        passing in a safe nonce getter tied to an agent.
+        Defaults to setting it to the result of `get_transaction_count`.
```
dpaiton, line 252: no default

src/agent0/core/hyperdrive/interactive/exec/execute_agent_trades.py
```diff
@@ -385,8 +371,10 @@ async def _async_match_contract_call_to_trade(
         The Hyperdrive API interface object.
     trade_envelope: Trade[HyperdriveMarketAction]
         A specific Hyperdrive trade requested by the given agent.
-    nonce: Nonce, optional
-        Override the transaction number assigned to the transaction call from the agent wallet.
+    nonce_func: Callable[[], Nonce] | None
+        A callable function to use to get a nonce. This function is useful for e.g.,
+        passing in a safe nonce getter tied to an agent.
+        Defaults to setting it to the result of `get_transaction_count`.
```
dpaiton, line 377: no default

src/agent0/core/hyperdrive/interactive/hyperdrive_agent.py
```diff
@@ -153,6 +153,9 @@ def __init__(
         self.current_nonce = 0
 
     def _get_nonce_safe(self) -> Nonce:
```
dpaiton, line 155: might as well make these top-line comments a function docstring, right?

```
Get agent nonces in a thread-safe manner.

We pass in the callable function to underlying ethpy calls so that we get nonce when we sign the transaction.

Returns
--------
Nonce
   The agent's current nonce value.
   ```

PR #1616, opened 2024-07-19
chainsync safe closeout

dpaiton:
Chainsync uses `calc_close_*` for estimating closeout value, which could fail if the pool is in a situation where positions cannot be closed. We ware working on a general solution for this here: https://github.com/delvtech/hyperdrive-rs/pull/153 

This PR is a temporary solution that approximates the closeout value using spot price.

### DIFF
--- a/src/agent0/chainsync/analysis/calc_position_value.py
+++ b/src/agent0/chainsync/analysis/calc_position_value.py
@@ -48,6 +48,32 @@ class _suppress_stdout_stderr:
             os.close(fd)
 
 
+def calc_scaled_normalized_time_remaining(
+    maturity_time: FixedPoint,
+    latest_checkpoint_time: FixedPoint,
+    position_duration: FixedPoint,
+) -> FixedPoint:
+    """Calculate the scaled and normalized time remaining.
+
+    TODO: This exists in hyperdrive-rs; add it to hyperdrivepy.
+
+    Arguments
+    ---------
+    maturity_time: FixedPoint
+        The maturity time of the position.
+    latest_checkpoint_time: FixedPoint
+        The timestamp for the latest checkpoint.
+    position_duration: FixedPoint
+        The pool config setting for position duration.
+
+    Returns
+    -------
+    FixedPoint
+        The scaled and normalized time remaining.
+    """
+    return (maturity_time - latest_checkpoint_time) / position_duration
+
+
 def calc_single_closeout(
     position: pd.Series,
     interface: HyperdriveReadInterface,
@@ -86,6 +112,7 @@ def calc_single_closeout(
     token_type = position["token_type"]
     maturity = 0
     position_duration = hyperdrive_state.pool_config.position_duration
+    checkpoint_time = hyperdrive_state.checkpoint_time
 
     if token_type in ["LONG", "SHORT"]:
         maturity = int(position["maturity_time"])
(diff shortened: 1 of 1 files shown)


src/agent0/chainsync/analysis/calc_position_value.py
```diff
@@ -98,7 +125,17 @@ def calc_single_closeout(
...
-            logging.info("Chainsync: Exception caught in calculating close long, ignoring: %s", exception)
+            logging.info(
+                "Chainsync: Exception caught in calculating close long: %s\nApproximating with spot price.", exception
+            )
+
+            # TODO: We can use the rust `calculate_market_value_*` functions once
+            # https://github.com/delvtech/hyperdrive-rs/pull/153 is merged.
+            # Long value = users_longs * spot_price * term_remaining
+            normalized_time_remaining = calc_scaled_normalized_time_remaining(
+                maturity, FixedPoint(checkpoint_time), FixedPoint(position_duration)
```
slundqui, line 136: No need to cast to `FixedPoint` here, the interface should already have this as FixedPoint. Removing it will also ensure type safety, just in case `checkpoint_time` and `position_duration` is an int.

src/agent0/chainsync/analysis/calc_position_value.py
```diff
@@ -145,7 +182,19 @@ def calc_single_closeout(
...
-            logging.info("Chainsync: Exception caught in calculating close short, ignoring: %s", exception)
+            logging.info(
+                "Chainsync: Exception caught in calculating close short: %s\nApproximating with spot price.", exception
+            )
+
+            # TODO: We can use the rust `calculate_market_value_*` functions once
+            # https://github.com/delvtech/hyperdrive-rs/pull/153 is merged.
+            # Short value = users_shorts * ( 1 - spot_price ) * term_remaining
+            normalized_time_remaining = calc_scaled_normalized_time_remaining(
+                maturity, FixedPoint(checkpoint_time), FixedPoint(position_duration)
```
slundqui, line 193: Again, no cast to `FixedPoint`

slundqui (approved):

slundqui (approved):
LGTM, but this is assuming the equation itself is correct in estimating PNL.

src/agent0/chainsync/analysis/calc_position_value.py
```diff
@@ -48,6 +48,32 @@ def __exit__(self, *_):
             os.close(fd)
 
 
+def calc_scaled_normalized_time_remaining(
+    maturity_time: FixedPoint,
+    latest_checkpoint_time: FixedPoint,
```
slundqui, line 53: This should be an integer, not fixedpoint

src/agent0/chainsync/analysis/calc_position_value.py
```diff
@@ -48,6 +48,32 @@ def __exit__(self, *_):
...
+    position_duration: FixedPoint,
```
slundqui, line 54: Position duration should also be a int

src/agent0/chainsync/analysis/calc_position_value.py
dpaiton: the interface provides ints 
image

slundqui: Scratch that, both `checkpoint_time` and `position_duration` should be integers, representing epoch time instead of a `FixedPoint` number.

PR #1622, opened 2024-07-24
add interface functions for calculating market estimates

dpaiton:
- updates interface for latest hyperdrivepy (including some changes to fn signatures)
- adds `calculate_market_value_*` functions

We will want to follow up with a PR that uses these functions in chainsync to compute the unrealized gains.

### DIFF
--- a/tests/min_txn_amount_test.py
+++ b/tests/min_txn_amount_test.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 from fixedpointmath import FixedPoint
-from utils import expect_failure_with_funded_bot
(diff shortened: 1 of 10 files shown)


tests/min_txn_amount_test.py
```diff
@@ -6,7 +6,7 @@
 
 import pytest
 from fixedpointmath import FixedPoint
-from utils import expect_failure_with_funded_bot
+from utils import expect_failure_with_funded_bot  # type: ignore
```
slundqui, line 9: What's the type error here?

slundqui (approved):

PR #1631, opened 2024-07-25
Don't use threads in checkpoint bot, and make checkpoint bot use async waits

slundqui:


wakamex (approved):
lgtm I guess we still use async_runner elsewhere

PR #1636, opened 2024-08-01
Wrap getting anvil state dump in a chain config flag, defaulting to False

slundqui:
We don't use anvil state dumps for crash debugging anymore, due to complexities around loading state from a state dump. Additionally, known failures on testnet attempt to get anvil state dump, which pollutes alchemy error loggings.


dpaiton (approved):
please add a test ensuring the dumped state is in the crash report if the bool is set to True so that this code does not decay

PR #1637, opened 2024-08-05
Fuzz fork of testnet

slundqui:
This PR allows us to fuzz all registered pools on sepolia testnet. We fund these accounts through transfers from whales.

TODO
- Allow advancing time, which require implementing accuring interest based on the yield source.
- Allow randomizing variable rate, which require implementing accuring interest based on yield source.

### DIFF
--- a/src/agent0/hyperfuzz/system_fuzz/run_fuzz_bots.py
+++ b/src/agent0/hyperfuzz/system_fuzz/run_fuzz_bots.py
@@ -162,7 +162,7 @@ def run_fuzz_bots(
     ---------
     chain: Chain
         The chain to run the bots on.
-    hyperdrive_pools: Hyperdrive | list[Hyperdrive]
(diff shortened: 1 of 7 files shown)


src/agent0/hyperfuzz/system_fuzz/run_fuzz_bots.py
```diff
@@ -136,7 +136,7 @@ def generate_fuzz_hyperdrive_config(rng: Generator, lp_share_price_test: bool, s
 
 def run_fuzz_bots(
     chain: Chain,
-    hyperdrive_pools: Hyperdrive | list[Hyperdrive],
+    hyperdrive_pools: Hyperdrive | Sequence[Hyperdrive],
```
dpaiton, line 139: change docstring too

dpaiton (approved):

PR #1645, opened 2024-08-08
Ethpy interface fixes

slundqui:
- Adds cache to getting read interface from read write interface.
- Adds known (but no special handling) hyperdrive kinds to read write interface.


wakamex (approved):
some day we may want to import the `HyperdriveKind` enum outside of `read_interface.py` in which case defining it separately would be nice, but that day isn't today

PR #1654, opened 2024-08-19
Dashboard updates

slundqui:
- Dataframes views in dashboard gets converted to strings to avoid warning when viewing small numbers.
- Dashboard defaults to finding an open port by default.
- Chain config `anvil_verbose` changed to `verbose`, and also controls dashboard subprocess output.

### DIFF
--- a/src/agent0/core/hyperdrive/interactive/local_chain.py
+++ b/src/agent0/core/hyperdrive/interactive/local_chain.py
@@ -734,24 +734,44 @@ class LocalChain(Chain):
         IFrame
             A dashboard IFrame that can be shown in a Jupyter notebook with the `display` command.
         """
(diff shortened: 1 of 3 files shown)


src/agent0/core/hyperdrive/interactive/local_chain.py
```diff
@@ -806,24 +806,36 @@ def run_dashboard(self, blocking: bool = False) -> None:
...
-                str(self.config.dashboard_port),
-                "--server.address",
-                "localhost",
-            ]
-        )
+        streamlit_cli_flags = [
+            "--server.address",
+            "localhost",
+        ]
+        if self.config.dashboard_port is not None:
```
dpaiton, line 813: make this change in `get_dashboard_iframe` as well please

dpaiton (approved):

PR #1659, opened 2024-08-21
Update dependencies

slundqui:


src/agent0/ethpy/base/transactions.py
```diff
@@ -499,7 +546,7 @@ async def _async_send_transaction_and_wait_for_receipt(
     unsent_txn["nonce"] = nonce
 
     signed_txn = signer.sign_transaction(unsent_txn)  # type: ignore
-    tx_hash = web3.eth.send_raw_transaction(signed_txn.raw_transaction)
+    tx_hash = web3.eth.send_raw_transaction(signed_txn.rawTransaction)
```
slundqui, line 549: This seems weird, `eth_account` made an update to change this variable name. Looks like it's resolving to v0.10.1, which isn't the newest version

dpaiton: can you pin the `eth_account` version to the latest?

slundqui: Resolved, see comments in `pyproject.toml`

dpaiton (approved):

PR #1668, opened 2024-08-28
Updating agent0 for pypechain support for link referencing in deploy.

slundqui:
Requires pypechain >= 0.0.37


sentilesdal (approved):
approve, but fix contract name keys

PR #1675, opened 2024-09-05
docs: update INSTALL.md

eltociear:
virual -> virtual


dpaiton:
There was a failed check, so we didn't approve. The fix is no longer relevant due to other refactors.

PR #1682, opened 2024-09-11
Don't checkpoint on paused pools. Don't check for previous checkpoints in invariance checks if pool is paused

slundqui:


src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -301,6 +301,10 @@ def _check_previous_checkpoint_exists(
     ):
         return InvariantCheckResults(failed=False, exception_message=None, exception_data={}, log_level=None)
 
+    # We ignore this test if the pool is paused
+    if interface.get_pool_is_paused():
```
dpaiton, line 305: should this be `if not interface.get_pool_is_paused():`?

dpaiton (approved):

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
slundqui: No, this returns immediately with `failed=False` if pool is paused

PR #1688, opened 2024-09-23
Add invariance check for circuit breaker trip

dpaiton:


slundqui (approved):

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -202,6 +205,52 @@ class InvariantCheckResults(NamedTuple):
...
+    previous_weighted_spot_apr = interface.calc_rate_given_fixed_price(
+        previous_pool_state.checkpoint.weighted_spot_price,
+        FixedPoint(previous_pool_state.pool_config.position_duration),
+    )
+    current_weighted_spot_apr = interface.calc_rate_given_fixed_price(
+        pool_state.checkpoint.weighted_spot_price,
+        FixedPoint(pool_state.pool_config.position_duration),
+    )
+
+    if abs(current_weighted_spot_apr - previous_weighted_spot_apr) >= delta_rate_epsilon:
```
jalextowle, line 239: It seems to me that this warning is only going to let us know when the attack is possible. Could we instead do this update when the spot price is greater than the previous weighted spot price? This would tell us that there may be a problem as soon as the trade occurs.

dpaiton: greater than? would this not happen every long?

PR #1700, opened 2024-10-08
Removing retries in agent0 in favor of web3py's http provider retry system (enabled by default)

slundqui:
Also adds caching provided via provider.

Default retry behavior:
https://web3py.readthedocs.io/en/stable/internals.html#retry-requests-for-http-providers

Added caching behavior:
https://web3py.readthedocs.io/en/stable/internals.html#retry-requests-for-http-providers


dpaiton (approved):
✌️

PR #1709, opened 2024-10-21
Support for fuzzing on forks

slundqui:
This PR adds whale addresses and various fixes to allow for fuzzing of deployed pools via forking.

Minor changes:
- Bug fix for db ingestion with chains with accelerated block mining.
- Logging caught exceptions in fuzzing scripts.
- Renaming `chain.chain_id` to `chain.chain_obj_id`, which is used for identifying an agent0 chain object, to avoid confusion with Ethereum chain ids.
- Adding `chain_id` property to `chain` that reports the Ethereum chain id.


src/agent0/chainsync/db/hyperdrive/chain_to_db.py
```diff
@@ -143,6 +144,13 @@ def checkpoint_events_to_db(
         # NOTE we get all numeric arguments in events as string to prevent precision loss
 
         from_block = get_latest_block_number_from_checkpoint_info_table(db_session, interface.hyperdrive_address) + 1
+
+        # Don't look back earlier than the defined earliest block for this chain
```
dpaiton, line 148: you have this comment twice

dpaiton (approved):

PR #1711, opened 2024-10-22
Refork periodically

slundqui:
Minor fixes:
- Bugfix to random policy to not attempt to remove liquidity if the amount of LP the wallet contains is less than the minimum transaction amount.
- Bugfix for minimum transaction amounts in the case where agent0 is dealing with share amounts.
- Adding rollbar ignores for fork fuzzing to avoid spamming rollbar.


dpaiton (approved):

src/agent0/core/hyperdrive/policies/random.py
```diff
@@ -113,9 +113,16 @@ def get_available_actions(
             A list containing all of the available actions.
         """
         pool_state = interface.current_pool_state
+        # The amount of minimum transaction amount is dependent on if we're trading with
```
dpaiton, line 116: > amount of minimum transaction amount
reword

PR #1725, opened 2024-10-31
Minor fixes for aero pool

slundqui:
- Adding whale account in fork fuzzing.
- Fund as much as possible in the case where whale doesn't have enough balance.
- Call `totalShares` in `get_vault_shares` with fallback.
- Set `vault_shares_token_contract` to None if it's the zero address.
- `get_variable_rate` returns None if `vault_shares_token_contract` is None.


scripts/fork_fuzz_bots.py
```diff
@@ -25,6 +25,7 @@
 # with the value as the whale address.
 # Note that if a token is missing in this mapping, we will try to
 # call `mint` on the trading token to fund.
+# TODO we need a better method to ensure this list of whales doesn't get stale
```
dpaiton, line 28: would be an easy test to add to CI
fund a bot from the whale, ensure success for some X base

dpaiton (approved):

PR #1728, opened 2024-11-04
add `block_number_before_timestamp`; `chain.mine()`; ezeth price check script

dpaiton:


src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
```diff
@@ -318,15 +318,23 @@ def _check_negative_interest(interface: HyperdriveReadInterface, pool_state: Poo
         # Different error messages and log levels if the pool is paused
         if interface.get_pool_is_paused():
             exception_message = (
-                "Negative interest detected on paused pool. "
+                "Negative interest detected between block "
+                f"{previous_pool_state.block_number}"
```
slundqui, line 322: Can you add a timestamp here?

src/agent0/utils/block_before_timestamp.py
```diff
@@ -0,0 +1,74 @@
...
+    earlier_block_timestamp = web3.eth.get_block(earlier_block_number).get("timestamp", None)
+    assert earlier_block_timestamp is not None
+    avg_time_between_blocks = (current_block_timestamp - earlier_block_timestamp) / earlier_block_delta
+
+    # Estimate the block number corresponding to the user provided timestamp
+    delta_time = current_block_timestamp - block_timestamp
+    estimated_block_number = current_block_number - (delta_time // avg_time_between_blocks)
+
+    # Use a binary search to find the block
+    left = int(max(0, estimated_block_number - 100))  # search 100 blocks before estimated block
```
slundqui, line 49: Can you handle the case where we make a bad guess, and the query block timestamp is outside of initial left/right?

slundqui:
```
    # Ensure bounds of binary search is within the target timestamp
    if web3.eth.get_block(left)["timestamp"] > block_timestamp:
        left = 0
    if web3.eth.get_block(right)["timestamp"] < block_timestamp:
        right = current_block_number
```

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
dpaiton: I figured we could look up the timestamp online. Do you want timestamp AND block number or instead of? I was worried that both would clog up the discord notif.

scripts/check_ezeth.py
```diff
@@ -0,0 +1,74 @@
...
+        start_pool_state = ezeth_pool.interface.get_hyperdrive_state(block_identifier=start_block_number)
+        start_vault_share_price = start_pool_state.pool_info.vault_share_price
+
+        lookback_timestamp = start_block_timestamp - lookback_length
+        lookback_block_number = block_number_before_timestamp(web3, lookback_timestamp)
+        lookback_pool_state = ezeth_pool.interface.get_hyperdrive_state(block_identifier=lookback_block_number)
+        lookback_vault_share_price = lookback_pool_state.pool_info.vault_share_price
+
+    print(f"time = {start_block_timestamp}; vault share price = {start_vault_share_price}")
+    print(f"time = {lookback_timestamp}; vault share price = {lookback_vault_share_price}")
```
slundqui, line 53: For the sake of ensuring `block_number_before_timestamp` is correct, you should also print out the block number

src/agent0/hyperfuzz/system_fuzz/invariant_checks.py
slundqui: I guess that's true, but I'd rather have the information there, shouldn't be too much more info

slundqui (approved):

